PR #4, opened 2015-08-31
Aggregation improvements

jgallagher59701:
The aggregation_improvements branch includes code in the dap module to free Array variable memory once those variables have been serialized. This reduces the memory required by the BES when serializing responses with several variables. Note that the code calls clear_local_data() for every variable, but it's primarily Arrays where the benefit is seen.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,8 @@ addons:
   ssh_known_hosts: www.opendap.org
   coverity_scan:
     project:
-      name: "OPENDAP/bes"
+    name: "OPENDAP/bes"
+     description: "Build submitted via Travis CI"
     notification_email: jgallagher@opendap.org
     build_command: "make -j7"
     branch_pattern: coverity_scan
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,6 @@ language: cpp
 
 compiler:
   - gcc
-  # - clang
 
 # Whitelist; assume tags for releases use the familiar dot notation
 # Building a tag will trigger a deployment; see the 'deploy' section
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,8 +34,8 @@ addons:
   ssh_known_hosts: www.opendap.org
   coverity_scan:
     project:
-    name: "OPENDAP/bes"
-     description: "Build submitted via Travis CI"
+      name: "OPENDAP/bes"
+      description: "Build submitted via Travis CI"
     notification_email: jgallagher@opendap.org
     build_command: "make -j7"
     branch_pattern: coverity_scan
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -82,21 +83,22 @@ install:
 #
 # We want the hdf5 handler build to find the hdf5 lib headers from the
 # source tarball we just built so run configure using CPPFLAGS=...
+# removed CPPFLAGS=-I$HOME/deps/include added --with-hdf5... jhrg 9/3/15
 
 before_script:
   - ln -s configure_modules.ac configure.ac
   - autoreconf --force --install --verbose
-  - ./configure --prefix=$HOME/build CPPFLAGS=-I$HOME/deps/include
+  - ./configure --prefix=$HOME/build --with-hdf5=$HOME/deps
```
mrpatrickwest: Why is hdf5 required to build the BES? I would think that the only dependency would be libdap?

install_bes_source_deps.sh
```diff
@@ -2,18 +2,31 @@
```
mrpatrickwest: Is this file part of the travis-ci build? There's no comments saying what this file is for.

modules/ncml_module
```diff
@@ -1 +1 @@
-Subproject commit 92e95489cdf742ef807a753a09f858a6b44c87e5
```
mrpatrickwest: Same question as above, curious why ncml_module is a dependency, submodule in this case, for the BES? Same for xml_data_handler

timing/mkAllTests
```diff
@@ -0,0 +1,59 @@
+#!/bin/bash
```
mrpatrickwest: Is this something run by travis? Again, no comments explaining what this file is for.

timing/mkCurlTests
```diff
@@ -0,0 +1,98 @@
+#!/bin/bash
```
mrpatrickwest: Same comment as above for mkAllTests

mrpatrickwest:
where's the code for this branch. Comment says that there's a code fix in the dap module to free memory, but I don't see source code committed for the pull request.

install_bes_source_deps.sh
jgallagher59701: It is used to install bes source dependencies for the travis ci build

timing/mkAllTests
jgallagher59701: No, this is not run by travis. All the files in 'timing' are for running timing analysis on the BES.

modules/ncml_module
jgallagher59701: Because the travis build runs the 'modules' build of the BES.

PR #5, opened 2015-08-31
Compile, test and deploy BES using standard configuration

mrpatrickwest:
Attempting to build a standard configuration of the BES using just the
modules that I need and without all of the dependencies that I don't
need. So using the configure_standard.ac configuration file.

The besd.in file which builds the besd initialization script that goes
into /etc/init.d was not setting up the prefix variables correctly so
the bes was not launching using that script.

configure_standard.ac seemed to have not been updated with the other
configuration file as it appears to be working.

dapreader Makefile.am was missing the inclusion of the dap server libs
to resolve the Ancillary methods used, specificall reading a das
ancillary file.

### DIFF
--- a/besd.in
+++ b/besd.in
@@ -12,6 +12,8 @@
 # Path to the besctl script. Note: If the BES is installed using a prefix
 # that is not set in the besctl script, add '-i <bes_prefix>' to the value
 # of besctl here. This should only be needed in unusual situations.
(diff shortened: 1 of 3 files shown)


jgallagher59701:
Should I delete this branch or would you like it to stick around?

mrpatrickwest:
You can delete it

PR #6, opened 2015-09-23
Sigpipe fix

jgallagher59701:
Please look over these changes to ServerApp.cc - They are a fix for ticket https://bugs.earthdata.nasa.gov/browse/ODIP-149 (which I hope you can see...). Summary: We've noticed (as have others) that the child listeners can 'runaway' when a client (curl, ...) or tomcat exit (kill -9, ...). This is not so bad most of the time, because most requests finish quickly. But for large aggregations... 

Catching sigpipe seems to be the fix, even though a more typical approach is to test error returns from write(2). That doesn't seem to work for us - maybe an ostream thing, maybe a buffering on AWS thing

### DIFF
--- a/server/BESServerHandler.h
+++ b/server/BESServerHandler.h
@@ -22,7 +22,7 @@
 //
 // You can contact University Corporation for Atmospheric Research at
 // 3080 Center Green Drive, Boulder, CO 80301
- 
+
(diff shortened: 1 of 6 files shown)


mrpatrickwest:
:shipit:

PR #7, opened 2015-09-25
Besdebug xml bug

jgallagher59701:
Fixed an issue where I had commented out the xmlCleanupParser(); call for libxml2. This was causing some really odd behavior on linux (never on osx) where certain BESDEBUG context switches would cause the Dimension elements of an Array in DAP4 DMR responses to differ. I fixed this in several places - and am looking to see if it's an issue for SAX parsing as well, but that's a separate question. 

I also removed some  of the BESDEBUG("besxml" ...) calls because they made it hard to read the xml module code. Ive kept the ones that seem to provide the most help if we need to debug this code.

### DIFF
--- a/server/DaemonCommandHandler.cc
+++ b/server/DaemonCommandHandler.cc
@@ -889,14 +889,10 @@ void DaemonCommandHandler::execute_command(const string &command, BESXMLWriter &
 	}
 	catch (...) {
 		xmlFreeDoc(doc);
-		xmlCleanupParser();
 		throw;
 	}
 
 	xmlFreeDoc(doc);
-	// Added this call based on a problem (BES-40) in BESXMLInterface.cc.
-	// jhrg 9/25/15
-	xmlCleanupParser();
 }
 
 static void send_bes_error(BESXMLWriter &writer, BESError &e)
--- a/server/DaemonCommandHandler.cc
+++ b/server/DaemonCommandHandler.cc
@@ -262,182 +262,6 @@ static void write_file(const string &name, const string &buffer)
 	}
 }
 
-#if 0
-
-// This has an infinite loop in it under some circumstances.
-
-/** This version of get_bes_log_lines() is an attempt to improve the efficiency
- * of the process using an estimate of the size of the log lines to skip over
- * the vast majority of the log file.
- *
- * @note if num_lines is zero, that means get the whole file. This will have
- * the awful side affect of allocating memory for the whole file, which may be
- * very large.
- *
- * @todo Change the estimated number of chars per log line dynamically.
- */
-#define BES_LOG_CHARS_EST_PER_LINE 126
-
-static char *get_bes_log_lines(const string &log_file_name, long num_lines)
-{
-	ifstream infile(log_file_name.c_str(), ios::in | ios::binary);
-	if (!infile.is_open())
-	throw BESInternalError("Could not open file for reading (" + log_file_name + ")", __FILE__, __LINE__);
-
-	// This is used to save time counting lines in large files
-	static ifstream::pos_type prev_end_pos = 0;
-	// static long prev_line_count = 0;
-
-	ifstream::pos_type start_from_pos = 0;
-	// num_line == 0 is special value that means get all the lines
-	if (num_lines > 0)
-	{
-		// Get size of file in bytes, then estimate where to start looking for
-		// num_lines lines, then set the file pointer there.
-		infile.seekg(0, ios::end);
-		ifstream::pos_type end_pos = infile.tellg();
-		long est_num_lines = end_pos / BES_LOG_CHARS_EST_PER_LINE;
-		if (num_lines >= est_num_lines)
-		infile.seekg(0, ios::beg);
-		else
-		infile.seekg((est_num_lines - num_lines) * BES_LOG_CHARS_EST_PER_LINE, ios::beg);
-		ifstream::pos_type start_from_pos = infile.tellg();
-
-		BESDEBUG("besdaemon", "DaemonCommandHandler()::get_bes_log_lines() - end_pos: " << end_pos << " start_from_pos: " << start_from_pos << endl);
-
-		// No laughing at my goto...
-		retry:
-		// start_from_pos points to where we start looking for num_lines
-		long count = 0;
-		while (!infile.eof() && !infile.fail())
-		{
-			infile.ignore(1024, '\n');  // Assume no line is > 1024
-			++count;
-		}
-
-		infile.clear(); // Needed to reset eof() or fail() so tellg/seekg work.
-
-		// if the start_from_pos is too far along (there are not num_lines
-		// left), scale it back and try again.
-		if (count < num_lines) {
-			BESDEBUG("besdaemon", "DaemonCommandHandler()::get_bes_log_lines() - Retrying; Log length  (count)" << count << ", num_lines " << num_lines << endl);
-			// 10 isa fudge factor...
-			long size = start_from_pos;
-			size -= ((num_lines - count + 10) * BES_LOG_CHARS_EST_PER_LINE);
-			infile.seekg(size, ios::beg);
-			start_from_pos = infile.tellg();
-			goto retry;
-		}
-
-		infile.seekg(start_from_pos, ios::beg);
-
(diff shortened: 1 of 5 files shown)


mrpatrickwest:
This certainly smells like a memory issue, when adding or removing debug statements seems to fix the issue.

I could see removing most of those module debug statements in favor of debug statements at the end of module initialization that would dump the various lists, like response handler list, request handler list, function list, transmitter list, catalog list, reporter list, registered debug, etc, etc, etc... That way we don't have to depend on module developers to put in those types of debug statements.

Seems that most of the changes in this pull request are removing '#if 0' blocks, reformatting lines, removing debug statements, etc... and the one change that's of interest is adding the parser cleanup call. Correct?

Is depending on the order of xml elements in an xml document a good idea in general? Perhaps we need to re-think how we determine if tests succeed or not?

ndp-opendap:
Depending on the order of elements in an XML document (document order) is a widespread and reasonable approach. The primary repair here:

> "Fixed an issue where I had commented out the xmlCleanupParser(); call for libxml2”

Fixed the problem, as the failure to reset the parser left it in an indeterminate state of the next test/usage.

Basically, if we can’t rely on document order when processing XML documents then we must abandon entirely the use of XML as a representation for the DMR and DDX. But we CAN rely on it, as pretty much every DOM API that I am aware of has methods/functions that allow you to iterate through child elements in document order. W.r.t. SAX parsers, these are more specialized and care must be taken when implementing them and using as we have seen in this particular issue.

Basically the take away is that failing to reset the state of a SAX parser instance that is going to be used in a reentrant manner is a no-no.



> On Sep 26, 2015, at 6:46 AM, Patrick West <notifications@github.com> wrote:
> 
> This certainly smells like a memory issue, when adding or removing debug statements seems to fix the issue.
> 
> I could see removing most of those module debug statements in favor of debug statements at the end of module initialization that would dump the various lists, like response handler list, request handler list, function list, transmitter list, catalog list, reporter list, registered debug, etc, etc, etc... That way we don't have to depend on module developers to put in those types of debug statements.
> 
> Seems that most of the changes in this pull request are removing '#if 0' blocks, reformatting lines, removing debug statements, etc... and the one change that's of interest is adding the parser cleanup call. Correct?
> 
> Is depending on the order of xml elements in an xml document a good idea in general? Perhaps we need to re-think how we determine if tests succeed or not?
> 
> —
> Reply to this email directly or view it on GitHub.
> 

= = =
Nathan Potter                        ndp at opendap.org
OPeNDAP, Inc.                        +1.541.231.3317

jhrg:
> On Sep 26, 2015, at 8:05 AM, Nathan Potter <notifications@github.com> wrote:
> 
> 
> Depending on the order of elements in an XML document (document order) is a widespread and reasonable approach. The primary repair here:
> 
> > "Fixed an issue where I had commented out the xmlCleanupParser(); call for libxml2”
> 
> Fixed the problem, as the failure to reset the parser left it in an indeterminate state of the next test/usage.
> 
> Basically, if we can’t rely on document order when processing XML documents then we must abandon entirely the use of XML as a representation for the DMR and DDX. But we CAN rely on it, as pretty much every DOM API that I am aware of has methods/functions that allow you to iterate through child elements in document order. W.r.t. SAX parsers, these are more specialized and care must be taken when implementing them and using as we have seen in this particular issue.

Right. Order is preserved, so you can rely on it, although it’s the sort of thing that can provoke debates…
> 
> Basically the take away is that failing to reset the state of a SAX parser instance that is going to be used in a reentrant manner is a no-no.

Close. I learned more. The code in question is not using SAX but DOM. And the libxml2 docs seem to say that calling xmlCleanupParser() is fairly benign - but when I applied a similar patch/fix/change to the besdeamon, I got a runtime error (double free). I think libxml2 is not without its issues… All the same, this change to the beslistener seems to be an improvement.

> 
> 
> 
> > On Sep 26, 2015, at 6:46 AM, Patrick West <notifications@github.com> wrote:
> >
> > This certainly smells like a memory issue, when adding or removing debug statements seems to fix the issue.
> >
> > I could see removing most of those module debug statements in favor of debug statements at the end of module initialization that would dump the various lists, like response handler list, request handler list, function list, transmitter list, catalog list, reporter list, registered debug, etc, etc, etc... That way we don't have to depend on module developers to put in those types of debug statements.
> >
> > Seems that most of the changes in this pull request are removing '#if 0' blocks, reformatting lines, removing debug statements, etc... and the one change that's of interest is adding the parser cleanup call. Correct?

Yes. I tend to remove code using #if 0 … #endif and then actually remove the stuff after several commits - it’s annoying because the removal happens twice but it’s easier than depending on cvs/svn/git to get stuff back when I need it.

> > Is depending on the order of xml elements in an xml document a good idea in general? Perhaps we need to re-think how we determine if tests succeed or not?

I’d like to make the tests smarter (although I think element order is dependable) but it takes considerable time, so I cannot see us doing that anytime soon. Bummer, but we’re stuck with string and regex compares for now and the latter are a royal pain.

Thanks for getting back to me - I’ll try to make my pull requests focus on just changes we need to think about and not include lots of reformatting junk.

James
> >
> > —
> > Reply to this email directly or view it on GitHub.
> >
> 
> = = =
> Nathan Potter ndp at opendap.org
> OPeNDAP, Inc. +1.541.231.3317
> 
> 
> 
> 
> —
> Reply to this email directly or view it on GitHub <https://github.com/OPENDAP/bes/pull/7#issuecomment-143447882>.
> 

--
James Gallagher
jhrg at mac.com

mrpatrickwest:
> On Sep 27, 2015, at 11:20 AM, James Gallagher <notifications@github.com> wrote:
> 
> 
> > On Sep 26, 2015, at 8:05 AM, Nathan Potter <notifications@github.com> wrote:
> > 
> > 
> > Depending on the order of elements in an XML document (document order) is a widespread and reasonable approach. The primary repair here:
> > 
> > > "Fixed an issue where I had commented out the xmlCleanupParser(); call for libxml2”
> > 
> > Fixed the problem, as the failure to reset the parser left it in an indeterminate state of the next test/usage.
> > 
> > Basically, if we can’t rely on document order when processing XML documents then we must abandon entirely the use of XML as a representation for the DMR and DDX. But we CAN rely on it, as pretty much every DOM API that I am aware of has methods/functions that allow you to iterate through child elements in document order. W.r.t. SAX parsers, these are more specialized and care must be taken when implementing them and using as we have seen in this particular issue.
> 
> Right. Order is preserved, so you can rely on it, although it’s the sort of thing that can provoke debates…
> > 
> > Basically the take away is that failing to reset the state of a SAX parser instance that is going to be used in a reentrant manner is a no-no.
> 
> Close. I learned more. The code in question is not using SAX but DOM. And the libxml2 docs seem to say that calling xmlCleanupParser() is fairly benign - but when I applied a similar patch/fix/change to the besdeamon, I got a runtime error (double free). I think libxml2 is not without its issues… All the same, this change to the beslistener seems to be an improvement.
> 
> > 
> > 
> > 
> > > On Sep 26, 2015, at 6:46 AM, Patrick West <notifications@github.com> wrote:
> > >
> > > This certainly smells like a memory issue, when adding or removing debug statements seems to fix the issue.
> > >
> > > I could see removing most of those module debug statements in favor of debug statements at the end of module initialization that would dump the various lists, like response handler list, request handler list, function list, transmitter list, catalog list, reporter list, registered debug, etc, etc, etc... That way we don't have to depend on module developers to put in those types of debug statements.
> > >
> > > Seems that most of the changes in this pull request are removing '#if 0' blocks, reformatting lines, removing debug statements, etc... and the one change that's of interest is adding the parser cleanup call. Correct?
> 
> Yes. I tend to remove code using #if 0 … #endif and then actually remove the stuff after several commits - it’s annoying because the removal happens twice but it’s easier than depending on cvs/svn/git to get stuff back when I need it.
> 
> > > Is depending on the order of xml elements in an xml document a good idea in general? Perhaps we need to re-think how we determine if tests succeed or not?
> 
> I’d like to make the tests smarter (although I think element order is dependable) but it takes considerable time, so I cannot see us doing that anytime soon. Bummer, but we’re stuck with string and regex compares for now and the latter are a royal pain.
> 
> Thanks for getting back to me - I’ll try to make my pull requests focus on just changes we need to think about and not include lots of reformatting junk.

No worries … I just wanted to make sure that I didn’t miss another change. Feel free to clean up formatting.

PR #8, opened 2015-10-06
Cache control patch 02

ndp-opendap:
I discovered serious flaws in the way that BESFileLockingCache and it's child classes were being used.

1. One problem was that whichever child class of BESFileLockingCache was instantiated first defined the cache size, cache prefix, and cache directory. After this no other instance could alter the values and this happened silently so that it was not clear what was happening.
2. All of the various caches were using the SAME cache control file (ouch).
3. Another issue was that the caching code was generating overly long file names. This is still an issue, but now there is a trap in place to do something intelligent when it does happen. In the future (i.e. soon!) we will need to start using a hash algorithm for the cache file names.

Stuff I did:
* Refactored BEFileLocking Cache and all of its child classes. 
* Eliminated the use of the deeply borked BESCache3 class (although I did not eliminate the file at this time I did gut it using "#if 0 ... #endif" and I removed it from the build)
* Created a BESUncompressCache class (because that's the right name for it) with the correct specializations for file naming etc. I edited ALL of the bes.conf.in files to explicitly use this class in place of the now defunct BESCache3 by changing the BES Keys associated with the decompression cache.
* Created GatewayCache for gateway responses.
* Fixed a bunch of lifecycle problems in the various children of BESFileLockingCache


mrpatrickwest:
This all looks good to me. More control over different caching mechanisms, deleting instance data at exit, better error messages, more debugging, updated config files, ...

jhrg:
Sent from my iPhone

> On 6/10/2015, at 7:27 PM, Patrick West <notifications@github.com> wrote:
> 
> This all looks good to me. More control over different caching mechanisms, deleting instance data at exit, better error messages, more debugging, updated config files, ...
> 

Yes. I think we're ready to merge and then bring in the new join existing caching in the ncml handler. 

James

PR #10, opened 2015-10-16
Cachename bug

ndp-opendap:
I fixed a small bug in the way that the BES builds paths for accessing files. In particular, BESContainerStorageVolatile.cc was using a pattern of:

`string path = part1 + "/" + part2;`

Which caused, depending on the request and configuration, the resulting value of `path` to contain one, two, or even three "/" characters between `part1` and `part2`. This works fine for file access, but was causing the cache file names to contain multiple "#" inlace of the "/", and since other parts of the code were independently assembling the value of `path` without the multiple slash characters the same file was being referenced but the cache file name was different (cue sound of glass breaking). 

This patch addresses the issue by adding a the method BESUtils::assemblePath() and then using it where needed.

### DIFF
--- a/dap/BESDapResponseBuilder.cc
+++ b/dap/BESDapResponseBuilder.cc
@@ -90,6 +90,7 @@
 #include "BESContextManager.h"
 #include "BESDapResponseCache.h"
 #include "BESStoredDapResultCache.h"
+#include "BESUtil.h"
 #include "BESDebug.h"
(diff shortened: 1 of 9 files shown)


jgallagher59701:
Looks like a good patch to me - we'll need to merge the corresponding branches in the handlers. I thought that was in the pull request email, but I don't see it here. Anyway, that is not a big deal.

PR #14, opened 2015-12-30
Bes 38

jgallagher59701:
I've added a timeout feature to the BES by modifying the code in BESInterface so that it sets the alarm() to a value specified by a BES Context (that can be passed in from the OLFS, etc.). The handler for the alarm signal uses longjmp to stop whatever is happening when the signal is delivered and return an error to the BES's client. We've added a new kind of BES Error for this, so the OLFS can sort it out. In the process of doing this I have made a number of changes to the timeout code in BESDapResponseBuilder, removing most of what was there for timeouts (it did not work correctly).

Tests for the feature are on the bes-38 branch of the debug_functions module. You may also need to pull libdap to build - at one point that was needed, but I think I removed the dependency. Regardless, I'll be reworking the timeout code there too, since this scheme is much better (i.e., it works).

I'd like to get this code out to the folks at LARC soon - before ESIP next Wed.

Thanks,
James


dap/BESDapResponseBuilder.cc
```diff
@@ -80,11 +83,13 @@
 #include <mime_util.h>	// for last_modified_time() and rfc_822_date()
 #include <escaping.h>
 #include <util.h>
+#if USE_LOCAL_TIMEOUT_SCHEME
```
mrpatrickwest: Are you using a macro here because the alarm doesn't work on some systems?

dap/BESDapResponseBuilder.cc
```diff
@@ -284,24 +272,104 @@ int BESDapResponseBuilder::get_timeout() const
...
+ * seconds unless timeout_off() is called first.
+ *
+ * @deprecated
+ */
+void
+BESDapResponseBuilder::timeout_on() const
+{
+#if USE_LOCAL_TIMEOUT_SCHEME
+#ifndef WIN32
+    alarm(d_timeout);
```
mrpatrickwest: We have a SignalHandler class that could handle some of this, if not all of this. Wouldn't have to worry about using the macros everywhere. timeout_on would call a method on the SignalHandler instance to turn on timeout, turn off timeout. The underlying code is encapsulated in the SignalHandler class.

dap/BESDapResponseBuilder.cc
jgallagher59701: On January 4, 2016 at 5:45:50 PM, Patrick West (notifications@github.com) wrote:

In dap/BESDapResponseBuilder.cc:

> @@ -80,11 +83,13 @@
>  #include <mime_util.h>      // for last_modified_time() and rfc_822_date()
>  #include <escaping.h>
>  #include <util.h>
> +#if USE_LOCAL_TIMEOUT_SCHEME
Are you using a macro here because the alarm doesn't work on some systems?

No. I had to competing timeout schemes in the code (here and in libdap::DDS) and I’ve replaced both with the new code. But, I’m hedging my bets by keeping the old code in the sources but excluding it from compilation using this #if. 

Thanks for looking at this. NB: I just wrote another test and pushed it (in module/debug_functions)

James


—
Reply to this email directly or view it on GitHub.


-- 
James Gallagher
jgallagher@opendap.org

dap/BESDapResponseBuilder.cc
```diff
@@ -1094,42 +1205,31 @@ void BESDapResponseBuilder::send_dap4_data_using_ce(ostream &out, DMR &dmr, bool
 
 void BESDapResponseBuilder::send_dap4_data(ostream &out, DMR &dmr, bool with_mime_headers)
 {
-    try {
```
mrpatrickwest: Is this part of the alarm branch? Or is it something else that probably should be in another branch?

dap/unit-tests/ResponseCacheTest.cc
```diff
@@ -42,6 +42,7 @@
 
 #include "BESDapResponseCache.h"
 #include "BESDapResponseBuilder.h"
+#include "TheBESKeys.h"
```
mrpatrickwest: seems there's a mix of bug fixes and timeout/alarm changes. I suggest branches be targeted to a specific task and not just generic "doing some stuff in a branch"

dispatch/BESInterface.cc
```diff
@@ -32,18 +32,18 @@
...
+#include <signal.h>
 #if HAVE_UNISTD_H
 #include <unistd.h>
 #endif
 
-using std::string;
-using std::ostringstream;
-using std::bad_alloc;
-using std::cout;
+#include <setjmp.h> // Used for the timeout processing
```
mrpatrickwest: should this be conditional based on if on windows system? I've seen the macros in other places but not here. Or is this a generic include?

mrpatrickwest:
Makes it very difficult to look through a change like adding timeout/alarm code when most of the changes are unrelated.

mrpatrickwest:
Travis-ci build is failing. I'm hoping the branch will be merged only after the build is successful.

jgallagher59701:
On January 4, 2016 at 6:09:31 PM, Patrick West (notifications@github.com) wrote:

Travis-ci build is failing. I'm hoping the branch will be merged only after the build is successful.

Yeah… It’s actually a problem with the libdap/bes branch incompatibility. There’s a bes-38 lib dap branch but the travis build doesn’t know to use it - it always uses lib dap master. But yes, it’ll build before I merge it.

Working backwards through your messages.


—
Reply to this email directly or view it on GitHub.


-- 
James Gallagher
jgallagher@opendap.org

jgallagher59701:
On January 4, 2016 at 6:08:48 PM, Patrick West (notifications@github.com) wrote:

Makes it very difficult to look through a change like adding timeout/alarm code when most of the changes are unrelated.

Yes, agreed but it takes me less time to work this way than to have two branches open and switch between them or make notes and cycle back to fix issues I find.


—
Reply to this email directly or view it on GitHub.


-- 
James Gallagher
jgallagher@opendap.org

dispatch/BESInterface.cc
jgallagher59701: On January 4, 2016 at 6:00:35 PM, Patrick West (notifications@github.com) wrote:

In dispatch/BESInterface.cc:

>  #if HAVE_UNISTD_H
>  #include <unistd.h>
>  #endif
>   
> -using std::string;
> -using std::ostringstream;
> -using std::bad_alloc;
> -using std::cout;
> +#include <setjmp.h> // Used for the timeout processing
should this be conditional based on if on windows system? I've seen the macros in other places but not here. Or is this a generic include?

AFAIK we don’t have a windows build for the BES. 


—
Reply to this email directly or view it on GitHub.


-- 
James Gallagher
jgallagher@opendap.org

dap/unit-tests/ResponseCacheTest.cc
jgallagher59701: On January 4, 2016 at 5:55:34 PM, Patrick West (notifications@github.com) wrote:

In dap/unit-tests/ResponseCacheTest.cc:

> @@ -42,6 +42,7 @@
>   
>  #include "BESDapResponseCache.h"
>  #include "BESDapResponseBuilder.h"
> +#include "TheBESKeys.h"
seems there's a mix of bug fixes and timeout/alarm changes. I suggest branches be targeted to a specific task and not just generic "doing some stuff in a branch"

I agree - but as noted this is how it unrolled this time


—
Reply to this email directly or view it on GitHub.


-- 
James Gallagher
jgallagher@opendap.org

dap/BESDapResponseBuilder.cc
jgallagher59701: On January 4, 2016 at 5:52:20 PM, Patrick West (notifications@github.com) wrote:

In dap/BESDapResponseBuilder.cc:

> @@ -1094,42 +1205,31 @@ void BESDapResponseBuilder::send_dap4_data_using_ce(ostream &out, DMR &dmr, bool
>   
>  void BESDapResponseBuilder::send_dap4_data(ostream &out, DMR &dmr, bool with_mime_headers)
>  {
> -    try {
Is this part of the alarm branch? Or is it something else that probably should be in another branch?

Yes, it’s part of the branch. One request from NASA is that the timeout not include the time needed to transmit the result. so the alarm is turned on in BESInterface (and turned off there as well in the generic case) but optionally turned off in the dap module once it starts sending data. This is far from optimal since it introduces a complex kind of coupling between the operation of the framework and the handler, but I don’t see a way around that.


—
Reply to this email directly or view it on GitHub.


-- 
James Gallagher
jgallagher@opendap.org

dap/BESDapResponseBuilder.cc
jgallagher59701: On January 4, 2016 at 5:49:44 PM, Patrick West (notifications@github.com) wrote:

In dap/BESDapResponseBuilder.cc:

> @@ -284,24 +272,104 @@ int BESDapResponseBuilder::get_timeout() const
>      return d_timeout;
>  }
>   
> +/**
> + * Turn on the alarm. The code will timeout after d_timeout
> + * seconds unless timeout_off() is called first.
> + *
> + * @deprecated
> + */
> +void
> +BESDapResponseBuilder::timeout_on() const
> +{
> +#if USE_LOCAL_TIMEOUT_SCHEME
> +#ifndef WIN32
> +    alarm(d_timeout);
We have a SignalHandler class that could handle some of this, if not all of this. Wouldn't have to worry about using the macros everywhere. timeout_on would call a method on the SignalHandler instance to turn on timeout, turn off timeout. The underlying code is encapsulated in the SignalHandler class.

SignalHandler (and AlarmHandler): I assume you mean the code in libdap are not really that great. They are one of those OOP ideas that never really panned out. My opinion (I copied those from Dr Dobbs quite some time ago) is that it’s better to make the basic unix stuff follow the common patterns. By regardless of that, this is about setting the stage for removing the code above. Note that USE_LOCAL… is #undef at the top of the file.


—
Reply to this email directly or view it on GitHub.


-- 
James Gallagher
jgallagher@opendap.org

jgallagher59701: Ugh. Build now fails a test on travis but not machine(s). I’ll have to look at this on a linux machine - and XML element error unrelated to the timeouts. 

James

On January 5, 2016 at 8:41:45 AM, James Gallagher (jgallagher@opendap.org) wrote:



On January 4, 2016 at 5:49:44 PM, Patrick West (notifications@github.com) wrote:

In dap/BESDapResponseBuilder.cc:

> @@ -284,24 +272,104 @@ int BESDapResponseBuilder::get_timeout() const
>      return d_timeout;
>  }
>    
> +/**
> + * Turn on the alarm. The code will timeout after d_timeout
> + * seconds unless timeout_off() is called first.
> + *
> + * @deprecated
> + */
> +void
> +BESDapResponseBuilder::timeout_on() const
> +{
> +#if USE_LOCAL_TIMEOUT_SCHEME
> +#ifndef WIN32
> +    alarm(d_timeout);
We have a SignalHandler class that could handle some of this, if not all of this. Wouldn't have to worry about using the macros everywhere. timeout_on would call a method on the SignalHandler instance to turn on timeout, turn off timeout. The underlying code is encapsulated in the SignalHandler class.

SignalHandler (and AlarmHandler): I assume you mean the code in libdap are not really that great. They are one of those OOP ideas that never really panned out. My opinion (I copied those from Dr Dobbs quite some time ago) is that it’s better to make the basic unix stuff follow the common patterns. By regardless of that, this is about setting the stage for removing the code above. Note that USE_LOCAL… is #undef at the top of the file.


—
Reply to this email directly or view it on GitHub.


-- 
James Gallagher
jgallagher@opendap.org
-- 
James Gallagher
jgallagher@opendap.org

PR #16, opened 2016-01-15
Besdebug ndebug

jgallagher59701:
I have been doing some performance measurements of Hyrax and one (small) thing that has bugged me for a while is that the BESDEBUG macros run no matter what. I did a quick analysis and found that for operations like 'get a DDS' the time for besstandalone to return the DDS from an HDF4-based aggregation is about 0.9s when none of the BESDEBUG names are set (so no debugging output). The time drops to about 0.5s when BESDEBUG is set to a null macro. There are performance improvements for Data access as well, but not 40%. Still... 

This branch contains a simple modification to the BESDEBUG macro so that when NDEBUG is set, the macro becomes a no-op (just as the old C assert() does. The --enable-developer option controls whether NDEBUG is defined or not.

James


ndp-opendap:
Yay! This has been a problem for a long time and the performance testing we did for Raytheon/NASA replay highlighted how much of a problem this has been. Thanks for patching this up, a much needed repair.
:shipit:

ndp-opendap:
This change is great for performance but it comes with a small caveat: If you have configured without using the *--enabled-developer* switch then the BESDBUG macros will not be compiled. Which is good, unless you are writing new ones in which case you may be writing garbage and the compiler will not mention it since your new calls are not being compiled. This shouldn't actually be an issue for most of us as we typically work with the code base in developer mode. But if for some reason that's not the case then beware.

mrpatrickwest:
Kind of a bummer that we won't be able to turn on and off debugging through configuration, but if it impacts performance that much then this is a good thing.

:shipit:

jhrg:
On January 29, 2016 at 09:04:21, Patrick West (notifications@github.com) wrote:

Kind of a bummer that we won't be able to turn on and off debugging through configuration, but if it impacts performance that much then this is a good thing.

Right. I was thinking that as a separate change some of the current BESDEBUG calls would be replaced by Verbose log calls (I’ve written two macros that make writing writing both log and verbose log entries more streamlined. I didn’t want to mix that change in with the BESDEBUG change, though.

James




—
Reply to this email directly or view it on GitHub.


-- 
James Gallagher
jhrg@mac.com

PR #19, opened 2016-03-22
Function result unwrapper

ndp-opendap:
This collection of changes allows the ugrid restrict functions to return DAP and NetCDF output that can be read ugrid clients.

These changes provide a more cohesive implementation of our unwrap hack in which server side functions that need to return collections of variables that appear at the top level of the resulting dataset can achieve this by wrapping said collection into a Structure with a magic name that gets intercepted downstream by the various Transmitters.

The following modules also contain code for this feature implementation:

* dap-server
* fileout_gdal
* fileout_json
* fileout_netcdf
* xml_data_handler

### DIFF
--- a/dap/BESDapModule.cc
+++ b/dap/BESDapModule.cc
@@ -69,6 +69,10 @@ using std::endl;
 #include "BESExceptionManager.h"
 #include "BESDapError.h"
 
+#include "DapFunctionUtils.h"
+#include "ServerFunctionsList.h"
+
+
 void BESDapModule::initialize(const string &modname)
 {
 	BESDEBUG("dap", "Initializing DAP Modules:" << endl);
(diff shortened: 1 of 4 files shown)


jgallagher59701:
+1

At this point I'd remove the #if 0 sections and also compact the empty lines. It may just be me, but I find the big open line parts slow down me down when I'm looking at the code and trying to understand it. Beyond that, I say 'merge.'

ndp-opendap:
OK I did what you suggested and I'll merge!

jgallagher59701:
On March 22, 2016 at 12:43:12, Nathan Potter (notifications@github.com) wrote:

OK I did what you suggested and I'll merge!

Woo Hoo.

NB: I Have a pull request out too! ;-)


—
You are receiving this because you commented.
Reply to this email directly or view it on GitHub


-- 
James Gallagher
jgallagher@opendap.org

PR #23, opened 2016-05-17
Response cache

jgallagher59701:
Adds a cache for function responses.

One thing that would be good would be to rename the main class/object to BESDapFunctionResponseCache.


ndp-opendap:
I think it would be prudent for us to run performance tests to compare this branch to the master branch (or even the most recent release) before we actually merge. This branch should represent a measurable performance increase. If it does not, then we should find out why and fix it, or abandon the changes. Using a cache always increases complexity and risk, and if the result doesn't include the hoped for performance gain there's no point in keeping the changes.

jgallagher59701:
rgr the performance tests.

I also want to 'resolve the conflicts'.

jgallagher59701:
Merged with master after resolving submodule conflicts (used git reset HEAD <submodule path>).

PR #24, opened 2016-05-24
Mem cache

jgallagher59701:
The ObjMemCache - now in bes/dap - implements a [memory cache](https://docs.google.com/a/opendap.org/document/d/1ewY9y2Uiiv4pBkl2msn-3BLUZW-QxfkCyOfWaxlLI7U/edit?usp=sharing). I've used this with the netcdf handler and while that's not the best test case, it does produce faster object access times (about 10+ times faster) with little measurable overhead. Unfortunately, the netcdf handler is a submodule, so you'll need to look at that code separately. See the mem_cache branch there. I may add this to the hdf4/5 handlers if there's time.

### DIFF
--- a/configure_modules.ac
+++ b/configure_modules.ac
@@ -844,10 +844,11 @@ AM_COND_IF([DAP_MODULES],
 	AC_CONFIG_FILES([modules/xml_data_handler/tests/generate_data_baseline.sh],
 		[chmod +x modules/xml_data_handler/tests/generate_data_baseline.sh])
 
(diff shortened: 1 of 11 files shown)


kyang2014:
So the memory cache is still not part of BES?

jgallagher59701:
On May 24, 2016 at 15:23:46, kyang2014 (notifications@github.com) wrote:

So the memory cache is still not part of BES?
It is part of the BES - it’s in bes/dap. Use git to checkout the branch ‘mem_cache’ to see the code or wait for me to merge it. Each handler has to be modified a little to use the cache.

James

PR #26, opened 2016-10-29
Geo functions

jgallagher59701:
the scale_array() and scale_grid() server functions now work for 2D data.


jgallagher59701:
I think I fixed the problems

PR #27, opened 2016-12-28
Hyrax 292

jgallagher59701:
Fix for the logrotate bug. This is more of a hack than a real fix.


jgallagher59701:
I'm going to test this on test.opendap.org over night and merge it the server survives there for 24 hours.

PR #28, opened 2017-01-06
Decoupled BESBasicInterface from BESXMLInterface

jjeichert:


jgallagher59701:
If we wrote a new interface that did not use XML, would it have to duplicate code you've just moved into BESXMLInterface? If that's the case, maybe it would be better to move the BASBasicInterface code back up into BESInterface and make that class concrete (or less abstract...)?

jjeichert:
Yes.  That's a good point. I'll have to look at that.

jjeichert:
Should I back out the changes made in BESXMLinterface and then move the BASBasicInterface into BESInterface?

PR #31, opened 2017-01-26
Refactor 1

jgallagher59701:
I've cleaned up a number of things that I saw during the timeout_bug work. BESDebug.h, TcpSocket, PPTStreamBuf, etc. I've tested this using builds both with and without --enable-developer (which might matter for changes to BESDebug) and all the tests pass.

Now I've updated the submodules so Travis will get the corrected files in those repos.


ndp-opendap:
:shipit: 
 🤘

PR #43, opened 2017-03-30
Clear alarms

ndp-opendap:
Added call to clear timeout alarms to the various transmitters. This was being done in BESDapResponder but the Transmitters were in need. Should fix https://opendap.atlassian.net/browse/HYRAX-354


ndp-opendap:
I updated the code to disable the alarm just prior to intern_data() activities. This seems to have it working nicely on my dev system. (tested +1)

PR #46, opened 2017-03-31
Dmr++

ndp-opendap:
This is the cloudydap proof of concept code that uses range GET to retrieve parts of files/objects held in AWS S3.


ndp-opendap:
Ouch! The Travis-CI build failed because curl_multi_perform is part of a more modern curl library than is available in CentOS-6ville. I built and tested the code on a CentOS-7 system for this very reason. So... Not sure how to proceed... Park the branch until later?

PR #47, opened 2017-04-03
Moved clear timeout code to BESUtil, added call to BESCatalogDirectory

ndp-opendap:
No big directories don't timeout


jgallagher59701:
merge

PR #49, opened 2017-04-11
Roi subset unwrap

ndp-opendap:
This should address https://opendap.atlassian.net/browse/HYRAX-342 by utilizing the "_unwrap" hack in the name of the roi_subset() returned Structure name.


jgallagher59701:
Lets merge it!

PR #52, opened 2017-04-15
Cache lifecycle

ndp-opendap:


ndp-opendap:
@jhrg   -

Can you have a look at this? All the tests pass for me and the log output doesn't seem to contain a build failure.

N

ndp-opendap:
@jhrg  - I just made a fresh CentOS-6 and tried it all there - tests pass but there may be issues depending on if Travis-CI does the build by running first _make_ and then _make check_ (which should be working) or if just goes straight to _make check_ (which I think can't work, see https://opendap.atlassian.net/browse/HYRAX-369 for more on what's happening there)

PR #56, opened 2017-04-20
Fixed a bug in BESDebug.h and reformatted BESStopWatch.cc

jgallagher59701:
Short and sweet - this fixes one problem where the bes debug context "timing" was noever enabled when --enable-debug was used. In fact, it was only eabled when it _was not_ used. I also removed some old code, but only in one place.


jgallagher59701:
I'll just merge this... it's a tiny bit of code.

PR #63, opened 2017-05-23
Distcheck fix

jgallagher59701:
This is a small collection of fixes for the distcheck target. I've added a _Debugging distcheck_ page on docs.o.o...


ndp-opendap:
If the tests pass then :shipit:

PR #86, opened 2017-10-26
Dmr++

jgallagher59701:
This code should now only build when --enable-developer is used.


ndp-opendap:
Seems like the file `DmrppArray.cc.tmp` should get dropped.

ndp-opendap:
Also the build failed in `make dist` with the error:
----
make[4]: Entering directory `/home/travis/build/OPENDAP/bes/modules/dmrpp_module/data'
make[4]: *** No rule to make target `endian_convert.h', needed by `distdir'.  Stop.
make[4]: Leaving directory `/home/travis/build/OPENDAP/bes/modules/dmrpp_module/data'
----
That directory has code in it that we want, but may not want to build as part of the `dist` target so maybe we exclude it?

ndp-opendap:
Yes since you managed to get it under control I think it's important to keep it in the loop. Are we sure any side branches have been merged in at this point? If so then :shipit:

jgallagher59701:
Merged but keeping the dmr++ branch for now since this was such a big effort with so many commits. After we go through the next release, then I'd feel comfortable deleting it.

PR #99, opened 2018-01-09
Hyrax 561

ndp-opendap:
This branch repairs the issue wherein a client requests a slow response (say a big data response from an aggregation as a netcdf file) and then the client exits before the response is transmitted. The issue is that the temporary file created by the fileout_netcdf handler was not being correctly removed in this circumstance. The TempFile class was reworked to utilize a specialized signal handler for SIGPIPE during the lifecycle of the TempFile class. Tests indicate that this is now repaired. Unit tests have been written for the new TempFile class and the old code has been retired.


jgallagher59701:
This branch passes distcheck on OSX. Lets see how it does on Ubuntu...

PR #100, opened 2018-02-09
Set contexts

jgallagher59701:
A new command that sets a number of contexts in one shot. This command also uses a Null Response Handler; it's a tiny bit more efficient than than running a bunch of set context commands.

This still needs some tests to be written, but I'd like to see if the CI build works.


jgallagher59701:
I'm going to merge this, given the level of refactoring and that there is a new and potentially useful command here.

PR #102, opened 2018-02-14
File locking cache tests

jgallagher59701:
Will this work? Can this trigger a failure on Ubuntu?


jgallagher59701:
That's a good idea - I think - to use the -f option. But, they are all using a different cache directory and I think that makes the times more predictable. The build failed because of the hyrax-deps taking too long... I think this is a travis issue we're running into.

PR #127, opened 2018-03-24
changed how configure looks for cppunit

captainkirk99:
This switches to the new way to find cppunit. See 
https://bugzilla.redhat.com/show_bug.cgi?id=1311694 for an explanation of why cppunit stopped supporting cppunit-config.

In this fix the same thing is accomplished with PKG_CHECK_MODULES.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -254,9 +254,10 @@ AC_ARG_WITH(libwrap,
     
 AC_HEADER_TR1_FUNCTIONAL
 
-AM_PATH_CPPUNIT(1.12.0,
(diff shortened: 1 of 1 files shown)


captainkirk99:
I believe we can also eliminate file  conf/cppunit.m4, but I need to get this package building and testing before I start removing files. ;-)

captainkirk99:
Looks like these failed Travis for the same AWS S3 reason that libdap4 failed.

jgallagher59701:
Thanks.

PR #128, opened 2018-04-01
Have configure check for libdap and fail if it is not found

captainkirk99:
Add checking for libdap4 library in configure. Fail if libdap is not found.


jgallagher59701:
"Add checking for libdap4 library in configure. Fail if libdap is not found."

That's OK for us, but technically, the BES can be used w/o libdap. So we may get pushback on requiring libdap. A heads up.

jgallagher59701:
I think this is a good idea, but I'm not ready to say that the BES can exist only with libdap.

PR #138, opened 2018-04-25
Dmrpp test refactor

jgallagher59701:
I moved tests (all the testing done in DmrppTypeReadTest and DmrppHttpReadTest) that were better coded using autotest into a new autotest suite. I also removed those tests from the unit-tests/Makefile.am.


jgallagher59701:
I should add: I did this because I need to be able to run tests that check reading using libcurl and have the entire handler loaded, including new classes I'm writing that will be singletons. That's hard to do with unit tests.

PR #147, opened 2018-05-24
Adding new BES command 'showBesKey'

ndp-opendap:
Seems to wrk, the response format is overly complex, but BESInfo is not really setup to produce my desired result so I am inclined to just get on with it.


ndp-opendap:
I fixed the copyright and attributions, but I would like to try and fix an API deficiency that prevents me from generating a simpler response document than the one I am producing now. I have looked at the BESInfo code and I think I see how to move it forward but I do have some questions.

PR #152, opened 2018-05-26
BESInfo API Addition

ndp-opendap:
Added a new begin_response() method to BESInfo that allows one to add attributes to the begin_response element header. Since the only part of the BESInfo ecosystem that is using this is using XML and thus BESXMLInfo I implemented it there.


jgallagher59701:
I say merge it!

PR #158, opened 2018-06-06
Mds xml base

jgallagher59701:
For this branch I decided that if the xml:base context is not set, the MDS should do nothing. This preserves the tests, which is more than just a convenience, since they test the MDS by reusing baselines from a set of tests that don't use the MDS - i.e., does the server return _exactly_ the same thing with and without the MDS? 

Since the OLFS should never omit the xml:base context, it should not be a problem. If it is, then we can add a testing mode.

### DIFF
--- a/dap/GlobalMetadataStore.cc
+++ b/dap/GlobalMetadataStore.cc
@@ -130,7 +130,7 @@ static void transfer_bytes(int fd, ostream &os)
     }
 }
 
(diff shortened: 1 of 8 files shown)


ndp-opendap:
mmmmmmmmm :shipit:

jgallagher59701:
Done.

PR #160, opened 2018-06-11
Fix for ticket HYRAX-775 (xml:base value fail)

jgallagher59701:
The code in the GlobalMetadataStore that inserts the value of xml:base
was failing because some responses were larger than a single buffer (16k)
and that caused the xml:base attribute to be inserted multiple times.


ndp-opendap:
If the tests pass you must :shipit:

PR #161, opened 2018-06-12
Fixed the DDX response so that it'll work when using the MDS.

jgallagher59701:


ndp-opendap:
When the tests they pass, you've got to :shipit:! :shipit: good!

PR #179, opened 2018-08-04
Ncml memory fixes

jgallagher59701:
Really a patch - There's a problem with the use of the Array copy constructor, but the NCML code doesn't need to copy the variables. Instead it can use the add_var_nocopy() methods of DDS and Array instead of the add_var() methods that trigger a copy. This patches the problem and is more efficient, but it means that the bug is still there waiting for just the right hack to baffle us once more...


jgallagher59701:
> On Aug 4, 2018, at 12:18, Nathan Potter <notifications@github.com> wrote:
> 
> Do ou know how to fix the general case? Did you capture that?
> 

I don’t know why the copy ctor is failing. But not using it definitely fixed the error. 

My guess is that both of these ‘join’ operations are changing the size of the array after the internal buffer is set. Thus the copy ctor is moving , eg, 20 bytes from source to dest, but source was originally only 10 bytes. Now it’s 20 bytes in size, but is still using the 10 byte buffer. 

I’ll try to find that, but first a bike ride...

PR #180, opened 2018-08-07
Test for Ncml Aggregations using the DMR++ handler

jgallagher59701:
Both local file and data from S3 are tested - we might have to
remove the S3 tests at some point (when the bucket goes away).
For now, this shows it really works reading from S3.


ndp-opendap:
We could use test.opendap.org instead of S3 in the tests... HTTP and all that...

jgallagher59701:
Will add a test using test.o.o's HTTP

PR #186, opened 2018-08-23
Dmrpp profile

jgallagher59701:
Ready for a merge? This code is not complete, in the sense that all of the optimizations have been performed on all of the read() methods, but it's correct and much faster than before. I think we should merge it now and open a second branch to continue the optimization work. That will entail modifying the read_chunks() code to mirror the changes made to read_chunks_unconstrained() and to improve parallelism in read_chunks_contiguous().


jgallagher59701:
I'd like to merge this and then run it through coverity. The code might not be complete (in the sense that I have more opt work to do), but it runs.

jgallagher59701:
Huh? A test failed - it's a show node test in bes/xmlcommand. I'll try to sort this out today...

ndp-opendap:
passed tests now...

:shipit:

PR #190, opened 2018-09-06
Hyrax 1.15

ndp-opendap:
So far this branch:

1. Remedies the the requirement that ALL catalogs have an accessible root directory on the local file system
1. Updates CmrCache so that is sha2 hashes resource names to be used as cache file names (thus limiting their size to something acceptable by most filesystem variants)


ndp-opendap:
I pulled your preloaded MDS cache commit and am running it on 52.23.165.57:8080 now.

PR #191, opened 2018-09-08
RC1

jgallagher59701:
I think this is ready... We'll see what linux says.


jgallagher59701:
This branch has both changes I made while building on CentOS 7 for the timing tests and changes made while (trying to) build(ing) on CentOS 6. It now compiles with and without enable-developer. I think this should be merged to master and then master merged into hyrax-1.15.

PR #194, opened 2018-09-19
Lots of changes; I'm not sure I understand why these were needed

jgallagher59701:
...maybe rpmbuild changed.

This should not affect the CI build. but the Makefile.am files are different.


jgallagher59701:
I want to hold off on merging this because I'm not sure about C6 and C7 RPMs. Slav is working on C6.

PR #195, opened 2018-09-27
Debian

kneumiller:


jgallagher59701:
I have an idea WRT fixing the build. In the rules file you have:
override_dh_auto_install:
	$(MAKE) DESTDIR=$$(pwd)/debian/bes install

That overrides 'make install.' if there is an equivalent for 'make all,' or 'make check,' whichever one causes the problem with /usr/share/mds, then you can provide a value for DESTDIR and I that might get '/usr/share/mds' to be $(DESTDIR)/usr/share/mds.

Something like:

override_dh_auto_check
        $(MAKE) DESTDIR=$$(pwd)/debian/bes check

Also, for those, if you add $(MFLAGS) it might speed things up, since flags like -j9 and so on can then be passed into the child make processes.

PR #202, opened 2018-10-29
Hk246 bbox comb

vskorole:


jgallagher59701:
We need to fix the tests before merging this branch. Here is the log: https://api.travis-ci.org/v3/job/447967172/log.txt

vskorole:
Found 
637: Comparing dods response for dmrpp/s3_dmrpp_join_new.ncml to baseline ${abs_srcdir}/baselines/dmrpp/s3_dmrpp_join_new.ncml  FAILED (dmrpp_tests.at:13)

functions/BBoxCombFunction.cc
```diff
@@ -0,0 +1,136 @@
...
+    unsigned int rnk1= 0;
+    unsigned int rnk2 = 0;
+
+    switch (argc) {
+    case 0:
+    case 1:
+        // Must have 2 arguments
+        throw Error(malformed_expr, wrong_args);
+
+    default:
```
jgallagher59701: Use 'case 2:' and not default and have default: throw an exception. As the code is now, the number of args can be 3, 4, ..., although only the first two will be used.

functions/BBoxCombFunction.cc
```diff
@@ -0,0 +1,136 @@
...
+#include "BBoxCombFunction.h"
+#include "roi_util.h"
+
+using namespace std;
+using namespace libdap;
+
+namespace functions {
+
+/**
+ * @brief Combine two bounding boxes with different shapes, forming their combination.
```
jgallagher59701: shorten @brief to just "Combine two bounding boxes with different shapes"

functions/BBoxCombFunction.cc
```diff
@@ -0,0 +1,136 @@
...
+ *
+ * This combines 2 BBox variables (Array of Structure) forming
```
jgallagher59701: Change to:
This combines to bounding boxes (Array of Structure) of rank N and M into a single bounding box of rank N+M.

functions/unit-tests/BBoxCombFunctionTest
```diff
@@ -0,0 +1,210 @@
+#! /bin/sh
```
jgallagher59701: This is a generated file. It should not be in git

jgallagher59701:
oh, this should be merged!

PR #212, opened 2018-12-14
Repair misclassified read error in the stream handler

ndp-opendap:


jgallagher59701:
If this passes, merge it.

PR #213, opened 2018-12-20
Feedback url add

jgallagher59701:
This branch adds a new feature - if the BES.AnnotationServiceUrl parameter is set, then an attribute named AnnotationService is included in the DODS_EXTRA container for every DAS/DDX/DMR returned by Hyrax. The intent is that this can be used to access an annotation service and provide 'feedback/annotation' for a given URL. 

This feature is added in bes/dap in the ResponseHandler classes for DAS, DDS, DDX and DMR. There are tests in bes/dap/tests.


ndp-opendap:
:shipit: woot woot

PR #216, opened 2018-12-21
Container store refactor

jgallagher59701:
Cleaned up the code around the SetContainer command. Found that the persistence 'default' that was bound to the ContainerStorageVolatile object was not ever used and that the handlers install the ContainerStorageCatalog are doing the right thing (albeit with a confusing name - 'catalog'). As a result I removed the ContainerStorageVolatile from the default module startup code. 

Along the way a few other things were made simpler and maybe a bit faster.

We could also refactor the DefinitionStorageVolatile out of the code since the DefinitionStorage is pure abstract and there's only one child. But that's for another refactor.


ndp-opendap:
nice. simple is often better. :shipit:

PR #226, opened 2019-01-14
Hk 95

ndp-opendap:
I added single quotes around various string values in the ncml error messages ao that it is easier to tell when there are empty strings.


jgallagher59701:
Looks good; merge

PR #227, opened 2019-01-18
Httpd minor refactor

jgallagher59701:


ndp-opendap:
This build is failing because:
```
Error during module initialization: BES: Invalid entry >>>>>>> master in configuration file /home/travis/install/etc/bes/modules/httpd_catalog.conf:  '=' character not found.

Could not initialize the modules to get the log contexts.
Developer Mode: Not setting group or user ids
FAILED: BES: Invalid entry >>>>>>> master in configuration file /home/travis/install/etc/bes/modules/httpd_catalog.conf:  '=' character not found.
```
I looked at the **httpd_catalog.conf.in** file and I noticed that you added spaces between the "=" signs and the keys and values:

```
HttpResourceCache.dir = /tmp/httpd_cache
```
&lt;red-herring&gt;
 I maybe wrong (wouldn't be the first time), but if I remember correctly that will not work. The parser is not going to handle the spaces. I think PWest told me this a long time ago...
&lt;/red-herring&gt;

ndp-opendap:
:shipit:

PR #228, opened 2019-01-24
Site conf proto

ndp-opendap:
Added prototype site.conf file


jgallagher59701:
Well, we really should not be building for this....

jgallagher59701:
I made some spelling edits, then tried to commit as [skip ci] but that failed...

jgallagher59701:
The RPMs built - I think we should merge this.

jgallagher59701:
Should we merge it?

jgallagher59701:
OK

PR #230, opened 2019-01-25
Hk 313

jgallagher59701:
Now it works


jgallagher59701:
And it has a test for access to remote resource - I added some more, but they depend on the hdf5 handler and I'd rather not include that because a broken hdf5 then leads to cascading fails. One hdf5 test and baseline are in github, although commented out of testuite.at. I added some notes about configuration to the README.

PR #231, opened 2019-01-30
Unlinking cache files when retrival from remote fails.

ndp-opendap:
This patch should remove cache files associated with failed attempts to retrieve remote resources. This affects *cmr_module*, *gateway_module*, and *httpd_catalog_module*.

### DIFF
--- a/modules/cmr_module/RemoteHttpResource.cc
+++ b/modules/cmr_module/RemoteHttpResource.cc
@@ -192,7 +192,14 @@ void RemoteHttpResource::retrieveResource()
         if (cache->create_and_lock(d_resourceCacheFileName, d_fd)) {
 
             // Write the remote resource to the cache file.
-            writeResourceToFile(d_fd);
+            try {
+                writeResourceToFile(d_fd);
+            }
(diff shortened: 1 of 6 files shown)


ndp-opendap:
I think we should look at this one more time. Here's my thought: 

Both _cmr_module_ and _httpd_catalog_module_ create two cache files, one for the retrieved object (primary cache file)  and one for the http headers (aka header file). When the retrieval fails (via an exception)  the cache file is unlinked and the exception is re-thrown. (At that point the header file has not yet been created.)

Similarly, when the header file is created and written if an exception is encountered the header file is unlinked and the exception is re-thrown. I think this will leave the primary cache file (already written) in place without it's header file.

QUESTION: If creating the header file encounters an exception, should we unlink  _both_ the primary cache file and the associated header file?

Easy to fix if we choose. Let's have a look at this today.

jgallagher59701:
Yes, the cache(s) should treat these writes as one atomic operation, so if one fails, both should be treated as failures and neither file should be in the cache.

PR #233, opened 2019-01-31
Attempt to repair support for systemctl/systemd found on CentOS 7

jgallagher59701:


ndp-opendap:
woot! let's hope that makes things more stable.

PR #235, opened 2019-02-06
Added sonar cloud scan to the BES

jgallagher59701:
Lets try this. Might not work for PRs


ndp-opendap:
Hmmm:
```
ERROR: 

The only way to get an accurate analysis of C/C++/Objective-C files is by using the SonarSource build-wrapper
and setting the property "sonar.cfamily.build-wrapper-output", but it was not specified.

If you don't want to analyze C/C++/Objective-C files, then prevent them from being analyzed by setting the following properties:
```
Seems like it needs its special compiler gizwat...

jgallagher59701:
Lets not merge this... I want to find a better way to integrate the build results into travis.

PR #236, opened 2019-02-13
Sonar cloud2

jgallagher59701:
New travis code plus some fixes


ndp-opendap:
Hmmm looks like the sonar cloud stuff needs some adjustment to it's pass/fail metric

jgallagher59701:
I should say that this branch not only contains fixes, etc., base on feedback from sonar scan, it also contains a blue print for breaking a large scan into parts are running them in parallel.

jgallagher59701:
This branch's error was some sort of AWS timing fail. I'm going to merge since I restarted the build and it worked.

PR #240, opened 2019-03-09
Hk 352

vskorole:


jgallagher59701:
Lets find time on Monday to make sure I understand the changes, especially the affect on the returned data when we go from Float32 to Byte. I can see how it will make the data more accessible, and that 's a good thing.

PR #241, opened 2019-03-20
Hk 351

vskorole:


jgallagher59701:
I should write tests for the 'feature' I added, but  quick look with QGIS shows that they are tiff files but I didn't see any data values. Regardless, both Preview and QGIS did open them, so they are recognizable as Tiff files. I also added the jp2k tests back into the build. I removed those tests a long time ago because they did not work. They do now (and the resulting files look good as well).

I think we should merge this if it passes the CI build.

PR #242, opened 2019-03-20
Response limit fix

jgallagher59701:
Fixes HK-22 - Response Limit not working. This works with besstandalone, bescmdln and the OLFS (incl the IFH). The problem was a mixup in the way the size was recorded versus computed combined with differences in the way the parameter value was processed by the DAP2 and DAP4 implementations.


ndp-opendap:
Looks good. The `#if 0 ... #endif` blocks don't have any comment/reason they're disabled, which, while obvious at the moment may become less so with time. Just sayin...

:shipit:

jgallagher59701:
I'll fix the comment blocks and merge.

PR #243, opened 2019-03-28
C++11 support

jgallagher59701:
This needs a special branch of the hdf4_handler code (typeof-removal) to build. The hash for that starts with 3bbeae.


ndp-opendap:
do we need a pull request for hdf4 also?

jgallagher59701:
This code is not ready. It needs work in at least the hdf5 handler to build in g++ with --std=cxx11

PR #246, opened 2019-03-28
Covjson patch take2

jgallagher59701:


jgallagher59701:
Kent looked this over, as fi fI. He agrees we should leave the test in place. I'll ask the CF experts what we should do about LongName.

PR #249, opened 2019-04-16
Added a fix from Aron.Bartle@mechdyne.com.

jgallagher59701:
Netcdf file responses were not compressed when they should have been
if the dataset had Structures. Fixed by patching FONcStructure as Aron
suggested.


ndp-opendap:
:shipit:
!!

PR #250, opened 2019-04-16
Stare indexing

kneumiller:


functions/stare/Makefile.am
```diff
@@ -0,0 +1,43 @@
...
+
+# Headers in 'tests' are used by the arrayT unit tests.
+
+#AM_CPPFLAGS = $(BES_CPPFLAGS) -I$(top_srcdir)/dispatch -I$(top_srcdir)/dap -I$(top_srcdir)/functions $(DAP_CFLAGS) $(GDAL_CFLAGS)
+#AM_LDADD =  $(DAP_SERVER_LIBS) $(GF_LIBS) $(top_builddir)/dispatch/libbes_dispatch.la
+
+
+#Define __macosx for STARE
+#Remove hard code for filepath
+AM_CPPFLAGS = $(DAP_CFLAGS) -D__macosx -I$(top_srcdir)/../hyrax-dependencies/src/STARE/include -I$(prefix)/deps/include
```
jgallagher59701: This line right here is  problem. Don't use the hyrax-dependencies/src/ directory tree. The BES should only be using STARE headers that installed. If the STARE headers are going in include/STARE, the add -I$prefix)/deps/include/STARE

jgallagher59701: This line and the one below need to be changed. The two values of -I are both not going to work in general. For most builds $prefix is not set and hyrax-dependencies won't be anywhere predictable.

kneumiller:
We'll have to wait for the changes I made to the STARE CMake files to go through in the original repository before we can see if the changes I made here will work.

jgallagher59701:
Looks good. If the tests pass, lets merge this. We'll need to sort out the representation issue - HDF5 Compound array or Compound with arrays or just plain arrays (aka the EnableCF issue). Once that's done, we should work with Mike to write some tests. But this can be merged as is if the build works.

jgallagher59701:
I'll leave this branch open since another change is pending...

PR #251, opened 2019-04-17
Hk 128

vskorole:


vskorole:
> Are there tests for this new feature? Also, are we (or please make sure to) test for the case where the regex is turned on, but none match and there is a format file that works using the one-to-one basename string match (i.e., the way the handler works normally).

I left one bescmd in tests but for sure I will add tests for cases you mentioned.

jgallagher59701:
Lets merge this.

PR #257, opened 2019-05-23
Hacked the stare code in configure so it'll detect the STARE lib

jgallagher59701:


jgallagher59701:
The failures are build failures on the master branch, which does not have the updated STARE library build code. There's a bug in the Travis yml in that it clones the BES on master when we're running a package from a PR. That's something for Jira.

PR #263, opened 2019-06-21
Bes leak fix

jgallagher59701:
This is an improvement over the previous state where each call to the libxml2-based SAX parser leaked 256 bytes. But it's not perfect - there's still memory from libcurl being leaked and I think it is libxml2 using libcurl that is causing these leaks.


jgallagher59701:
There's still a leak from libcurl which calls libnsspem, and I know a similar (identical?) leak was fixed when I fixed the DMR and DMR++ parsers. But that's not happening here. Maybe we need to switch to the push pull context parser, but that seems wrong somehow. For now this fixes some of the libxml-related leaks.

ndp-opendap:
Looks good! :shipit:

PR #273, opened 2019-08-16
Whitelist file patch

ndp-opendap:
**_I think this is ready now_**
I am opening this pull request prior to resolving the issue, primarily because I don't want to loose track of this. I think the old code, in `Whitelist.cc `, that handled the whitelisting of file URLs was simply wrong (because the semantics of the c++ `string::compare()` method are not even remotely equivalent to boolean). This initial check-in is somewhat better but does not correctly handle the case when the `catalog_root` is longer the the file path.

The way this should work is that all file URLs must begin with the `catalog_root` path or be rejected. So clearly, if the `file_url` is shorter than the `catalog_root` then it must be rejected.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,7 @@ branches:
   - coverity_scan
 
 addons:
+  ssh_known_hosts: www.opendap.org
   coverity_scan:
     project:
       name: "OPENDAP/bes"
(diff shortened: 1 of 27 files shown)


jgallagher59701:
Should This be merged?

PR #276, opened 2019-09-25
Pthread improvements

kneumiller:
Updated the code to change the number of child chunks generated based on the size of data read in.

### DIFF
--- a/modules/dmrpp_module/DmrppArray.cc
+++ b/modules/dmrpp_module/DmrppArray.cc
@@ -374,7 +374,7 @@ void DmrppArray::read_contiguous()
     // If we want to read the chunk in parallel. Only read in parallel above some
     // threshold. jhrg 9/21/19
     // Only use parallel read if the chunk is over 2MB, otherwise it is easier to just read it as is kln 9/23/19
-    if (DmrppRequestHandler::d_use_parallel_transfers && master_chunk_size > 2 * (1024*1024)) {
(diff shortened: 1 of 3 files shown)


modules/dmrpp_module/DmrppArray.cc
```diff
@@ -35,6 +35,7 @@
 #include <cerrno>
 
 #include <pthread.h>
+#include <math.h>
```
jgallagher59701: It's better to use <cmath> than <math.h>

modules/dmrpp_module/DmrppArray.cc
```diff
@@ -368,45 +369,50 @@ void DmrppArray::read_contiguous()
...
+    if (DmrppRequestHandler::d_use_parallel_transfers && master_chunk_size > DmrppRequestHandler::d_min_size) {
 
         // Allocated memory for the 'master chunk' so the threads can transfer data
         // from the child chunks to it.
         master_chunk.set_rbuf_to_size();
 
+		// The number of child chunks are determined based on the size of the data.
+        // If the size of the master chunk is 3MB then 3 chunks will be made. We will round down
+        //  when necessary and handle the remainder later on (3.3MB = 3 chunks, 4.2MB = 4 chunks, etc.) kln 9/23/19
+        unsigned int num_chunks = floor(master_chunk_size/(1024*1024));
```
jgallagher59701: Make the 1MB constant value a cont unsigned int or # define.

PR #279, opened 2019-10-02
Combined log

ndp-opendap:
This PR creates the option of creating a combined BES/OLFS log (in the BES log file). The newly added BES code checks for the presence of the OLFS log content (in the request document) and when found it builds a combined log line with both the BES and OLFS log information. The OLFS and it's configuration determine if the information is sent to the BES, and if it is not sent the BES produces the same log entries as prior to this addition.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -221,6 +221,8 @@ before_deploy:
 
 # The deploy section copies the snapshot build product our S3 bucket and to www.opendap.org
 deploy:
+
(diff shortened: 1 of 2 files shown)


ndp-opendap:
> Add words to the docs (README/NEWS/AsciiDoc?) stuff about mods to the OLFS controlling information in the bes.log.

README and NEW will be updated throughout he release process. 
AsciiDoc??? Do you mean the Hyrax Guide?

PR #301, opened 2019-12-02
HK-530, make the logic clear and turn off the expected HDF5 dataset open error message

kyang2014:
The change is small. The original code is within #if 0 #endif block.
The main part is to turn off the HDF5 error message when an HDF5 dataset  fails to open under the case that the code cannot find the fullnamepath attribute for this variable.
Tested at CentOS 7.

### DIFF
--- a/modules/dmrpp_module/build_dmrpp.cc
+++ b/modules/dmrpp_module/build_dmrpp.cc
@@ -428,6 +428,38 @@ static void get_chunks_for_all_variables(hid_t file, D4Group *group)
         // relies on the varaible's position in the DAP dataset hierarchy.
         D4Attribute *attr = d4_attrs->get("fullnamepath");
         string FQN;
+        // I believe the logic is more clear in this way: 
+        // If fullnamepath exists and the H5Dopen2 fails to open, it should throw an error.
+        // If fullnamepath doesn't exist, we should ignore the error as the reason described below:
+        // (However, we should supress the HDF5 dataset open error message.)  KY 2019-12-02
+        // It's not an error if a DAP variable in a DMR from the hdf5 handler
+        // doesn't exist in the file _if_ there's no 'fullnamepath' because
+        // that variable was synthesized (likely for CF compliance)
+        hid_t dataset = -1;
+        if(attr) {
+            if(attr->num_values() == 1) 
+                FQN = attr->value(0);
+            else
+                FQN = (*v)->FQN();
+            VERBOSE(cerr << "Working on: " << FQN << endl);
+            dataset = H5Dopen2(file, FQN.c_str(), H5P_DEFAULT);
+            if(dataset <0) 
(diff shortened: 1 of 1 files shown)


modules/dmrpp_module/build_dmrpp.cc
```diff
@@ -428,6 +428,38 @@ static void get_chunks_for_all_variables(hid_t file, D4Group *group)
...
+        if(attr) {
+            if(attr->num_values() == 1) 
+                FQN = attr->value(0);
+            else
+                FQN = (*v)->FQN();
+            VERBOSE(cerr << "Working on: " << FQN << endl);
+            dataset = H5Dopen2(file, FQN.c_str(), H5P_DEFAULT);
+            if(dataset <0) 
+                 throw BESInternalError("HDF5 dataset '" + FQN + "' cannot be opened.", __FILE__, __LINE__);
+
```
jgallagher59701: Extra blank line

modules/dmrpp_module/build_dmrpp.cc
```diff
@@ -428,6 +428,38 @@ static void get_chunks_for_all_variables(hid_t file, D4Group *group)
...
+        // It's not an error if a DAP variable in a DMR from the hdf5 handler
+        // doesn't exist in the file _if_ there's no 'fullnamepath' because
+        // that variable was synthesized (likely for CF compliance)
+        hid_t dataset = -1;
+        if(attr) {
+            if(attr->num_values() == 1) 
+                FQN = attr->value(0);
+            else
+                FQN = (*v)->FQN();
+            VERBOSE(cerr << "Working on: " << FQN << endl);
```
jgallagher59701: VERBOSE will write to the log file. I think this should be BESDEBUG. Neither of those should include 'cerr' since they write to streams set elsewhere.

modules/dmrpp_module/build_dmrpp.cc
```diff
@@ -428,6 +428,38 @@ static void get_chunks_for_all_variables(hid_t file, D4Group *group)
...
+            if(dataset <0) 
+                 throw BESInternalError("HDF5 dataset '" + FQN + "' cannot be opened.", __FILE__, __LINE__);
+
+        }
+        else {
+            // The current design seems to still prefer to open the dataset when the fullnamepath doesn't exist
+            // So go ahead to open the dataset. Continue even if the dataset cannot be open. KY 2019-12-02
+            H5Eset_auto2(H5E_DEFAULT,NULL,NULL);
+            FQN = (*v)->FQN();
+            VERBOSE(cerr << "Working on: " << FQN << endl);
```
jgallagher59701: Same as above, make this BESDEBUG.

kyang2014:
Change the VERBOSE to BESDEBUG. Find that  there are quite a few VERBOSE and cerr lines in the build_dmrpp.cc file. Submitted a ticket (HK-543) to handle those in the future.

PR #302, opened 2019-12-12
HK-551, the attrs->append_attr() call changes the attrs in a for loop…

kyang2014:
… of attrs iterator. This causes the segmentation fault of when checking the next iterator of the original attrs, which doesn't exist any more. Since the next line of this call writes the attribute to the netCDF file, appending this attribute to attrs is not necessary. So just comment out this line.

After fixing this, the program runs and exits normally.

### DIFF
--- a/modules/fileout_netcdf/FONcAttributes.cc
+++ b/modules/fileout_netcdf/FONcAttributes.cc
@@ -373,7 +373,7 @@ void FONcAttributes::add_attributes_worker(int ncid, int varid, const string &va
                 string new_name_fillvalue = "Orig_FillValue";
                 BESDEBUG("fonc",
                          "FONcAttributes::add_attributes_worker - New attribute value is original value: " << val.c_str() << endl);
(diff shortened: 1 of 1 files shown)


kyang2014:
The for loop is in the same FoNCAttribute.cc,    for (; i != e; i++) {  } listed below

AttrTable::Attr_iter i = attrs.attr_begin();
        AttrTable::Attr_iter e = attrs.attr_end();
        for (; i != e; i++) {

BESDEBUG("fonc","add_attributes(): before calling get_attr_num "<<endl);
            unsigned int num_vals = attrs.get_attr_num(i);
BESDEBUG("fonc","add_attributes(): after calling get_attr_num "<<endl);


//KENT
            if (num_vals) {
                add_attributes_worker(ncid, varid, var_name, attrs, i, prepend_attr);
            }
        }

kyang2014:
Thanks James. I've added comments and #if 0 #endif block as you suggested. Please check again and merge the request.

jgallagher59701:
Thanks! Please merge.

PR #303, opened 2019-12-16
Signed s3 urls

ndp-opendap:
This PR adds BES support for multiple credentials sets that can be used when accessing other systems. More work remains to allow secure credentials storage/ingestion.


jgallagher59701:
Should we merge this code given it has passed the CI build?

ndp-opendap:
Yes! :shipit:

PR #304, opened 2019-12-19
Remove the remote access credentials from the bes.conf foodchain

ndp-opendap:
* Implement `kvp_utils` from `TheBESKeys`, make tests based on `keysT` (done)
* Convert `TheBESKeys` and the `CredentialsManager` to use the new `kvp_utils`.

When the PR was opened `kvp_utils` was implemented and passing tests


ndp-opendap:
The tests are failing because of conflict between code that tests file permissions and the way that git (fails) to preserve any permission besides the execute bit. All files checked out from git are set to either 644 or 755 (depending on the execute bit).

I tried adding calls in TravisCI to set the permissions on the file prior to testing but had no luck.

We should discuss if a solution can be found that allows us to get the permissions set correctly, or failing that, if we can/should make the test a friend that can access a private method that skips the file permissions step.

ndp-opendap:
Test pass! Woot!

PR #306, opened 2019-12-27
Env creds

ndp-opendap:
Adds a reserved value for `CredentialsManager.config` in the `bes.conf` chain that triggers the server to ingest access credentials from the env.


jgallagher59701:
If it passes, merge. This will cover the security gripe about env vars. (It will be interesting to see if SonarScan catches the use of environment variables...).

jgallagher59701:
Merge

PR #316, opened 2020-02-01
Fixed a memory leak in the AWS v4 sig code (HK-574)

jgallagher59701:


jgallagher59701:
This may actually be a duplicate of PR 314 which I just merged!

jgallagher59701:
This issue was fixed, and in a better way, by PR 314.

PR #318, opened 2020-02-13
HK-450

vskorole:
I should replace test tiff files in fileout_gdal module because they changed. Here is comparison result for one file:
_python gdalcompare.py coads_climatology.nc.0.bescmd.tif stdout_coads_climatology.nc.0.tif_
Files differ at the binary level.
Differences Found: 1
From GDAL doc:
"Image pixels, and various metadata are checked. There is also a byte by byte comparison done which will count as one difference. So if it is only important that the GDAL visible data is identical a difference count of 1 (the binary difference) **should be considered acceptable**."


jhrg:
> On Feb 14, 2020, at 10:36 AM, Slav Korolev <notifications@github.com> wrote:
> 
> I should replace test tiff files in fileout_gdal module because they changed.

OK.

ndp-opendap:
**STOP!**
I don't think this should be merged unless we are committed to the libdap4 API changes in HK-450, which I think we are not.  I certainly have concerns regarding the changes on `bes::HK-450` and `libdap4::shared_dim_fix` and if the HK-450 code should be merged to master ever. 

If the goal is to update the GDAL test targets then I think this should be handled on a separate child branch of `bes::master`, merged to `bes::master` when the tests pass, and then merged from `bes::master` to `bes::HK-450` once the master branch is passing tests

vskorole:
Yes, you are right. Need to merge HK-450 in libdap first. Sorry.

PR #320, opened 2020-02-18
Modified handler_tests_macros.m4 so it works in places with 4 letter time zones

jgallagher59701:
New Zealand has a four-letter time zone. This also includes a fix for two tests in the httpd_catalog
testsuite. We added a directory to test.opendap.org/data and that changed the response from
the remote httpd.


jgallagher59701:
It's a small change and I think it will pave the way for Kent and Slav's PRs to pass once they merge with this code.

PR #322, opened 2020-02-27
Added optional dataset inventory checking to the

ndp-opendap:
Update the `get_dmrpp` program to allow for optional testing of the dmr++ produced.


ndp-opendap:
This was failing because when I add test files to test.opendap.org/data I broke a bunch of httpd_catalog_module tests. fixed now, and merged into this branch

PR #323, opened 2020-02-28
Fixed baselines in httpd_catalog_module

ndp-opendap:
Fixed baselines in httpd_catalog_module that were broken by changes on test.o.o.
Repaired bug in RemoteHttpResource.cc where empty header lines or incorrectly constructed
headers were causing exceptions to be thrown. This is not detected and the offending line is skipped.


ndp-opendap:
Tests passed, I need to merge this ASAP (because tests are broken without it) so that's what I am doing.

PR #324, opened 2020-03-04
Automated array value testing in dmr++ production.

ndp-opendap:
**Adding automated array value testing to `get_dmrpp`**

This adds a simple, optional value test to the testing phase of `get_dmrpp`. 

- The DMR is checked for arrays of Float32
- One is picked at random from the list (assumes there is one)
- Counts the dimensions of the array
- builds a dap2 constraint expression that asks for the origin corner [0][0]...
- Constructs BES command files using the constraint expression.
-  Utilizes `besstandalone` to run the commands and request the DAP2 data response from both
  - The source data file
  - The dmr++ file that was constructed upstream by the script
- The results are compared using `cmp`

**Rudimentary Timing:**
- Just build dmr++:   ~0.2 s
- Build and inventory check:  ~1.2 s
- Build and value check:  ~2.6 s
- Build, inventory check, value check:   ~4.0 s

**Needs:**
- The code only tries for a Float32 array. I believe to needs to be more flexible, trying (all of the?) other atomic numeric types if the first does not have a manifestation in the dataset
- It should not fail if no arrays are located.
- Alternately/Additionally we could make a command line switch to have the user pass in the type of array to utilize. 

**Questions:**
- Should it compute the approx. center of the array and ask for that value? Dan observed that the periphery of a lot of science data is fill values.
- Should it have an option take N random samples from the dataset and compare them?


jgallagher59701:
**Yes**: Alternately/Additionally we could make a command line switch to have the user pass in the type of array to utilize.

**Yes**: Should it compute the approx. center of the array and ask for that value? Dan observed that the periphery of a lot of science data is fill values.

**Maybe, I think you get not so much benefit from this, plus you have to take into account different sizes, etc.**: Should it have an option take N random samples from the dataset and compare them?

I'll look at the code next...

jgallagher59701:
I think it looks good. Merge!

PR #328, opened 2020-03-09
Ngap

ndp-opendap:
NGAP PI-20


jgallagher59701:
Merge

modules/dmrpp_module/CredentialsManager.cc
```diff
@@ -25,6 +25,7 @@
 
 #include "config.h"
 
+#include <stdlib.h>
```
jgallagher59701: #include <cstdlib>

jgallagher59701: `Use #include <cstdlib>`

ndp-opendap: done

PR #329, opened 2020-03-10
Ngap hyrax 88

sawer297:
> 1. Code Cleanup
> - clean/groom code making it conform to nominal OPeNDAP code style. <-- ( Done SBL )
> - Add explanatory comments <-- ( Done SBL )
> - Remove unused portions of code <-- ( Done SBL )

cleaned code, added comments, and removed unused code

> 2. Implement thoughtful control over printing info to cout & cerr - in particular using stdout only for dmr++ output and Kerr for everything else. <-- ( Done NDP)

believe nathan when through and fixed this already

> 
3. Run through valgrind to check for memory and lost allotted space checking <-- ( Done SBL )

ran valgrind on the ngap_build_dmrpp.cc using a .h5 file in the data directory
the output from that run is in the Earthdata Jira ticket Hyrax-88

### DIFF
--- a/modules/dmrpp_module/.gitignore
+++ b/modules/dmrpp_module/.gitignore
@@ -8,5 +8,3 @@
 /build_dmrpp
 /ngap_build_dmrpp
 /bes_default_conf.h
-/foo.dmrpp
-/airs.dmrpp
--- a/modules/dmrpp_module/.gitignore
+++ b/modules/dmrpp_module/.gitignore
@@ -9,3 +9,4 @@
 /ngap_build_dmrpp
 /bes_default_conf.h
 /foo.dmrpp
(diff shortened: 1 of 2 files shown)


modules/dmrpp_module/.gitignore
```diff
@@ -9,3 +9,4 @@
 /ngap_build_dmrpp
 /bes_default_conf.h
 /foo.dmrpp
+/airs.dmrpp
```
ndp-opendap: I think we shouldn't ignore these generated .dmrpp files but rather we might remove them for the repo

sawer297: alright, i'll make those changes this morning

PR #334, opened 2020-03-16
Inject data url

ndp-opendap:
This PR:

-  Adds code into `dmrpp_module/data/get_dmrpp.in` that places a default string value (a template string) as the value off the dmr++ `dmrpp:href` attribute if no URL is provideda at runtime.

- Added the `RemoteHttpResource::filter_retrieved_resource(std::string template_str, std::string update_str)` method that allows the caller to do a global search and replace for the template string in the cached resource. The code uses advisory file locking to make the changes. This method is then called from NgapContainer to inject the dataset URL retrieved from CMR into the dmr++ document retrieved from S3.


jgallagher59701:
Looks good!

PR #335, opened 2020-03-18
HK-599. Convert the FillValue type to the data type.

jhrg:
if this works, then we should look at HK-612.


ndp-opendap:
If pass the test it does, then you must :shipit:

PR #336, opened 2020-03-18
Fonc test cleanup

jgallagher59701:
Clean up the fileout_netcdf tests a bit. I wound up doing this because the tests were failing and it was hard to diagnose the underlying problem (an out of data hdr5 handler). I also had made a series of edits to the 'common' test macros to make for test output that's easier to use, but found that was lost (my fail). I'll rework that on a different branch.


jgallagher59701:
Merge this, then merge test-macros-fix, then double check the fileout_netcdf tests and look at moving all of the tests to use the common file. Or the common file plus a local file.

PR #338, opened 2020-03-19
Ngap edl to cmr

ndp-opendap:
Adds code to send the EDL auth token and Application ID to CMR with the CMR request


jgallagher59701:
merge

PR #340, opened 2020-03-21
Tests macros fix

jgallagher59701:
Better macros for tests - three or four of the modules now use these.


ndp-opendap:
w00t w00t the tests have passed so you must :shipit: w00t w00t

PR #341, opened 2020-03-24
cmr url injection

ndp-opendap:
This PR adds code to support identifying the CMR query endpoint URL in the bes.conf chain.


modules/ngap_module/NgapApi.cc
```diff
@@ -79,6 +79,17 @@ namespace ngap {
             "kNumberType"
     };
 
+
+    NgapApi::NgapApi() : d_cmr_endpoint_url(DEFAULT_CMR_ENDPOINT_URL) {
+        bool found;
+        string key_value;
+        TheBESKeys::TheKeys()->get_value(NGAP_CMR_ENDPOINT_URL, key_value, found);
```
jgallagher59701: No need to change this code - it works - but there are other functions in TheBESKeys that are easier to use.

modules/ngap_module/NgapCache.cc
```diff
@@ -237,7 +237,7 @@ NgapCache::get_hash(const string &s)
     }
 
 
-    string NgapCache::get_cache_file_name( const string &src,  bool /*mangle*/){
+    string NgapCache::get_cache_file_name( const string &,  bool /*mangle*/){
```
jgallagher59701: Throw an exception.

ndp-opendap: It already does that

modules/ngap_module/NgapApi.cc
ndp-opendap: I bam not aware of these. All of the examples of this that I I have found are of this form.

PR #347, opened 2020-04-01
Hyrax 104

dh-opendap:
Modified make components for bes/modules/dmrpp_module/data/get_dmrpp to generate an RPM-compatible executible script.


jgallagher59701:
This is passing its tests - lets merge it. The distclean issue where two of the dmrpp scripts were being deleted has gone away.

PR #350, opened 2020-04-08
Patch dmrpp

jgallagher59701:
This code can modify a dmrpp so that it includes information about variables 
in HDF5 files that are 'missing' WRT the CF conventions. It does this using the
CF modes of the HDF5 format and NetCDF response handlers, creating and 
referencing the missing data. The code is not currently integrated into the BES
build process, so this PR should work fine. Getting code onto the master branch 
will help speed the integration process.


jgallagher59701:
Nathan, Dan
I asked you to review this because I think we all need to get up to speed on Kent's latest work.
Thanks

PR #358, opened 2020-04-29
Hk 399

dh-opendap:
Added modules/dmrpp_module/tests/testsuite_dmrpp_tests to local check tests.   This uses build_dmrpp directly using input files located in ../data/dmrpp, notably *.h5, *.h5.dmr, and *.h5.dmrpp.baseline.


jgallagher59701:
For the `make distcheck` part of the build, the error was:
make[4]: *** No rule to make target `bes_dmrpp_tests.conf.in', needed by `distdir'.  Stop.
That means that the difference between the source directory and the build directory were confused/conflated. Check the Makefile.am and put $srcdir/ and/or $builddir/ in the targets/dependencies as needed. You can run make distcheck on your own machine and, once that's run once, cd into bes/bes-3.xx/_build/... and run `make check` to see what happens. It's a little weird since the $builddir is always in/under _build while the $srcdir is rooted at bes/bes-3.xx

PR #360, opened 2020-05-04
The xml:base context bug. A state problem with the DMR cache.

ndp-opendap:
When DMR objects are cached they contain the xml:base of the request that brought the DMR instance into being. Unfortunately, the value of xml:base for a particular dataset may change per request, as different clients may access the server through different domain names. This PR adds the insertion of the current request's xml:base context value into the DMR after a DMR copy is retrieved from cache.


jgallagher59701:
Merge!

PR #362, opened 2020-05-08
Added cmr_module to RPM package builds.

ndp-opendap:


jgallagher59701:
If this works, merge it.

PR #363, opened 2020-05-08
Distcheck 1.16 fix

jgallagher59701:
This PR introduces fixes needed to use automake 1.16 and it's 'distcheck' target. This will be needed when Travis updates automake, which could happen soon.


jgallagher59701:
NB: This passed here, but not on my machine running automake 1.16, so we should hold off on a merge.

jgallagher59701:
This push runs on my local machine.

PR #364, opened 2020-05-11
Refactoring the way that command line options to configure are processed.

ndp-opendap:
Based on: https://autotools.io/autoconf/arguments.html


ndp-opendap:
This seems to work, but I hav some concerns based on what I found here:
https://www.gnu.org/software/automake/manual/html_node/Usage-of-Conditionals.html

PR #365, opened 2020-05-12
Relaxing TypeMatch for get_dmrpp to allow any input filename.

ndp-opendap:
Because NASA and others are entrenched in naming netcdf-4 files as *.nc and because there are other know file naming convention problems I have relaxed the BES.Catalog.catalog.TypeMatch used in the bes.conf here document in the get_dmrpp so that it will accept any filename as an input (and still decompress as needed)


jgallagher59701:
Merge

PR #367, opened 2020-05-13
Added --with-cmr to configuration of bes.spec.all_static

ndp-opendap:


jgallagher59701:
Depending on how the spec file is written, you may have to add lines to the spec file to install the module and its conf file. That may be done by regex/glob or not, I cannot remember. If not you'll get an error about uninstalled files.

PR #368, opened 2020-05-13
Adjusting authentication parameters in the cmr_module.

ndp-opendap:
* Adjusted the cmr_modules curl_utils.cc so that it can utilize netrc files injected in the bes configuration
for authentication credentials.
* Changed cmr_module user agent from the cURL library
version to hyrax-cmr-virtual-directories
* Increased max redirects to 20.


jgallagher59701:
Looks good. If the tests pass, merge.

PR #370, opened 2020-05-14
Temp check dmr fnetcdf

kyang2014:
DAP4 DMR to FilenetCDF-4 with the 64-bit integer support, no groups, no structures.


kyang2014:
James asked me to merge.

PR #375, opened 2020-05-27
Hyrax 97

vskorole:


ndp-opendap:
This [last commit](https://github.com/OPENDAP/bes/pull/375/commits/efacc8fb633f002d0fa3d0707c3db22aa5325bc6) concerns me. 

- What happened here?
- Why are you choosing to replace the rapidjson that we have been using with **"folder from gihub."** 
- What does that mean?

If you are importing another version of rapidjson from a different project I would like to know more about what you are doing.

jgallagher59701:
> This [last commit](https://github.com/OPENDAP/bes/pull/375/commits/efacc8fb633f002d0fa3d0707c3db22aa5325bc6) concerns me.
> 
> * What happened here?
> * Why are you choosing to replace the rapidjson that we have been using with **"folder from gihub."**
> * What does that mean?
> 
> If you are importing another version of rapidjson from a different project I would like to know more about what you are doing.

So, it turns out that there was two sets of 'rapidjson' files and the ones that wound up in bes/rapidjson didn't compile on Linux. So I asked Slav to look into the differences, we found that there were 2-3 headers that included <limits> differently. The code that compiled and used #include <limits> and the code that compiled on OSX but not on Linux (Ubuntu) used #include <c++/v1/limits>.

To fix this, Slav went back to the original source for the library and installed code from there, following the original source's directions, into bes/rapidjson.

PR #379, opened 2020-06-05
Changes to make these tests more dependable

jgallagher59701:
Also, switched to the common handler_tests_macros.m4


jgallagher59701:
> Why is this PR not running the tests?

no idea

jgallagher59701:
I'll fix this and push when I'm home. I forgot to commit/push a directory.

PR #383, opened 2020-06-11
Hyrax 157

dh-opendap:
Two options added to get_dmrpp:

`-M` : This command line option has no argument, if selected it will initiate the pattern to determine if domain variables are missing from the generated dmrpp, and if so will synthesize those missing variables into a file named relative to the input_data_file but inserting the string '_missing' before the filename suffix (.i.e., <missing-file-name>).  A dmrpp is then generated against this missing variables file and merged into the original dmrpp.

This option requires that the '-o' option has been specified, and will report that it's required if missing. The missing data file will always be written to the same directory as data file name specified using `-o`.

`-p` <runtime_location_URL_of_the_missing_data_file> : This command line option is used to provide the runtime location of the missing data file. The server is expecting this to be either a file (file://) or http (http(s)://) URL or the server will not be able to utilize it. _This option is only utilized if option `-M` is selected, otherwise it will be ignored._  

- If option `-p` is used then the chunk element for any missing variables will contain the dmrpp:href="WhateverStringTheUserSupplied".
- If option `-p` is not used then the chunk elements for any missing variables will contain the href:"OPeNDAP_DMRpp_MISSING_DATA_ACCESS_URL" 

>**Previously**:  -p <path_to_save_missing_files> : this cmd_line option provides the fully-qualified path to the directory to save missing variable files.  At present the missing variable file will follow any relative path defined by the input_data_file.   This option operates similar to option '-b' but defines the MISSING_DATA_ROOT. 
If option '-p' is selected then the chunk element for any missing variables will contain the href:"<MISSING_DATA_ROOT>/<missing-file-name>".
If option '-p' is not selected then the chunk element for any missing variables will contain the href:" OPeNDAP_DMRpp_MISSING_DATA_ACCESS_URL" 

The files located in modules/dmrpp_module/data/patch_dmrpp/test_data/ provide test files.

The regex I use works but isn't correct, but I've been unable to determine to correct usage for this case.   Basically I need to identify all, and only, filename suffixes h5,he5,hdf5 and any camel-case variants.  This regex does, but isn't exclusive enough.


dh-opendap:
Requested change was made, the <chunk href=""> attribute will contain the exact string provided with '-p' option argument, or if the 'p' option is not used then the attribute will contain OPeNDAP_DMRpp_MISSING_DATA_ACCESS_URL.

dh-opendap:
Modified build_dmrpp to only write chunk information when chunksize is greater than 0.  This resolves the problem the TRMM data caused for the missing_variable synthesized files.

modules/dmrpp_module/build_dmrpp.cc
```diff
@@ -318,9 +318,9 @@ static void get_variable_chunk_info(hid_t dataset, DmrppCommon *dc)
             }*/
             VERBOSE(cerr << "    Addr: " << cont_addr << endl);
             VERBOSE(cerr << "    Size: " << cont_size << endl);
-
-            if (dc) dc->add_chunk("", cont_size, cont_addr, "" /*pos in array*/);
-
+            if (cont_size > 0) {
```
ndp-opendap, line 321: Well that was simple!

PR #391, opened 2020-07-08
Rpm fix

jgallagher59701:
This should address the HYRAX-195 ticket. We were setting the owner of /var/run where we used to set the owner of /var/run/bes... However, the latter is not supported by CentOS SA tools.


ndp-opendap:
When the tests pass you must :shipit:??

PR #394, opened 2020-07-21
Cache last redirect url

ndp-opendap:


jgallagher59701:
I think we should merge this.

ndp-opendap:
Summary:

This PR contains the following:

- Lots of debugging instrumentation in the http and dispatch packages.
- Added support to TheBESKeys for keys with values of:
  - vector<string>
  -  map<string,string>
- Added code that will retry http requests:
  - Triggered by a an HTTP return status of 5XX or 422
  - A new `Http.No.Retry.Regex` bes key to control what may be retried (some things are better left alone/in their original form)
  - Can be easily assessed using the new function`curl::is_retryable(string url)`

- Per Process cookie jars for each `beslistener`

- Added a prototype implementation of a cache_redirect_url() function. When enabled this will cause the ngap_module and the dmrpp_module to attempt to follow a target data URL through all of it's redirects (auth etc.) and cache the final resolved URL in oder to skip the intervening steps for the next access. This is controlled  in the dmrpp_module by the bes.conf keys:
```
Dmrpp.cache.final.redirect.urls=true
Dmrpp.no.cache.redirect.urls.regex.pattern=^https:\/\/.*s3\..*\.amazonaws\.com\/.*$
```
and from the NGAP side: 
```
NGAP.cache.terminal.urls=true
```

I think that before we merge this PR we should look at the cache_redirect URL implementation. I feel like there is unnecessary duplication in its application. This developed because we added the caching call to `NgapContainer::access()`:
```
if(cache_terminal_urls()){
    curl::cache_final_redirect_url(data_access_url_str, 0);
}
```
But when we set up _ngap.opendap.org_ we needed to have it work for local dmr++ files with CloudFront/TEA URLs inside of them, so I added similar calls to the `DmrppParserSaz2::dmr_start_element()`:
In _paser_start_ case, line 827:
```
if(parser->use_last_accessed_urls()){
    BESDEBUG(PARSER, prolog << "Attempting to locate and cache last redirect URL for Chunk URL: " << parser->dmrpp_dataset_href << endl);
    curl::cache_final_redirect_url(data_url,parser->get_no_cache_redirect_urls_regex());
}
```
and in case _inside_dmrpp_object_ on line 1037:
```
if(parser->use_last_accessed_urls()){
    BESDEBUG(PARSER, prolog << "Attempting to locate and cache last redirect URL for Chunk URL: " << parser->dmrpp_dataset_href << endl);
    curl::cache_final_redirect_url(data_url,parser->get_no_cache_redirect_urls_regex());
}
```

At this point I believe that we need to discuss some cleanup from the code/test frenzy, primarily consolidating the cache redirect url controls controls to the http package, and changing the semantics of the BES keys to match the use pattern and to reflect the vocabulary of the cURL lib. i.e. 
  1. Replace the BES keys `NGAP.cache.terminal.urls` and `Dmrpp.cache.final.redirect.urls` and with `Http.cache.effective.urls`
  1. Consolidate the implementations of `DmrppParserSax2::use_last_accessed_urls()` and `NgapContainer::cache_terminal_urls()` into `CurlUtils::cache_effective_urls()` 
  1. Replace the BES key `Dmrpp.no.cache.redirect.urls.regex.pattern` with `Http.cache.effective.urls.skip.regex.pattern` and in the process consider making it a cache if match pattern rather than a skip if match.
  1. Consolidate the relevant parts of`DmrppParserSax2::load_no_cache_redirect_urls_regex()` along with `bool NgapContainer:: cache_terminal_urls()` into `bool CurlUtils::cache_terminal_urls()`
  1. Consider changing the implementation of:
```
curl::cache_final_redirect_url(string url,BESRegex no_cache_regex);
```
So that it deals with the `no_cache_regex` behind the scenes:
```
curl::cache_final_redirect_url(string url);
```
Since the value of the regex is retrieved from `TheBESKeys` this is easy.


Since that actual caching implementation 
CurlUtils::cache_final_redirect_url()

ndp-opendap:
Changes made. Tests pass...

PR #397, opened 2020-08-19
Dynamic config

ndp-opendap:
This PR adds a dynamic configuration feature that utilize one or more regular expressions to associate specific configuration settings with matching  path/dataset names.


xmlcommand/BESXMLInterface.cc
```diff
@@ -145,6 +149,15 @@ void BESXMLInterface::build_data_request_plan()
                 // BESXMLCommand object
                 string node_name = (char *) current_node->name;
 
+                if(node_name == SETCONTAINER_STR){
+                    string name;
+                    string value;
+                    map<string,string> props;
+                    BESXMLUtils::GetNodeInfo(current_node, name, value, props);
+                    BESDEBUG(MODULE, prolog << "In "  << SETCONTAINER_STR << " element. Value: " << value << endl);
+                    TheBESKeys::TheKeys()->load_dynamic_config(value);
```
ndp-opendap, line 158: Once we do this, what happens of the is a non-fatal `BESError`? How will we undo? Do we want to call unload before this, or even as the 1st order pf business in `TheBESKeys::TheKeys()->load_dynamic_config()` - Like clear before load always? Because:
- `TheBESKeys::TheKeys()->unload_dynamic_config()` is a no-op if there is not a dynamic config in play.
- If no dynamic configuration exists OR there is no match the configuration remains as loaded from disk.
- If a dynamic configuration is loaded, then the configuration get's restored to the as loaded from disk state before the new configuration is applied.

I think I am trying to say that this should be an atomic operation and not require the `TheBESKeys::TheKeys()->unload_dynamic_config()` be called elsewhere as that might cause a lot of grief for us down the road.

jgallagher59701: What to do in case of an error is a good point. The code could always call the unload method first, to clean up from a previous error. Can we look at the code in BESInterface (I mean that loosely) to put a call to the unload op in the catch blocks there? Or - maybe the best option - is wrap the load/unload operation in an object with a lifetime limited to the scope of the XMLInterface::execute_data_request_plan() method.

ndp-opendap: But if we call unload() inside load() so that:
```
load(){
    unload();
    load_stuff_or_not_as_needed();
}
```
Then we are always in the correct state after load(), and the state going into load() is always handled correctly.

jgallagher59701:
I think we should merge this, test it, and decide to fix/change/optimize based on those tests.

PR #398, opened 2020-08-19
Stare 3

jgallagher59701:
This adds a fourth server function for STARE that will subset an array and return the masked result. That is, the original array is all there, but the pixels not in the ROI as specified by the STARE indices will be masked out.


jgallagher59701:
Merge this?

PR #404, opened 2020-09-10
Hyrax 258

dh-opendap:
dmrpp_module/Base64.h added to provide static functions for base64 encode() and decode().  To be used by DmrppParserSax2 and DmrppArray, and potentially other Dmrpp<type> instances.
dmrpp_module/build_dmrpp.cc modified get_variable_chunk_info() to handle COMPACT layout variables in H5 files.


jgallagher59701:
Merge this.

PR #405, opened 2020-09-13
Curl init v1

ndp-opendap:
Moving towards consolidated cURL init with EDL tokens in place. Next, tests...


jgallagher59701:
Merge!

ndp-opendap:
There is more to do that we need to discuss w.r.t. dmrpp code.

PR #408, opened 2020-09-18
Dmrpp refactor

jgallagher59701:


modules/dmrpp_module/CurlHandlePool.cc
```diff
@@ -62,8 +62,10 @@
 #define CURL_VERBOSE 0  // Logs curl info to the bes.log
 
 #define prolog std::string("CurlHandlePool::").append(__func__).append("() - ")
+#define NEW_WAY 1
 
 static const int MAX_WAIT_MSECS = 30 * 1000; // Wait max. 30 seconds
+// FIXME retry_limit is not used. jhrg 9/18/20
```
ndp-opendap, line 68: retry_limit was moved to line 73 of CurlUtils.cc it can be removed

jgallagher59701: OK. I want to get the PR building before I do that. But good to know.

PR #416, opened 2020-09-23
Dmrpp refactor 2

jgallagher59701:
This refactor removes the last bits of the curl multi handle API. The parallel access is now always handled by pthreads. The code also contains the start of a RAII class for the dmrpp_easy_handle code. It's not used, but the class will be part of the build - just to see if the compilers barf.

### DIFF
--- a/modules/dmrpp_module/CurlHandlePool.cc
+++ b/modules/dmrpp_module/CurlHandlePool.cc
@@ -242,10 +242,17 @@ dmrpp_easy_handle::~dmrpp_easy_handle() {
 #if NEW_WAY
 
 /**
- * @brief This is the read_data() method for serial transfers.
+ * @brief This is the read_data() method for all transfers.
  *
- * See below for the code used for parallel transfers. Note that
- * this code is also used for the pthreads parallel transfers.
+ * This method is used by Chunk::read_data() which is used for all
+ * data transfers by the dmrpp module classes (DmrppArray, D4Opaque,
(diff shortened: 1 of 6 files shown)


modules/dmrpp_module/CurlHandlePool.cc
```diff
@@ -242,10 +242,17 @@ dmrpp_easy_handle::~dmrpp_easy_handle() {
 #if NEW_WAY
```
ndp-opendap, line 242: If we're going to commit to doing this the NEW_WAY then lets drop the #if conditional...

modules/dmrpp_module/CurlHandlePool.h
```diff
@@ -121,6 +121,35 @@ class CurlHandlePool {
...
 
+/**
+ * How a collection of dmrpp_easy_handles that are being used together on
+ * a single logical transfer. By definition, if one of these fails, they all
+ * fail, are stopped and the easy handles reset and returned to the pool.
+ * This class is used to portect leaking handles when one thread of a
+ * parallel transfer fails and an exception is thrown taking the flow of
+ * control out of the handler to the command processor loop.
+ */
+class SwinLane {
```
ndp-opendap, line 132: I think you meant "SwimLane" and not "SwinLane"

modules/dmrpp_module/CurlHandlePool.cc
jgallagher59701: Done

PR #417, opened 2020-09-25
Dmrpp swim lanes

jgallagher59701:
I have the curl handle pool able to recover from threads throwing exceptions. I did not use the SwimLane idea in the end, but found that catching the exception in the thread was key, because the exceptions do not propagate out f child threads to the main thread. I use the thread exist status, test for an error and re-throw after cleaning up the handle. The re-thrown exception, now in the main thread is caught and all the outstanding threads joined (and thus all outstanding handles returned to the pool) before the exception is thrown yet again, this time to the BES outer level.

I've added a unit test for this using mock classes for DmrppArray and Chunk and stand in code for the DmrppArray thread control code. That is in CurlHandlerPoolTest. It is as close as a unit test as I could get.

Some of this can be consolidated more because there is duplication between the three 'read' methods in DmrppArray, but we should wait because better option might be available now that we can move onto real c++11 (which includes thread support).


ndp-opendap:
:shipit:

ndp-opendap:
:shipit:

PR #419, opened 2020-10-01
Logging mods for better downstream log management.

ndp-opendap:
Migrating to ERROR macro from LOG macro where appropriate


ndp-opendap:
I think we can merge this. The old logging mode can be enabled using` --without-newlog` configure switch:
```
./configure --without-newlog 
```
And the tests pass.

Merge @jgallagher59701 ???

jgallagher59701:
merge

PR #422, opened 2020-10-20
Pick up the changes in the hdf4 handler for the 2.20 version of HDFEOS

jgallagher59701:
The signatures of some of the functions changed and I hacked the
hdf4 handler to accommodate those changes.


captainkirk99:
Is this going to be merged? Does it handle both 2.19 and 2.20? Or does it switch to 2.20 and not work with 2.19?

Also, I thought from the discussion in #447 that we are only going to use 2.19? Hence PR #449

PR #423, opened 2020-10-20
HYRAX-278, remove groups and its attributes if there are no variables…

kyang2014:
… specified in the expression constraint under its group orits subgroups.


modules/fileout_netcdf/FONcTransform.cc
```diff
@@ -467,14 +474,35 @@ void FONcTransform::transform_dap4_no_group() {
 
 void FONcTransform::transform_dap4_group(D4Group* grp,bool is_root_grp,int par_grp_id,map<string,int>&fdimname_to_id ) {
 
+    bool included_grp = false;
+    // Always include the root attributes.
+    if(is_root_grp == true)  
+        included_grp = true;
+    else {
+        // Check if this group is in the group list kept in the file.
+        set<string>::iterator iset;
```
jgallagher59701, line 483: I think you can remove iset here.

modules/fileout_netcdf/FONcTransform.cc
```diff
@@ -591,6 +623,54 @@ bool FONcTransform::check_group_support() {
...
+void FONcTransform::gen_included_grp_list(D4Group*grp) 
+{
+    bool grp_has_var = false;
+    if(grp) {
+        BESDEBUG("fnoc", "<coming to the D4 group  has name " << grp->name()<<endl);
+        BESDEBUG("fnoc", "<coming to the D4 group  has fullpath " << grp->FQN()<<endl);
+        //We always include root attributes, so no need to obtain grp_names for the root.
+        if(grp->var_begin()!=grp->var_end()) {
+            BESDEBUG("fnoc", "<has the vars  " << endl);
+            Constructor::Vars_iter vi = grp->var_begin();
```
jgallagher59701, line 636: No need to change this, but FYI, you could say "for (auto vi = ... )"

modules/fileout_netcdf/FONcTransform.h
```diff
@@ -65,6 +66,7 @@ class FONcTransform: public BESObj {
 	string _returnAs;
 	vector<FONcBaseType *> _fonc_vars;
 	vector<FONcBaseType *> _total_fonc_vars_in_grp;
+    set<string> _included_grp_names;
```
jgallagher59701, line 69: The STL 'set' class is sorted by name. If you want to preserve the dataset's ordering of the groups, there's an unordered_set you can use although it might break the CentOS 6 build. I'm going to remove that this Sprint or next sprint, so using unordered_set<> is an option.

jgallagher59701 (approved):
I think we should merge this now. Then address the issues. Also, I think some more tests would be good unless the one you added covers all the edge cases. That can be a separate branch & PR.

PR #426, opened 2020-10-29
More thread safety

ndp-opendap:
Fixes initialization issues with EffectiveUrlCache


dispatch/BESDebug.cc
```diff
@@ -86,7 +115,7 @@ void BESDebug::SetUp(const string &values)
     }
     else {
         strm = new ofstream(s_strm.c_str(), ios::out);
-        if (strm && !(*strm)) {
+        if (strm && strm->fail()) {
```
jgallagher59701, line 118: You don't need to test strm here - new will throw if it cannot allocate

jgallagher59701:
Merge

PR #428, opened 2020-11-04
Timing triggers

ndp-opendap:
Adding BESStopwatch timers to the various parts of the dmr++ serialization processes.

### DIFF
--- a/bes.spec.all_static
+++ b/bes.spec.all_static
@@ -201,6 +201,7 @@ exit 0
 %{_bindir}/get_dmrpp
 %{_bindir}/ingest_filesystem
 %{_bindir}/ingest_s3bucket
+%{_bindir}/retriever
 # There is now a separate project for a new build_sidecar program. This should not
 # be installed. jhrg 8/20/20
 # %{_bindir}/build_sidecar
--- a/bes.spec.all_static
+++ b/bes.spec.all_static
@@ -201,7 +201,7 @@ exit 0
 %{_bindir}/get_dmrpp
 %{_bindir}/ingest_filesystem
 %{_bindir}/ingest_s3bucket
-%{_bindir}/retriever
+#%{_bindir}/retriever
 # There is now a separate project for a new build_sidecar program. This should not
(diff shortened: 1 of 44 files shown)


modules/dmrpp_module/Makefile.am
```diff
@@ -70,7 +70,7 @@ libdmrpp_module_la_LDFLAGS = -avoid-version -module
 libdmrpp_module_la_LIBADD = $(BES_DISPATCH_LIB) $(BES_HTTP_LIB) $(DAP_SERVER_LIBS) $(DAP_CLIENT_LIBS) \
 $(H5_LDFLAGS) $(H5_LIBS) $(OPENSSL_LDFLAGS) $(OPENSSL_LIBS) -ltest-types
 
-bin_PROGRAMS = build_dmrpp check_dmrpp merge_dmrpp reduce_mdf
+bin_PROGRAMS = build_dmrpp check_dmrpp merge_dmrpp reduce_mdf retriever
```
jgallagher59701, line 73: This is where you could make 'retriever' a noinst_program would solve the problem of having the program in the RPM, go through a security audit, etc.

https://gnu.huihoo.org/automake-1.6.1/html_chapter/automake_2.html

The special prefix `noinst' indicates that the objects in question should be built but not installed at all. This is usually used for objects required to build the rest of your package, for instance static libraries (see section 9.2 Building a library), or helper scripts.

bes.spec.all_static
```diff
@@ -201,6 +201,7 @@ exit 0
 %{_bindir}/get_dmrpp
 %{_bindir}/ingest_filesystem
 %{_bindir}/ingest_s3bucket
+%{_bindir}/retriever
```
jgallagher59701, line 204: If you make the change I suggest further along, I think this can be removed (and the similar line in bes.spec)

jgallagher59701 (commented):
Try these changes.

bes.spec.all_static
ndp-opendap: Done

PR #430, opened 2020-11-09
11/9/20 - SonarFix - (Testing) vulnerability fixed

sawer297:
testing if Travis and Sonar passes

### DIFF
--- a/modules/dmrpp_module/reduce_mdf.cc
+++ b/modules/dmrpp_module/reduce_mdf.cc
@@ -101,6 +101,10 @@ short update_sha256_file(char* m_dmrpp_fname,char* m_h5_fname,char* m_sha256_fna
     // If the recording file that stores thesha256 doesn't exist, 
     // just create this file and write the sha256 etc information to the file.
 
+    /* removed 11.10.20 SBL
+     * removed due to sonarcloud claiming that having the following before the if statement on ln 126
+     * causes a race condition and a security vulnerability
+    */
     //if(access(m_sha256_fname,F_OK)==-1)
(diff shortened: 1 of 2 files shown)


modules/dmrpp_module/reduce_mdf.cc
```diff
@@ -100,8 +100,9 @@ short update_sha256_file(char* m_dmrpp_fname,char* m_h5_fname,char* m_sha256_fna
 
     // If the recording file that stores thesha256 doesn't exist, 
     // just create this file and write the sha256 etc information to the file.
-    if(access(m_sha256_fname,F_OK)==-1) 
-        return write_sha256_file(m_dmrpp_fname,m_h5_fname,m_sha256_fname,sha256_buf);
+
+    //if(access(m_sha256_fname,F_OK)==-1)
```
jgallagher59701, line 104: Lets clean up these comments. Explain why the above was removed (or just eliminate it), fix the typos in the 'If the recording...' comments and then explain briefly write_sha256_file() should be called if open() fails (because two years from now we'll look at this and go, "Whoa, what's going on here?").

Thanks.

jgallagher59701 (approved):
Fix the comment(s) and merge

modules/dmrpp_module/reduce_mdf.cc
sawer297: added comments

jgallagher59701 (approved):

jgallagher59701:
Merge when this PR passes. Thanks.

PR #437, opened 2020-12-05
fixed subdir-objects warnings in autoreconf

### LINKED ISSUE #435: autoreconf warnings about subdir-objects setting
Getting some warnings about the subdir-objects setting.

This setting causes object files to be always placed in the same build dir as their source, instead of the build dir of the Makefile.am that is using them. (This is the same directory unless the Makefile.am includes source files from other directories, as happens in several places in bes.) This is documented here: https://www.gnu.org/software/automake/manual/html_node/Program-and-Library-Variables.html

IIRC this is a result of changes in automake defaults. In olden times, subdir-objects was on by default, but now it is off by default.

This is fixed by adding subdir-objects to the list of options for AM_INIT_AUTOMAKE in configure.ac.

```
 functions/unit-tests/Makefile.am:140: warning: source file '../RangeFunction.cc' is in a subdirectory,
functions/unit-tests/Makefile.am:140: but option 'subdir-objects' is disabled
automake: warning: possible forward-incompatibility.
automake: At least a source file is in a subdirectory, but the 'subdir-objects'
automake: automake option hasn't been enabled.  For now, the corresponding output
automake: object file(s) will be placed in the top-level directory.  However,
automake: this behaviour will change in future Automake versions: they will
automake: unconditionally cause object files to be placed in the same subdirectory
automake: of the corresponding sources.
automake: You are advised to start using 'subdir-objects' option throughout your
automake: project, to avoid future incompatibilities.
functions/unit-tests/Makefile.am:132: warning: source file '../scale_util.cc' is in a subdirectory,
functions/unit-tests/Makefile.am:132: but option 'subdir-objects' is disabled
functions/unit-tests/Makefile.am:136: warning: source file '../scale_util.cc' is in a subdirectory,
functions/unit-tests/Makefile.am:136: but option 'subdir-objects' is disabled
modules/dmrpp_module/Makefile.am:78: warning: source file '$(srcdir)/../hdf5_handler/h5common.cc' is in a subdirectory,
modules/dmrpp_module/Makefile.am:78: but option 'subdir-objects' is disabled
modules/dmrpp_module/unit-tests/Makefile.am:127: warning: source file '$(top_srcdir)/modules/read_test_baseline.cc' is in a subdirectory,
modules/dmrpp_module/unit-tests/Makefile.am:127: but option 'subdir-objects' is disabled
modules/dmrpp_module/unit-tests/Makefile.am:130: warning: source file '$(top_srcdir)/modules/read_test_baseline.cc' is in a subdirectory,
modules/dmrpp_module/unit-tests/Makefile.am:130: but option 'subdir-objects' is disabled
```

captainkirk99:
Fixes #435 

This fixes autoreconf warnings.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -24,7 +24,7 @@ DODS_GCOV_VALGRIND
 AC_CONFIG_AUX_DIR([conf])
 AC_CONFIG_MACRO_DIR([conf])
 
-AM_INIT_AUTOMAKE([1.10 tar-pax])
+AM_INIT_AUTOMAKE([1.10 tar-pax subdir-objects])
 LT_INIT
 
 dnl flags for the compilers and linkers - set these before locating the


jgallagher59701:
Closed for because of encrypted env vars and upstream PRs. See https://docs.travis-ci.com/user/pull-requests/#pull-requests-and-security-restrictions

PR #440, opened 2020-12-05
trying github actions build

captainkirk99:
trying github actions build

### DIFF
--- a/.github/workflows/autotools_hdfeos2.20.yml
+++ b/.github/workflows/autotools_hdfeos2.20.yml
@@ -1,0 +1,70 @@
+name: Autotools ubuntu-latest HDFEOS2.20
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+
+    runs-on: ${{ matrix.os }}
+    strategy:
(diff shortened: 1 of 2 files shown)


captainkirk99:
@jgallagher59701 this is a modified version of the STAREmaster CI file. It will build a few of the dependencies and then try an autoreconf -i. It does nothing further, because I want to get that passing so it caches the dependencies. But it's not even running.

I expected it to run when I made it part of this PR, but it does not. Perhaps I need to be added to the list of developers on this repo? Or perhaps you need to merge this to get the ball rolling?

On all my other repos, it started working as soon as I submitted the PR with the workflow. So propably this is because I am not a developer on this github repo...

captainkirk99:
OK, now that I think about it, it makes sense that this would not work for people not on the team...

jgallagher59701:
This is running into our issues with the deps that come from our other NASA work. I'm working on a fix.

captainkirk99:
I think if you add me to the list of developers on the project I can get the GitHub actions working without worrying about the travis build.

But right now it does not even run my attempt on github actions, so we can't see how it is failing...

captainkirk99:
Trying this again as #454

PR #449, opened 2020-12-08
check version of hdfeos2 library in configure to ensure it is not 2.20

### LINKED ISSUE #447: Difficulties adjusting to different versions of HDFEOS2 library for HDF4 files
The HDFEOS2 library is for NASA HDF4 files, like MODIS.

As discussed in this issue (https://github.com/SpatioTemporal/STAREmaster/issues/111), there are two different versions of HDFEOS2 that should be accomodated, 2.19 and 2.20. Unfortunately there is some const changes which interact badly with the hdf4_handler file HDFEOS2.cc.

In HDFEOS2.cc there is code that passes some of the HDFEOS2 functions as function pointers to a function that calls the HDFEOS2 function. For example:

```
    vector<string> gridlist;
    if (!Utility::ReadNamelist(file->path.c_str(), GDinqgrid, gridlist)) {
        delete file;
        throw1("Grid ReadNamelist failed.");
    }

```

This is calling the Utility::ReadNamelist() function, which is also in HDFEOS2.cc:

```
// Read name list from the EOS2 APIs and store names into a vector
bool Utility::ReadNamelist(const char *path,
                           int32 (*inq)(const char *, char *, int32 *),
                           vector<string> &names)
{
    char *fname = const_cast<char *>(path);
    int32 bufsize;
    int numobjs;

    if ((numobjs = inq(fname, NULL, &bufsize)) == -1) return false;
    if (numobjs > 0) {
        vector<char> buffer;
        buffer.resize(bufsize + 1);
        if (inq(fname, &buffer[0], &bufsize) == -1) return false;
        HDFCFUtil::Split(&buffer[0], bufsize, ',', names);
    }
    return true;
}

```

Note that the second (unnamed) parameter to ReadNamelis() is a function pointer.

This ReadNamelist function is called only 3 times, for functions GDinqgrid(), SWinqswath(), and PTinqpoint().

The challenge here is that one of these functions, SWinqswath(), has a different first parameter in 2.19 and 2.20. In 2.20 it adds a const.

captainkirk99:
Fixes #447 

As noted in issue #447 only HDFEOS2-2.19 can be used. 2.20 should not be used and will cause compile errors.

In this PR I add a check in configure.ac to ensure that 2.20 is not being used.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -738,7 +738,6 @@ if test  -n "$HDFEOS2_DIR"; then
       choke me
 #endif
   ]])], [AC_MSG_ERROR([HDFEOS2 library must be version 2.19, not version 2.20.])], [])
-  AC_MSG_RESULT([$hdfeos2_20])
   CPPFLAGS=$OLD_CPPFLAGS  
 fi
 
--- a/configure.ac
+++ b/configure.ac
@@ -53,7 +53,7 @@ AC_ARG_ENABLE([developer],
 )
 AM_CONDITIONAL([BES_DEVELOPER], [test "x$enable_developer" = "xyes"])
 AM_COND_IF([BES_DEVELOPER],
-    [cxx_debug_flags="-g3 -O0  -Wall -W -Wcast-align $CXXFLAGS"
+    [cxx_debug_flags="-g3 -O0  -Wall -W -Wcast-align"
       AC_MSG_NOTICE([Developer Mode is enabled.]) ],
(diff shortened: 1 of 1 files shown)


jgallagher59701:
Ed, To run our Travis build, you need various encrypted env vars. But, Travis removes them from upstream PRs for security reasons. To test this, you'll need to wait for the github actions to start working for the BES or run these PRs from the OPeNDAP account. I think you can do that later as a collaborator. 

I'm going to close this PR and the other two that fail for the same reason.

PR #452, opened 2020-12-09
Fnetcdf group new

kyang2014:
This check-in includes significant enhancement of fileout netCDF-4 for the DAP4. It includes:
1) Make the expression constraint follow the DAP4 specification. 
2) Handle the netCDF-4 pure dimension case.
3) Handle scalar/string variables.
4) Handle the possible dimension name clashing case.
5) Test with sample NASA files and add the testsuite for NASA files.


jgallagher59701:
Reopened to trigger a build on Travis

modules/fileout_netcdf/FONcArray.cc
```diff
@@ -40,6 +41,7 @@
 #include "FONcMap.h"
 #include "FONcUtils.h"
 #include "FONcAttributes.h"
+#include <algorithm>
```
jgallagher59701, line 44: Not a huge deal, but I like to keep the different types of headers grouped together. C++ headers, then C headers, then libdap, bes and finally handler.

jgallagher59701 (approved):

jgallagher59701:
I'll merge this now.

PR #455, opened 2020-12-10
Chunk ptrs

ndp-opendap:
Migrated the collection of Chunk objects to shared_ptr's


ndp-opendap:
James - I think this is ready...

PR #456, opened 2020-12-16
NaN issue #420

sawer297:
fix for the NaN turning into 0.0


ndp-opendap (changes requested):
This looks good but I am thinking it needs a test. Probably a unit test where we use a little netcdf file with an "NaNf" valued attribute to exercise the bug/patch.

sawer297:
@ndp-opendap tests added

ndp-opendap (approved):
If the tests pass then you must :shipit:, :shipit:, you must :shipit:

PR #467, opened 2021-01-26
Removed the old sidecar builder which used an old libdap::Error ctor.

jgallagher59701:
This will fix issues with the build when the STARE library is included in the set of dependencies.


ndp-opendap (commented):
:shipit: when the tests pass

ndp-opendap (approved):

PR #468, opened 2021-01-26
Resty path update

ndp-opendap:
This is the new form of the NGAP restified path documented here: 

https://wiki.earthdata.nasa.gov/display/DUTRAIN/Feature+analysis%3A+Restified+URL+for+OPENDAP+Data+Access

This PR implements the new version and the old version of the restified path, and will route the request to them based on the presence/absence of the (old version) `/providers/` key string in the path.

I added unit tests for the new version, one for the path with the optional parts and one for the path without. The old tests for the old way are still working too.


dh-opendap:
Looks good, wasn't expecting the 'entry_title' for the previous incarnation internal test string to include embedded '/' but I assume that was(and is) handled correctly.  I think this is ready to merge.

ndp-opendap:
> Looks good, wasn't expecting the 'entry_title' for the previous incarnation internal test string to include embedded '/' but I assume that was(and is) handled correctly. I think this is ready to merge.
In practice all of the key strings are surrounded by "/" characters  in the path. By adding the "/" to the key definitions I can avoid problems caused by key names or "/" characters showing up in the values.

jgallagher59701:
Merge!

PR #469, opened 2021-01-26
First pass at loading modules and stuff after UID switch.

ndp-opendap:
When I added code to the `DmrppRequestHandler` that wrote to the `bes.log` during initialization we discovered that this broke the server because the log lines were being written into the log file BEFORE the server switched users from `root` to `bes`. This caused the server to be unable to start because the resulting `bes.log` file was owned by `root` and the `bes_listener` processes are all running as `bes`, unable to write to the `root` owned log file.

In this PR I have reorganized the code in `daemon.cc` to load the modules and signal handlers after the UID switch.


jgallagher59701:
Merge

PR #474, opened 2021-02-09
Refactor the singleton classes used by the dmrpp_handler to use c++11 concurrency machinery.

ndp-opendap:
This branch replaces the use of the `pthreads` library with more modern c++11 machinery such as:

- std::mutex
- std::lock_guard
- std::once_flag
- std::call_once()
- std::async()
- std::future

The result is simpler code based on more current libraries.


jgallagher59701:
Looks good to me.

I noticed the use of {}'s in the private fields to set the default value for the mutex. Also, use of override instead of virtual. Good.

ndp-opendap:
> I noticed the use of {}'s in the private fields to set the default value for the mutex. 
I think I managed to do that by accident, and I think it's a mistake. The `std::mutex` class has a no parameter constructor and is an actual class with methods, so I think that using {} after the declaration is just confusing and meaningless. (None of the examples I found of using `std::mutex` showed that pattern)

jgallagher59701 (approved):
I think it looks pretty good. Are there parts of this you want me to look at in particular?

PR #475, opened 2021-02-19
Update cache files

sawer297:
updating cache files


http/RemoteResource.cc
```diff
@@ -421,7 +344,119 @@ namespace http {
...
+            BESDEBUG(MODULE, prolog << "Replaced " << count <<
+                                    " instance(s) of template(" <<
+                                    template_key << ") with " << replace_value << " in cached RemoteResource" << endl);
+        }
+
+        //  -  -  -  -  -  -  -  -  -  -  -  -  -  -  -  -  -  -  -  -  -  -  -  -  -  -  -  -  -  -  -  -  -  -
+        // I think right here is where I would be able to cache the data type/response headers. While I have
+        // the exclusive lock I could open another cache file for metadata and write to it.
+        {
+            // FIXME THIS IS WHERE WE NEED TO USE NGAP HASH CACHE FILE NAMES NOT FLC NAMES.
```
jgallagher59701, line 388: Does this still need to be done? Also, rephrase the 'I think ...' comment unless there's still some question.

jgallagher59701 (commented):
Address the two issues with the comments - if the FIXME still needs to be fixed, then I might need to know more about the issue there.

Also, for unit tests, you will need four tests at least (one C++ program with four tests). Write tests for 

* RemoteResource::ifilter_retrieved_resource()
* RemoteResource::is_cached_resource_expired()
* RemoteResource::load_hdrs_from_file()
* RemoteResource::update_file_and_headers()

http/RemoteResource.cc
ndp-opendap: These comments are cruft and could be removed.

http/RemoteResource.cc
```diff
@@ -421,7 +344,119 @@ namespace http {
...
+        // own lock on the file.
+        unsigned long long size = cache->update_cache_info(d_resourceCacheFileName);
+        BESDEBUG(MODULE, prolog << "Updated cache info" << endl);
+
+        if (cache->cache_too_big(size)) {
+            cache->update_and_purge(d_resourceCacheFileName);
+            BESDEBUG(MODULE, prolog << "Updated and purged cache." << endl);
+        }
+        BESDEBUG(MODULE, prolog << "END" << endl);
+        d_initialized = true;
```
ndp-opendap, line 423: Just reading the headers is not sufficient to label the instance as initialized.
Setting the state variable `d_initialized` should happen in the calling methods that orchestrate the entire initialization/inflation of the object.

http/RemoteResource.cc
sawer297: removed

http/RemoteResource.cc
sawer297: so move the "d_initialized = true" to after the method call

sawer297: should the same be done with the load_hdrs_from_files()?

http/RemoteResource.cc
```diff
@@ -72,6 +72,8 @@ namespace http {
         d_response_headers = new vector<string>();
         d_http_response_headers = new map<string, string>();
 
+        d_expires_interval = expiredInterval;
```
ndp-opendap, line 75: Like this:
```
RemoteResource::RemoteResource(const std::string &url,const std::string &uid, long long expiredInterval= DEFAULT_EXPIRED_INTERVAL): d_expires_interval(expiredInterval), d_uid(uid),
```

http/RemoteResource.h
```diff
@@ -26,6 +26,7 @@
 
 #ifndef  _bes_http_REMOTE_HTTP_RESOURCE_H_
 #define  _bes_http_REMOTE_HTTP_RESOURCE_H_ 1
+#define DEFAULT_EXPIRED_INTERVAL 3600
```
ndp-opendap, line 29: This line can move to `RemoteResource.cc` which prevents it from leaking into the consumers of this class.

ndp-opendap (changes requested):

http/RemoteResource.cc
```diff
@@ -60,9 +61,8 @@ using namespace std;
 
 namespace http {
 
+    RemoteResource::RemoteResource(const std::string &url,const std::string &uid, long long expiredInterval){
```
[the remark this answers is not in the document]
jgallagher59701: Lets leave it in the header

http/unit-tests/RemoteResourceTest.cc
```diff
@@ -215,14 +218,135 @@ class RemoteResourceTest: public CppUnit::TestFixture {
     // Called after each test
     void tearDown()
     {
+        if(!d_temp_file.empty())
+            unlink(d_temp_file.c_str());
+
+        string temp_file_hdrs = d_temp_file + ".hdrs";
+        if(!temp_file_hdrs.empty())
```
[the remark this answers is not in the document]
jgallagher59701: I think you should remove the tests - the d_temp_file field will never be empty.

http/unit-tests/RemoteResourceTest.cc
```diff
@@ -215,14 +218,135 @@ class RemoteResourceTest: public CppUnit::TestFixture {
...
+            string expected_header_1 = "This ...";
+            string expected_header_2 = "Is ...";
+            string expected_header_3 = "A ...";
+            string expected_header_4 = "TEST";
+
+            if(debug) cerr << prolog << "Expected Header: " << expected_header_1 << "   <--/-->     Retrieved Header: " << header_1 << endl;
+            if(debug) cerr << prolog << "Expected Header: " << expected_header_2 << "   <--/-->     Retrieved Header: " << header_2 << endl;
+            if(debug) cerr << prolog << "Expected Header: " << expected_header_3 << "   <--/-->     Retrieved Header: " << header_3 << endl;
+            if(debug) cerr << prolog << "Expected Header: " << expected_header_4 << "   <--/-->     Retrieved Header: " << header_4 << endl;
+            CPPUNIT_ASSERT( header_1.compare(expected_header_1) && header_2.compare(expected_header_2) && header_3.compare(expected_header_3) && header_4.compare(expected_header_4));
```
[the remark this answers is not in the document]
jgallagher59701: Agree - use separate assert stmts

jgallagher59701 (changes requested):

http/unit-tests/RemoteResourceTest.cc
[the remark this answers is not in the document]
sawer297: == does not work
i tried using it first, switched to using the compare() when it decided to see blank lines instead of strings

http/unit-tests/RemoteResourceTest.cc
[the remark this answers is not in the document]
sawer297: done

http/unit-tests/RemoteResourceTest.cc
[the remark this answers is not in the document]
sawer297: resolved, thxs nathan

jgallagher59701:
Why was the PR build cancelled? 

If/when it's ready to merge, please use the 'Squash and merge' option to merge the PR.

PR #482, opened 2021-03-26
Update cache files v3

sawer297:
added Http.Cache.expires.time parameter to bes config files and code that uses same in RemoteResources

### DIFF
--- a/http/HttpCache.cc
+++ b/http/HttpCache.cc
@@ -85,6 +85,25 @@ namespace http {
         return size_in_megabytes;
     }
 
+    unsigned long long HttpCache::getCacheExpiresTimeFromConfig() {
+        bool found = false;
+        string time;
+        unsigned long time_in_seconds = 0;
+        TheBESKeys::TheKeys()->get_value(HTTP_CACHE_EXPIRES_TIME_KEY, time, found);
+
+        if (found) {
+            std::istringstream iss(time);
+            iss >> time_in_seconds;
+        } else {
+            stringstream msg;
+            msg << prolog << "The BES Key " << HTTP_CACHE_EXPIRES_TIME_KEY << " is not set.";
+            BESDEBUG(HTTP_MODULE, msg.str() << endl);
+            throw BESInternalError(msg.str(), __FILE__, __LINE__);
+        }
+
+        return time_in_seconds;
+    }
+
     string HttpCache::getCacheDirFromConfig() {
         bool found;
         string subdir = "";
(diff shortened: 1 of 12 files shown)


http/HttpCache.cc
```diff
@@ -85,6 +85,22 @@ namespace http {
         return size_in_megabytes;
     }
 
+    unsigned long long HttpCache::getCacheExpiresTime() {
```
jgallagher59701, line 88: Change this to return 'unsigned long' to match the type of 'time_in_seconds'.

http/HttpCache.h
```diff
@@ -97,6 +97,7 @@ namespace http {
         virtual std::string get_cache_file_name(const std::string &src, bool mangle=true);
 #endif
 
+        static unsigned long long getCacheExpiresTime();
```
jgallagher59701, line 100: unsigned long long --> unsigned long here too. Also, I'm not sure I understand why this method is 'static'.

http/HttpCache.cc
```diff
@@ -85,6 +85,22 @@ namespace http {
...
+        bool found = false;
+        string time;
+        unsigned long time_in_seconds = 0;
+        TheBESKeys::TheKeys()->get_value(HTTP_CACHE_EXPIRES_TIME_KEY, time, found);
+
+        if (found) {
+            std::istringstream iss(time);
+            iss >> time_in_seconds;
+        } else {
+            time_in_seconds = 3600;
```
jgallagher59701, line 98: Use a constant like 'REMOTE_RESOURCE_DEFAULT_EXPIRED_INTERVAL' here instead of 3600.

http/RemoteResource.h
```diff
@@ -40,8 +40,9 @@
 
 #include "RemoteResource.h"
 #include "rapidjson/document.h"
+#include "HttpCache.h"
 
-#define REMOTE_RESOURCE_DEFAULT_EXPIRED_INTERVAL 3600
+//#define REMOTE_RESOURCE_DEFAULT_EXPIRED_INTERVAL -1
```
jgallagher59701, line 45: put a #define for REMOTE_RESOURCE_DEFAULT_EXPIRED_INTERVAL in HttpNames.h

jgallagher59701 (changes requested):

ndp-opendap (changes requested):
James' suggestions are good, then it's done.

http/HttpCache.h
sawer297: its static because i just copy/pasted the other method that retrieves the size and changed it to suit

PR #485, opened 2021-03-30
Trusted cmr

ndp-opendap:


jgallagher59701:
Should we merge this?

jgallagher59701:
I think our Travis CI/CD does not actually need to run SonarScan.

ndp-opendap:
Also GitGuardian is false positive on the "secrets" - I wrote them about it and they replied:
```
Good evening Nathan,

First of all, thank you for using GitGuardian and for your feedback email.

Indeed, GitGuardian's detection engine raises a false alert for the input you sent.
Note that the alert raised is not of type aws_iam but it is a generic_high_entropy_secret.

The generic high entropy secret detector looks for a very wide range of credentials to increase our recall beyond well known providers and regex based detection. I strongly encourage you to have a look at the corresponding documentation as this may help to understand why and how we designed this detector. However, this approach sometimes raises false positives.

GitSecrets indeed raises no alerts on the given input. But the bad news is it would not raise any alert on this similar kind of input :
curl "https://api.my_website.com/v2/me?token=XYZERGHABGHU354HKJ" 

Our generic high entropy detector has your back in this kind of use case.
Finally, were you not satisfied by this generic detector, note that you can disable it in the Settings/Secrets Detection and Policy section of GitGuardian's dashboard.

Anyway, our team will preciously keep your feedback to improve our generic high entropy detector.

Best regards,

Eric LACAILLE
```

Issue #450, opened 2020-12-09
update the Hyrax server functions so that they use the new STAREmaster sidecar files


captainkirk99:
update the Hyrax server functions so that they use the new STAREmaster sidecar files

captainkirk99:
The first question is: where are the sever functions that need to be updated?

jgallagher59701:
Done.

PR #487, opened 2021-04-01
Working on reading sidecar files

captainkirk99:
Part of #450 

Now reading sidecar file...


captainkirk99:
OK, this is being taken down because we cannot yet include STAREmaster in hyrax-dependencies. When we can, then we will need some of these changes.

PR #490, opened 2021-04-08
Hyrax 388

kyang2014:
Fix the bug that the in-memory compression is not turned on when an netCDF-4 file output is requested for a DAP2 Grid.


ndp-opendap:
Tests passed merging...

PR #494, opened 2021-04-22
Change to new sidecar file reading routine, turn off tests depending on old sidecar files

captainkirk99:
We have change the sidecar files to netCDF-4 (and made other changes).

The tests under functions/stare used old sidecar files, in HDF5, which are no longer valid.

I have changed the bes code to use the new files, and in this PR commented out tests that depend on the old sidecar files. I will circle back around and get these tests working again in future PRs.


jgallagher59701:
I restarted the travis build that failed - We likely will need to fix our build system WRT dockerhub. However, I would like it if you would add tests to this branch without me merging it first. I have the old notebook working with the old sidecars and I'd like to checkout this branch and see how far I can get  with this code.

jgallagher59701:
Ed, I'm confused about this branch and what it contains. Are there tests? How close is this code to be complete enough to run the J notebook I developed as a simple demonstration of the functions?

PR #498, opened 2021-05-10
Added new implmentation of temp file transfer code for fileout_netcdf

ndp-opendap:


dispatch/BESUtil.cc
```diff
@@ -50,6 +50,7 @@
 
 #include <sstream>
 #include <iostream>
+#include <iostream>
```
jgallagher59701, line 53: remove duplicate include

modules/fileout_netcdf/FONcTransmitter.cc
```diff
@@ -588,4 +592,4 @@ void FONcTransmitter::write_temp_file_to_stream(int fd, ostream &strm) //, const
         nbytes = read(fd, block, sizeof block);
     }
 }
-
+#endif
```
jgallagher59701, line 595: Add newline

jgallagher59701 (changes requested):

PR #502, opened 2021-05-14
Hyrax 449 - comprenhensive data type check and implementation for the fileout netCDF module

kyang2014:
HYRAX-449, A comprehensive check of datatype match and implementation of ensuring _FillValue attribute type the same as the variable type. This check-in includes:
    1. Ensure the DAP4 to fileout netCDF-4 enhanced model data type one to one match for both variables and attributes, including ensuring _FillValue attribute match the variable type. The datatype includes all integer and float types DAP4 support.
    2. The implementation that ensures the _FillValue match is largely updated for DAP2 Array
    3. The implementation that ensures the _FillValue match is added for DAP4 Array
    4. The implementation that ensures the _FillValue match is added for scalar type.
    5. Two tests that includes all the integer and float types data(both array and scalar) and the intentionally mismatched _FillValue attribute types are added. The file is an HDF5 file.
        It tests
        1) the HDF5-CF-DMR to fileout-netCDF4 enhanced model
        2) the HDF5-default-DMR to fileout-netCDF4 enhanced model
    6. Also correct a few inconsistent datatype mapping between the following cases, mostly for DAP2 and classic model.
        1) DAP2 variable to fileout netCDF variable
        2) DAP2 attribute of a variable to fileout netCDF attributes of the same variable
    7. More tests will be added and code will be cleaned up in the next PR.


ndp-opendap:
This TravisCI build requests from this branch is are being rejected. There may be a problem with travis.yml that needs to be fixed.

jgallagher59701 (approved):
The changes look good. We have to figure out why Travis build is being rejected.

jgallagher59701:
Well, that's annoying. Apparently when Kent created the PR Travis has some rule that it doesn't start a build, but when I make the PR, Travis does.Since we pay for Travis now, it's (probably) limiting builds to only those who are formally part of Travis contract.

PR #505, opened 2021-05-21
Fonc refactor v2

sawer297:
streaming fileout netcdf responses


jgallagher59701:
If this passes CI/CD, lets squash and merge. We should probably add more tests since I found a bug using a dataset with multiple grids all sharing the same maps.

PR #509, opened 2021-05-27
No cout

ndp-opendap:
As an experiment I am dropping the use of the `stdout` stream out in favor of a regular `std::ostream` that we initialize with our custom `std::streambuf` implementation, **`PPTStreamBuf`**


ndp-opendap:
Unfortunately I have to merge this to really test it.

PR #513, opened 2021-06-08
Server handler refactor

ndp-opendap:


ppt/PPTServer.cc
```diff
@@ -237,7 +237,7 @@ void PPTServer::authenticateClient()
 #if defined HAVE_OPENSSL && defined NOTTHERE
 	BESDEBUG( MODULE, prolog << "Requiring secure connection: port = " << _securePort << endl );
 	// let the client know that it needs to authenticate
-	send( PPTProtocol::PPTSERVER_AUTHENTICATE );
+	send( PPTProtocol::PPT_SERVER_AUTHENTICATE );
```
jgallagher59701, line 240: Should PPT_SERVER_AUTHENTICATE be prefixed with PPTProtocol::? 
Same question on lines 251 and 256.

jgallagher59701 (commented):
Look at my question re lines in PPTServer.cc. Otherwise, It looks fine.

James

ppt/PPTServer.cc
ndp-opendap: I think the answer is that it matters not. That block of code might as well be inside of `#if 0` instead of `#if defined HAVE_OPENSSL && defined NOTTHERE`

PR #515, opened 2021-06-14
Hyrax 467

kyang2014:
1. Ensure  the filenetCDF generates the correct variable values for the classic model. 
2. Update the testsuite.
3. Fix a memory leaking in DapFunctionutils.cc. This was discovered during the test of fileout netcdf.


jgallagher59701 (approved):
If this passes the tests let's merge it.

Thanks

PR #516, opened 2021-06-16
HYRAX-467, 1) For the classic model, ensure the _fillvalue datatype t…

kyang2014:
…o be the same as the variable datatype.

2) Add a testing file and use the HDF5 handler to access the file to ensure the fillvalue is corrected.
3) Clean up the code,use #if !NDEBUG if possible.
4) Update the testsuite for the NASA files.
5) Make the BES.FollowSymLinks=Yes in the configuration file. This makes the testsuite for NASA files easily.


kyang2014:
The tests get passed. just merge the pull request.

PR #530, opened 2021-08-16
HYRAX-528,HYRAX-530, 1) DAP4: fix the segmentation fault issue when t…

kyang2014:
…he string attribute value is NULL. 2) History attribute: fix the issue when the previous history attribute information is not included. Nathan provided the fix.

3) Fixed a memory leaking issue when a DMR is allcoted with unique pointer but not deleted by fileout netcdf.


jgallagher59701:
Looks good - merge.

jgallagher59701 (approved):

PR #531, opened 2021-08-19
Freeform support fix

jgallagher59701:
This adds support for ancillary DAS files with compressed data. If the data are data.ext.gz, the DAS file must be named data.ext.das. Tests added for DAS, DMR, DODS and DAP responses.

Also, I 'normalized' the tests, modifying them to use the 'universal' handler_tests_macros.m4' file in bes/modules. Remove the local m4 file if these tests pass.


ndp-opendap (approved):
Yay... FF!

PR #533, opened 2021-09-02
First pass at truncating the download file on retries.

ndp-opendap:
But what happens when it's not a file? 
What happens to the "error doc" when we retry a request that is writing to a memory buffer and not a file? 
Does it make sense to move the retry cycle out of super_easy_perform() and into its callers?


jgallagher59701 (approved):
Lets merge this so it can be deployed on UAT.

PR #534, opened 2021-10-13
Hyrax 532

kyang2014:
The HDF5 handler is made to run CF and CF DMR by default. The fileout netcdf and NcML configuration files need to be updated.

The NASA testsuite for the fileout netCDF will be checked in later.

### DIFF
--- a/modules/fileout_netcdf/tests/bes.nc4.cfdmr.conf.in
+++ b/modules/fileout_netcdf/tests/bes.nc4.cfdmr.conf.in
@@ -88,7 +88,7 @@ H5.EnableCoorattrAddPath=true
 H5.RmConventionAttrPath=true
 #H5.EnableDMR64bitInt=true
 H5.NoZeroSizeFullnameAttr=true
-#H5.EnableCFDMR=true
+H5.EnableCFDMR=true
 
 H5.MetaDataMemCacheEntries=300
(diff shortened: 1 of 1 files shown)


ndp-opendap:
@kyang2014 said:
> The fileout netcdf and NcML configuration files need to be updated.
Has this been done? 
If not I think it's part of this ticket unless there are already specific tickets to accomplish this update.

jgallagher59701:
> @kyang2014 said:
> 
> > The fileout netcdf and NcML configuration files need to be updated.
> 
> Has this been done?
> 
> If not I think that it should be done as part of this ticket unless there are already specific tickets to accomplish this update task for the other modules.

I think I agree. 

Kent, do you think these can be combined? If not, will the tests fail?

kyang2014:
> > @kyang2014 said:
> > > The fileout netcdf and NcML configuration files need to be updated.
> > 
> > 
> > Has this been done?
> > If not I think that it should be done as part of this ticket unless there are already specific tickets to accomplish this update task for the other modules.
> 
> I think I agree.
> 
> Kent, do you think these can be combined? If not, will the tests fail?

Yes, the tests will not fail since I make sure the correct HDF5 BES keys are set for these tests.  I double checked them at two different machines. They both get passed. These should be merged.

PR #546, opened 2021-11-01
Added new dmr++ metadata including bes and libdap version

ndp-opendap:
Added dmr++ production metadata, including the bes and libdap versions and TheBESKeys contents to a top level attribute table in the dmr++ called  `build_dmrpp_meta` 
This is controlled by the new `-M` switch for build_dmrpp. The `get_dmrpp` script has been changed to always utilize the `-M` switch when invoking `build_dmrpp`

### DIFF
--- a/modules/dmrpp_module/build_dmrpp.cc
+++ b/modules/dmrpp_module/build_dmrpp.cc
@@ -32,7 +32,6 @@
 #include <libgen.h>
 
 #include <D4Attributes.h>
-
 #include <Array.h>
 
 //#define H5D_FRIEND		// Workaround, needed to use H5D_chunk_rec_t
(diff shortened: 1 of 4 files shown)


modules/dmrpp_module/build_dmrpp.cc
```diff
@@ -834,7 +898,7 @@ int main(int argc, char *argv[]) {
         // given HDF5 dataset
         if (!dmr_name.empty()) {
             // Get dmr:
-            unique_ptr<DMRpp> dmrpp(new DMRpp);
+            shared_ptr<DMRpp> dmrpp(new DMRpp);
```
jgallagher59701, line 901: shared_ptr<> works here, but it is less efficient than unique_ptr<>. In some of the profiling, I saw shared_ptr<> using much more time that I originally expected.

jgallagher59701 (approved):
Merge this but maybe change back to unique_ptr<>.

PR #549, opened 2021-11-09
adding version as a state variable in DMRpp

ndp-opendap:
This not done and I ran into the libdap `DMR::dmr_version()` ,method. the value of  `d_dmr_version` is serialized in the `<Dataset>` element's XML attribute `dmrVersion`.

Should we just use that and skip what I was doing here?

    /// The version of the DMR document
    std::string d_dmr_version;


jhrg:
I'm going to close this one since I merged this into my branch that uses the version information

PR #550, opened 2021-11-09
Filter back patch

jgallagher59701:
This branch adds a hack that looks in the DMR++ <Dataset> element for a 'dmrpp:builderVersion' attribute. If it is not found, we assume the DMR++ is an older (vintage) document where filter order is treated as an afterthought. The handler now corrects 'bad' filter order in these old docs. If the doc has the attribute, the handler assumes the order is correct.

In the future, we can be more picky about using the version number - for now the attribute is basically a boolean.


jgallagher59701:
I'm going to merge this so we can get on with it.

PR #552, opened 2021-11-13
Dmrpp refactor 42

jgallagher59701:
Code reduction. Done here but destined to be merged into the dmrpp-lazy-eval branch.

Does this pass CICD?


jgallagher59701:
NB: We can tighten up this code quite a bit and that will help us make the lazy eval changes. Also, duplication in the code suggests that we might look at revamping the DHI so it makes these kinds of refactorings simper and so it uses shared_ptr instead of its own bespoke scheme to manage dynamic memory.

PR #554, opened 2021-12-08
Fix libdap paths

jgallagher59701:
Some of the libdap headers - e.g., Array.h - clash with newer C++ headers - array.h - on OSX, where the file system 'preserves' but does not distinguish case. To fix this, and to generally address that some libdap headers clash with other librarys' headers (util.h, debug.h) I added the prefix 'libdap/' to all of the libdap includes in the bes source code.


jgallagher59701:
Yo, can we merge this?

PR #566, opened 2022-01-19
Fix for the issue where OSX 11 dlopen() fails in some cases.

jgallagher59701:
For libgdal, dlopen fails to find libcurl when the reference is within
a library loaded as a run-time loadable module. The fix is to switch
from RTLD_NOW to RTLD_LAZY.

This PR addresses an issue on ISX that is a blocker for development there.


dh-opendap:
Looks fine, I have used this fix locally and all BES tests pass on my Big Sur based platform, merge.

PR #569, opened 2022-01-21
Changed how the code links to the dap_module to fix 'non-portable' warnings

jgallagher59701:
The code was linking against dap_module.la which becomes dap_module.so
which is actually a 'module' and not a real library. This generates
warnings from libtool and also, on linux, with rpath. The changes are
mostly in unit-tests but also appears in the dmrpp tools like build_dmrpp.


dh-opendap:
I've run this branch locally on my os/x platform and all the tests pass.  Looks good.

PR #572, opened 2022-02-04
Hyrax 610

kyang2014:
covjson to support AIRS level 3 and GLDAS level 4.


kyang2014:
Tested at MacOS. The travis test gave fake failure.

PR #573, opened 2022-02-07
Use the covjson module's static archive to link the unit test

jgallagher59701:
Also fixed some warnings in the code where an error case was not handled
and suppressed annoying warnings about arrays and C99 using -Wno-vla-extension
where 'no' turns off the warning.

### DIFF
--- a/modules/fileout_covjson/FoDapCovJsonTransform.cc
+++ b/modules/fileout_covjson/FoDapCovJsonTransform.cc
@@ -231,64 +231,61 @@ unsigned int FoDapCovJsonTransform::covjsonSimpleTypeArrayWorker(ostream *strm, 
                 if(is_axis_t_sgeo) { 
                     string axis_t_value;
                     std::ostringstream tmp_stream; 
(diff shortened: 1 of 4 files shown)


jgallagher59701:
We should plan on backing out the -Wno-vla switch, but for now, it quiets these warnings and helps us see other things.

NB: https://stackoverflow.com/questions/61825661/why-does-clang-complain-about-using-variable-length-arrays-with-std-c99-flag: You are correct that conforming C99 implementations must support VLAs, whereas implementations can conform to later C standards without supporting VLAs. But I think you're missing the forest for the trees: I take the aforementioned difference between the standards to be exactly the point of the warning. It's not saying that your code, as you are now building it, may break. Rather, it is warning that you are relying on a feature that is not universally forward compatible, so that if you try to build your code somewhere else then it may break.

If Clang thought that VLAs were a non-standard extension to C99, then it ought not to accept your VLA-using code at all in -std=c99 mode.

I'll merge this now.

PR #579, opened 2022-02-16
HFRHANDLER-335, update the BES configuration files for the testing un…

kyang2014:
…der the modules/hdf4_handler.


kyang2014:
Only update the BES configuration files for the test.

PR #584, opened 2022-02-25
Get dmrpp s3

ndp-opendap:
Updates get_dmrpp so that it will recognize an s3:// URL as the input file and if an S3 URL is identified `get_dmrpp` will:
- Check the the AWS configuration for credentials and exit with an error if they are not present.
-  Download the S3 object to the local disk
- Build a dmr++ from the local copy of the S3 object.
- Optionally upload (-U) the new dmr++ file to the same S3 bucket with the same object and a suffix of `.dmrpp`


jgallagher59701 (approved):
Merge. Thanks.

PR #586, opened 2022-03-01
History mds fix

jgallagher59701:
To support the history_json (netcdf) attribute, we made a pretty extensive modification to the fileout_netcdf handler. 
That was good, and although I think the change could have been a bit simpler, the code we have now works and testing
new  code will be fraught since the feature involves one or maybe two other groups (Harmony and Data Services?).

When we implemented the history json attribute feature, we made two tries, and we left the start of the first
attempt in the netcdf_handler. This change was to _stop_ escaping the values of the string attributes. In practice 
this has  _no_ effect on the value of the history_json attribute (although it took me two days to sort that out). However,
it causes the MDS to fail since DAS responses in the cache cannot be parsed by the libdap DAS parser (since they are not 
syntactically valid). While it might be _nice_ to simplify the history_json code, no change was actually needed. I just 
turned off the errant behavior in the netcdf_handler and the problem with the MDS went away. Meanwhile, there was no
change in the behavior of the fileout_netcdf code.


jgallagher59701:
As a bonus, I worked on the history_utils unit tests (before I realized no changes were needed in fileout_netcdf). In doing
so, I wrote a new template function that will let us replace the wordy and repetitious main() in the unit test c++ code with a one-line function call. Look at bes/modules/bes_run_tests_cppunit.h.

jgallagher59701:
I'd like to merge this - please review.

@kyang2014 Note there are some change to the netcdf_handler tests: I have combined the two testsuites to simplify the Makefile.am and also made changes to reflect that we no longer think of netCDF4 as an optional feature by, instead, as required.
@ndp-opendap I made a handful of changes that increase the use of C++11, including the use of initializers in the class definition. This is the preferred way to initialize class members to constant values. See FONcTransform.cc

jgallagher59701:
> The scope of this PR makes it really difficult to see what is happening. It passes tests which is great, but I continue to have concerns that these changes may break things downstream for David Auty et al. 🤷 Unless we make a "test-deploy" condition in TravisCI (like I did for the OLFS) and deploy the build to UAT before merging I will have to accept the risk of it breaking these things and then if it does break them we'll just fix it later...

True. But, the real change - besides writing more tests and fixing some 'code smells' is that the netcdf handler has the original behavior of escaping attributes. That means baselines in a bunch of places changed, but the code for the history_json attribute didn't change and - this is the key bit - the tests for the fileout_netcdf handler didn't change.

So how long will it take to do the test-deploy thing? Can you help?

jgallagher59701:
Which is to say, how would I do this and actually test the code with Data Services?

modules/fileout_netcdf/FONcTransform.cc
```diff
@@ -509,7 +345,8 @@ void FONcTransform::transform_dap2(ostream &strm)
 
     stax = nc_set_fill(_ncid, NC_NOFILL, &current_fill_prop_vaule);
```
kyang2014, line 346: I didn't review the code when this line was added. That's why it might be a good idea to ask somebody to review the code. It seems that this purpose is for performance. However, with recent empty chunk value case for GHRSST, if the whole chunk is all covered with fillvalue, set a real fillvalue may have the performance gain. Again, this is the same issue as passing the compression ratio/shuffle from upstream file, which I did an experiment at 
https://github.com/oPENDAP/libdap4/tree/esdis-580
and https://github.com/OPENDAP/bes/tree/esdis-580

modules/fileout_netcdf/tests/Makefile.am
```diff
@@ -74,6 +74,7 @@ $(srcdir)/package.m4: $(top_srcdir)/configure.ac
 	echo 'm4_define([AT_PACKAGE_VERSION],   [@PACKAGE_VERSION@])' && \
 	echo 'm4_define([AT_PACKAGE_STRING],    [@PACKAGE_STRING@])' && \
 	echo 'm4_define([AT_PACKAGE_BUGREPORT], [@PACKAGE_BUGREPORT@])'; \
```
kyang2014, line 76: Have you tested this? Do they work? 
should you have && and remove ;
Something like: 
 echo 'm4_define([AT_PACKAGE_BUGREPORT], [@PACKAGE_BUGREPORT@])' && \

modules/gateway_module/tests/gateway/fnoc1.nc.nc.das.bescmd.baseline
```diff
@@ -28,9 +28,8 @@ Attributes {
     NC_GLOBAL {
         String base_time "88- 10-00:00:00";
         String title " FNOC UV wind components from 1988- 10 to 1988- 13.";
-        String history "removed date-timehyrax-1.16.4 http://test.opendap.org/opendap/data/nc/fnoc1.nc.nc
-";
-        String history_json "[{"$schema":"https:\/\/harmony.earthdata.nasa.gov\/schemas\/history\/0.1.0\/history-0.1.0.json","date_time":"removed date-time+0000","program":"hyrax","version":"1.16.4","parameters":[{"request_url":"http:\/\/test.opendap.org\/opendap\/data\/nc\/fnoc1.nc.nc"}]}]";
+        String history "removed date-timehyrax-1.16.4 http://test.opendap.org/opendap/data/nc/fnoc1.nc.nc\012";
```
kyang2014, line 31: A question: Why there is  \012 added at the end of  new "history" attribute?

kyang2014 (commented):
I appreciate you add the history_utility unit tests. That reminds me to add a similar unit test for the fileout coverage json time conversion routine. That's out of BES and DAP.

modules/bes_run_tests_cppunit.h
```diff
@@ -0,0 +1,103 @@
...
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+//
+// You can contact OPeNDAP, Inc. at PO Box 112, Saunderstown, RI. 02874-0112.
+
+/**
+ * This file contains a templated function that serves as the body of main()
+ * for a CppUnit-based unit test.
+ *
+ * @note This is a WIP now but is intended to cut done on duplicated code in both
```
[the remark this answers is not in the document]
jgallagher59701: fixed

modules/fileout_netcdf/FONcAttributes.cc
```diff
@@ -285,18 +283,16 @@ void FONcAttributes::add_attributes_worker(int ncid, int varid, const string &va
                     is >> uival;
                     vals[attri] = (short) uival;
                 }
-                stax = nc_put_att_short(ncid, varid, new_name.c_str(), NC_SHORT,
-                                        num_vals, &vals[0]);
+                stax = nc_put_att_short(ncid, varid, new_name.c_str(), NC_SHORT,  num_vals, &vals[0]);
                 if (stax != NC_NOERR) {
-                    string err = (string) "File out netcdf, "
-                                 + "failed to write byte attribute " + new_name;
+                    string err = (string) "File out netcdf, " + "failed to write byte attribute " + new_name;
```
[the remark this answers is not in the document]
jgallagher59701: right

modules/fileout_netcdf/FONcTransform.cc
```diff
@@ -195,56 +191,22 @@ FONcTransform::FONcTransform(BESResponseObject *obj, BESDataHandlerInterface *dh
...
 
-
 /** @brief Destructor
  *
  * Cleans up any temporary data created during the transformation
  */
-FONcTransform::~FONcTransform()
-{
+FONcTransform::~FONcTransform() {
     for (auto &b: _fonc_vars) {
```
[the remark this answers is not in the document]
jgallagher59701: I did not test with valgrind, but I did test with leaks (by hand)

modules/fileout_netcdf/tests/Makefile.am
jgallagher59701: Good catch  this did work, but I should use && there and not ;

modules/gateway_module/tests/gateway/fnoc1.nc.nc.das.bescmd.baseline
jgallagher59701: We returned to escaping the character in the attributes. This was the behavior until 1.16.3 when we stopped escaping the attrs. We did that as part of the history_json work (although we abandoned the work in netcdf_handler and implemented history_json in a completely different way here) and was left in place by accident.

kyang2014: Is this escaping feature also back to the HDF handlers? I also slacked you the same question.

kyang2014 (approved):

PR #587, opened 2022-03-03
Dmrpp string array fix

jgallagher59701:
A quick hack - lets see how this does with CI/CD.

There's much more work to do.


jgallagher59701:
I'm going to merge this - it's a simple patch that gets one case working but there's so much more to do because the DMR++ builder needs to be modified so that the DMR++ holds additional information about strings. There is an edge case where string length is described as 'variable' that will require research into how HDF5 _stores_ string data.

PR #590, opened 2022-03-09
Adding improved verbose and error output to build_dmrpp

ndp-opendap:


jgallagher59701:
Merge. 

I looked at the one new code smell and marked it as won't fix. I think their rule than any code block in a switch statement should be moved to a function is overdoing it. Certainly not for that although we do have a solution for appending an integer to a c++ string that does not depend on ostringstream. It's in the util code and it's a function like append_long_to_string(). 6 to one, half a dozen to the other.

PR #593, opened 2022-03-11
Esc attr 536 delta

jgallagher59701:
Hopefully this will run through CI/CD...


modules/hdf5_handler/bes-testsuite/hdf5_handlerTest.at
```diff
@@ -679,3 +739,13 @@ AT_BESCMD_CFDMR_RESPONSE_TEST([d_int64.h5.dmr.bescmd])
 AT_BESCMD_CFDMR_DAP4DATA_RESPONSE_TEST([d_int64.h5.dap.bescmd])
 AT_BESCMD_CFDMR_RESPONSE_TEST([t_scalar_dset.h5.dmr.bescmd])
 AT_BESCMD_CFDMR_DAP4DATA_RESPONSE_TEST([t_scalar_dset.h5.dap.bescmd])
+
+# These six tests use the new macro added at the top if the file. jhrg 3/12/22
+# tests for the H5.EscapeUTF8Attr bes key - using the default value of true first
+AT_BESCMD_BESCONF_RESPONSE_TEST([h5.cf/utf8_attr.h5.das.bescmd], [bes.cfdmr.conf])
+AT_BESCMD_BESCONF_RESPONSE_TEST([h5.cf/utf8_attr.h5.ddx.bescmd], [bes.cfdmr.conf])
+AT_BESCMD_BESCONF_RESPONSE_TEST([h5.cf/utf8_attr.h5.dmr.bescmd], [bes.cfdmr.conf], [xfail])
```
kyang2014, line 747: Why this is the expected failure? I don't understand.

kyang2014 (commented):
In the "make check", I see something like:

```
360: h5.cf/utf8_attr.h5.das.bescmd                   ok
361: h5.cf/utf8_attr.h5.ddx.bescmd                   ok
362: h5.cf/utf8_attr.h5.dmr.bescmd                   expected failure (hdf5_handlerTest.at:747)
363: h5.cf/utf8_attr.h5.das.bescmd                   ok
364: h5.cf/utf8_attr.h5.ddx.bescmd                   ok
365: h5.cf/utf8_attr.h5.dmr.bescmd                   ok

```
As in the first comment, I don't understand why the dmr test is the expected failure. 
Also the test "363-365" is supposed to test when H5.EscapeUTF8Attr = false. The output message is the same. After a while, we may forget.

jgallagher59701:
I’ll have to check this. The output should not be the same. As for the DMR, that test does look wrong to me. I think I need to look this over and we need to talk about it. 

I may not make it to stand up - I on vacation Monday to Thursday. 

James

--
James Gallagher
jgallagher at opendap.org

modules/hdf5_handler/bes-testsuite/hdf5_handlerTest.at
jgallagher59701: It turns out that there is a second function for the DMR attributes that I did not edit. Fixed now.

PR #594, opened 2022-03-21
Remove the old WWW interface - we now use code in the front end.

jgallagher59701:
We no longer need this code. Will the BES pass CI/CD without it?


ndp-opendap (approved):
Kooks great, retire away!
:shipit:

PR #595, opened 2022-03-23
Refactor of the code in bes/server

jgallagher59701:


ndp-opendap (commented):
Looks like it’s mostly (completely?) a C++11 update along with an expressed aversion to camel case in variable/function/method names.

Did I get that right?

ndp-opendap (approved):

jgallagher59701:
> Looks like it’s mostly (completely?) a C++11 update along with an expressed aversion to camel case in variable/function/method names.
> 
> Did I get that right?

Yes

PR #597, opened 2022-03-28
Bad alloc test

jgallagher59701:
I'd like to know if this branch passes CI/CD


jgallagher59701:
This PR does fix the bad_alloc, but it does add information to the log about the memory use at the time of the error. This can help us debug this issue. I fixed the issue where OSX returns bytes for the RSS with getrusage() while the rest of the world returns KB. Now the message always shows KB.

ndp-opendap (approved):
Looks niiiiice.

:shipit:

PR #603, opened 2022-04-05
Ncml auto ptr begone

jgallagher59701:


jgallagher59701:
I'll look at the 'code smells' but I think this is far enough along that we can call it good. There are no more instances of auto_ptr in the code and I think some of the more convoluted calls have been smoothed out.

PR #604, opened 2022-04-07
Size fix

jgallagher59701:
Apple's version of getrusage() returns size in bytes, not KB like the rest of the world (?). I fixed this with a #ifdef __APPLE__ ... block. At the same time I moved Sam's code into a new file so that we can more easily use it in the 'Transmitter handlers' at some point in the future.

The new file wraps the functions in a namespace instead of using a class filled with static methods.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -254,6 +254,8 @@ add_executable(bes
     dap/ShowPathInfoResponseHandler.h
     dap/TempFile.cc
     dap/TempFile.h
+		dap/DapUtils.cc
+		dap/DapUtils.h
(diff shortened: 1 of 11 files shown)


jgallagher59701:
Still to do, add use of this by our various transmitter handlers (ASCII, NetCDF, ...) or find a cool place to embed it in the framework. The code has to be called after the CE is parsed and before we are in the process of actually reading the data (because that might trigger and OOM error and we're trying to see what the projected memory use will be before that happens).

ndp-opendap (approved):
:shipit:

PR #608, opened 2022-04-08
Standalone fix

jgallagher59701:
Fixes for the StandAloneApp class. 

I also removed the unused BESDapErrorInfo class and an annoying error in the FFND code.


ndp-opendap (approved):
Oy... Again with the massive white space mods...

:shipit:

PR #609, opened 2022-04-11
HYRAX-667, change NULL to nullptr for the HDF4 handler. This can redu…

kyang2014:
…ce the code smell and also easy to find other issues from sonar cloud.

### DIFF
--- a/modules/hdf4_handler/HDF4RequestHandler.cc
+++ b/modules/hdf4_handler/HDF4RequestHandler.cc
@@ -258,7 +258,7 @@ bool HDF4RequestHandler::hdf4_build_das(BESDataHandlerInterface & dhi) {
             int32 sdfd    = -1;
             int32 fileid  = -1;
 
-            HDFSP::File *h4file = NULL;
+            HDFSP::File *h4file = nullptr;
 
             // Obtain HDF4 file IDs
             //SDstart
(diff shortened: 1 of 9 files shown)


modules/hdf4_handler/HDF4RequestHandler.cc
```diff
@@ -327,7 +327,7 @@ cerr<<"Don't output ecs metadata "<<endl;
                 //throw InternalErr(__FILE__,__LINE__,"read_das_use_eos2lib error");
 #endif
             }
-            if(eosfile != NULL)
+            if(eosfile != nullptr)
```
jgallagher59701, line 330: There's no pressing need to change this, but as you write new code, there is no need to test if the pointers are not null before calling 'delete' since the delete operator can safely be used on null pointers (unlike free).

jgallagher59701 (approved):
As I said in my one comment, no need to change those - the gain in efficiency is very small, but in the future you can just dall 'delete pointer' and drop the test since delete is safe to use with null pointers.

modules/hdf4_handler/HDF4RequestHandler.cc
kyang2014: Thanks,  this will be handled later and I am going to remove the "new delete" with unique_pointer later. There are other cases that NULL needs to be changed to nullptr (like passing NULL as a parameter) if we want to conform to C11.  Also sonor cloud gives a lot of code smell. It buries other good warnings.

> here is no need to test if the pointers are not null before calling 'delete' since the delete operator can safely be used on null pointers (unlike free).

PR #610, opened 2022-04-12
Add a README file to the top level of our test data hierarchy

ndp-opendap:
Because we don't have one now and I need something to test the retrieval of non data files from Hyrax.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1525,7 +1525,7 @@ add_executable(bes
     modules/ncml_module/OtherXMLParser.h
     modules/ncml_module/RCObject.cc
     modules/ncml_module/RCObject.h
-    modules/ncml_module/RCObjectInterface.cc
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -1525,7 +1525,7 @@ add_executable(bes
     modules/ncml_module/OtherXMLParser.h
     modules/ncml_module/RCObject.cc
     modules/ncml_module/RCObject.h
-    modules/ncml_module/RCObjectInterface.cc
+    # modules/ncml_module/RCObjectInterface.cc
```
jgallagher59701, line 1528: Good catch.

modules/data/README
```diff
@@ -0,0 +1,13 @@
+###############################################################################
```
jgallagher59701, line 1: Nice

jgallagher59701 (approved):
Merge when CMR is sorted and the tests pass.

PR #611, opened 2022-04-13
Makefile.am was 'cleaning' the testsuite scripts

jgallagher59701:
Cleaning these scripts with a parallel make meant that
in some cases the testsuite script was removed before it
could be run as per 'clean-local' resulting a failure.
Since the testsuite script is included in the tar ball,
I removed it from CLEANFILES in three Makefile.am files.
This seems to fix the issues with 'make clean --jobs=12'
failing.


jgallagher59701:
This is a fix for the make clean error you found a while ago.

ndp-opendap (approved):
OMG Finally! This problem has been breaking tests for sooooo long!

So, when the tests pass you must :shipit:!!

PR #612, opened 2022-04-13
H5 clean warning

kyang2014:
In addition to cleaning some warnings, fix a bug in the default option when handling string attribute. extra nulls were added to string attribute. This was not detected by DAS/DMR output but by the fileout netCDF ncdump -h when more than one string element is present in the attribute.


modules/hdf5_handler/h5commoncfdap.cc
```diff
@@ -724,41 +723,27 @@ void gen_dap_onevar_dmr(libdap::D4Group* d4_grp, const HDF5CF::Var* var, const h
         BaseType *bt = nullptr;
 
         switch (var->getType()) {
-            // TODO Remove extra ';' jhrg 3/9/22
 #define HANDLE_CASE(tid,type)                                  \
             case tid:                                           \
                 bt = new (type)(var->getNewName(),var->getFullPath()); \
             break;
```
jgallagher59701, line 729: Another way to handle this would be to use an inline template function. This way, using a macro, is fine, but I thought I'd mention the alternative for completeness sake.

modules/hdf5_handler/h5get.cc
```diff
@@ -524,8 +524,8 @@ void get_dataset(hid_t pid, const string &dname, DS_t * dt_inst_ptr)
         throw InternalErr(__FILE__, __LINE__, msg);
     }
 
-    hsize_t size[DODS_MAX_RANK];
-    hsize_t maxsize[DODS_MAX_RANK];
+    hsize_t size[ndims];
```
jgallagher59701, line 527: Could these be vector<hsize_t> as above on lines 216?

jgallagher59701 (approved):
One suggestion, but it's more of a question since the changes cover a number of files and it's hard to see the whole file in teh github review window.

I think these are really good changes over all and I notice the warnings are down quite a bit.

jgallagher59701:
I think you should merge this if the tests pass. If they don't, lets look at why and and see where to go from there.

modules/hdf5_handler/h5get.cc
kyang2014: Oh, somehow those were skipped. changed them  and committed.

modules/hdf5_handler/h5commoncfdap.cc
kyang2014: I will talk with you about this when I have another round of clean-up.

PR #613, opened 2022-04-17
Unit tests fix

jgallagher59701:
Use the new unit-test main<>() copied/modified from lib dap


dh-opendap (approved):
Looks like a worthy improvement, the templating for the test class in run_tests_cppunit.h.

PR #614, opened 2022-04-18
Hyrax 687

kyang2014:


modules/hdf5_handler/HDF5CF.h
```diff
@@ -798,7 +806,7 @@ class File {
     File(const char *h5_path, hid_t file_id) :
         path(std::string(h5_path)), fileid(file_id), rootid(-1), unsupported_var_dtype(false), unsupported_attr_dtype(false), unsupported_var_dspace(
             false), unsupported_attr_dspace(false), unsupported_var_attr_dspace(false), addeddimindex(0), check_ignored(
-            false), have_ignored(false), have_udim(false),_is_dap4(false)
+            false), have_ignored(false), have_udim(false),_is_dap4(false),iscoard(false)
```
jgallagher59701, line 809: There is a feature in C++-11 that can be used to simplify writing these constructors where fields like 'iscoard' can be initialized where they are defined in the header and don't need to be included in the constructors. The notion is that those initial values don't have to be repeated for each constructor. The SonarScan will include this under what it calls a 'code smell.'
There's no need to change this now, but it's something to keep in mind that might help writing new code easier.

modules/hdf5_handler/h5commoncfdap.cc
```diff
@@ -1632,6 +1634,113 @@ void add_var_dap4_attr(BaseType *var,const string& attr_name, D4AttributeType at
...
+            Array *t_a = static_cast<Array *>(*vi);
+
+            // The maps are essentially coordinate variables.
+            // We've sorted the coordinate variables already, so
+            // just save them to the vector. 
+            // Note: the coordinate variables we collect here are
+            // based on the understanding of the handler. We will
+            // watch if there are complicated cases down the road. KY 04-15-2022
+            bool is_cv = false;
+            for ( auto it_cv = coord_var_names.begin(); it_cv !=coord_var_names.end();++it_cv) {
```
jgallagher59701, line 1665: For loops that use iterators with '.begin()' and '.end()' methods, you can use the shorter:
for (auto cv : coard_cv_names) {
     ...
}

(another C++-11 feature).

Then, line 1666 becomes
if (cv == v->name()) { 
...
}

jgallagher59701 (approved):
Merge, but check over my comments. I think the two features of C++-11 are pretty useful for limiting errors.

modules/hdf5_handler/HDF5CF.h
kyang2014: Thanks, will handle this with another ticket.

modules/hdf5_handler/h5commoncfdap.cc
kyang2014: Thanks, will handle this later.

PR #616, opened 2022-04-19
Hdf5 fill value

jgallagher59701:
This is an early version of the Fill Value fix; I want to see if it passes tests on Linux.


jgallagher59701:
NB: I don't think we should merge this, but I thought it could do with a review. The code as it is now can lead into the next phase - actually using the fill value info. Note also that support for COMPACT will come in this second phase.

modules/dmrpp_module/build_dmrpp_util.cc
```diff
@@ -0,0 +1,963 @@
...
+using namespace dmrpp;
+
+namespace build_dmrpp_util {
+
+bool verbose = false;   // Optionally set by build_dmrpp's main().
+
+#define VERBOSE(x) do { if (verbose) (x); } while(false)
+
+#if 0
+///////////////////////////////////////////////////////////////////////////////
```
kyang2014, line 86: If you use this code, I think you should add a comment that they are adapted from the HDF5 handler . The same for the read_vlen_string().

modules/dmrpp_module/build_dmrpp_util.cc
```diff
@@ -0,0 +1,963 @@
...
+            // calls to H5Pclose(dcpl_id) for each error can be removed. jhrg 5/7/18
+            H5D_fill_value_t fvalue_status;
+            if (H5Pfill_value_defined(dcpl_id, &fvalue_status) < 0) {
+                H5Pclose(dcpl_id);
+                throw BESInternalError("Cannot obtain the fill value status.", __FILE__, __LINE__);
+            }
+            if (fvalue_status == H5D_FILL_VALUE_UNDEFINED) {
+                // Replace with switch(), here and elsewhere. jhrg 5/7/18
+                if (layout_type == 1)
+                    cerr << " The storage size is 0 and the storage type is contiguous." << endl;
```
kyang2014, line 308: If this file is commented out. Please ignore my review. Otherwise, I don't understand your logic here. I think you copy the code from an example code I gave to you but didn't do in the correct context. If the fillvalue is not defined, it doesn't mean the storage size is 0.  Maybe this is  only for the debugging purpose. If this is  the case, you may mark it differently.

modules/dmrpp_module/build_dmrpp_util.cc
```diff
@@ -0,0 +1,963 @@
...
+                else if (layout_type == 2)
+                    cerr << " The storage size is 0 and the storage type is chunking." << endl;
+                else if (layout_type == 3) cerr << " The storage size is 0 and the storage type is compact." << endl;
+
+                char *fvalue = NULL;
+                size_t fv_size = H5Tget_size(dtype_id);
+                if (fv_size == 1)
+                    fvalue = (char *) (malloc(1));
```
kyang2014, line 324: Why not using vector<char> to replace malloc() and free() ?

modules/dmrpp_module/build_dmrpp_util.cc
```diff
@@ -0,0 +1,963 @@
...
+                case 8:
+                    oss << *(double*)(value.data());
+                    break;
+                default:
+                    throw BESInternalError("Unable extract float fill value.", __FILE__, __LINE__);
+            }
+            break;
+
+        case H5T_STRING:
+            return "unknown"; // FIXME Add support for String fill values. jhrg 4/22/22
```
kyang2014, line 619: No need to handle fillvalue for string

modules/dmrpp_module/build_dmrpp_util.cc
```diff
@@ -0,0 +1,963 @@
...
+
+        case H5T_ARRAY:
```
kyang2014, line 621: This is even remote. Array datatype is seldom used and also not the fillvalue.

modules/dmrpp_module/build_dmrpp_util.cc
```diff
@@ -0,0 +1,963 @@
...
+            return "unknown"; // FIXME What to do for Array? jhrg 4/22/22
+
+        case H5T_REFERENCE:
```
kyang2014, line 624: fillvalue doesn't make sense for reference.

modules/dmrpp_module/build_dmrpp_util.cc
```diff
@@ -0,0 +1,963 @@
...
+        case H5T_COMPOUND:
```
kyang2014, line 625: Compound is something that may need to be considered later.

kyang2014 (changes requested):
I don't see you handle the empty datasets or empty chunks that are are filled with fillvalue yet.

kyang2014 (dismissed):

modules/dmrpp_module/build_dmrpp_util.cc
jgallagher59701: OK, but we have to handle this because some of our test data does have the fill value attribute set for H5T_ARRAY. It's odd, and I have no idea how to get the value. 

What should I do for this case? clear the fill value so that it does not appear to be set?

modules/dmrpp_module/build_dmrpp_util.cc
jgallagher59701: OK. Maybe I'll make these three cases (COMPOUND, ARRAY and STRING) "unsupported" for now so we have a mnemonic that this needs to be investigated in the future.

modules/dmrpp_module/build_dmrpp_util.cc
kyang2014: Are you sure? Where do you get the data? 

> OK, but we have to handle this because some of our test data does have the fill value attribute set for H5T_ARRAY. It's odd, and I have no idea how to get the value.

jgallagher59701: You're correct - there are no cases of ARRAY or COMPOUND in the dmrpp builder test data. There are some instances of STRING.

jgallagher59701:
I'm going to leave this PR open and start the next ticket. Thanks for the review work. I think it helped catch some issues before I went charging ahead.

ndp-opendap (commented):
Again the `#if 0`'d things make this super hard to understand.

kyang2014 (commented):
change to comment since having not reviewed the new code.

jgallagher59701:
Please review. Pay attention to the cc and h files in modules/dmrpp_module. Most of the other parts are baselines and the odd fix I made as problems appeared in the debugger during testing.

modules/dmrpp_module/data/fill_value/FValue.h5.dmr
```diff
@@ -0,0 +1,31 @@
...
+    </Int32>
+    <Int32 name="cont_all_fill">
+        <Dim size="4"/>
+        <Dim size="4"/>
+    </Int32>
+    <Int32 name="cont_some_fill">
+        <Dim size="4"/>
+        <Dim size="4"/>
+    </Int32>
+</Dataset>
```
kyang2014, line 31: Suggest you add a test to check the expression constraint across filvalue boundary. I think it  should work but a test is better to avoid any future potential issue.

modules/dmrpp_module/DMZ.cc
```diff
@@ -1062,17 +1173,53 @@ void DMZ::load_chunks(BaseType *btp)
...
     if (child) {
         chunks_found = 1;
         process_chunks(dc(btp), child);
+        auto array = dynamic_cast<Array*>(btp);
+        // It's possible to have a chunk, but not have a chunk dimension sizes element
+        // when there is only one chunk (e.g., with HDF5 Contiguous storage). jhrg 5/5/22
+        if (array && !dc(btp)->get_chunk_dimension_sizes().empty()) {
+            auto const &array_shape = get_array_dims(array);
+            size_t num_logical_chunks = logical_chunks(array_shape, dc(btp));
+            // do we need to run this code?
```
ndp-opendap, line 1188: // do we need to run this code?
Isn't the test here really for:
            // Are there missing chunks?

modules/dmrpp_module/DmrppArray.cc
```diff
@@ -946,27 +946,27 @@ void DmrppArray::read_chunks_unconstrained()
     super_chunks.push(current_super_chunk);
 
     // Make the SuperChunks using all the chunks.
-    for(const auto& chunk: get_immutable_chunks()){
+    for(const auto& chunk: get_immutable_chunks()) {
         bool added = current_super_chunk->add_chunk(chunk);
-        if(!added){
+        if (!added) {
```
ndp-opendap, line 951: Is this a style thing you want us all to adopt?

modules/dmrpp_module/DmrppArray.cc
```diff
@@ -946,27 +946,27 @@ void DmrppArray::read_chunks_unconstrained()
...
             sc_id.str(std::string());
             sc_id << name() << "-" << sc_count++;
             current_super_chunk = shared_ptr<SuperChunk>(new SuperChunk(sc_id.str(),this));
             super_chunks.push(current_super_chunk);
-            if(!current_super_chunk->add_chunk(chunk)){
+            if (!current_super_chunk->add_chunk(chunk)) {
```
ndp-opendap, line 956: ibid

modules/dmrpp_module/DmrppCommon.cc
```diff
@@ -91,11 +87,9 @@ void join_threads(pthread_t threads[], unsigned int num_threads)
             string *error = NULL;
             if ((status = pthread_join(threads[i], (void **) &error)) < 0) {
                 BESDEBUG(dmrpp_3, "Could not join thread " << i << ", " << strerror(status)<< endl);
-                // ERROR_LOG("Failed to join thread " << i << " during clean up from an exception: " << strerror(status) << endl);
```
ndp-opendap, line 94: These calls to ERROR_LOG are crucial to understanding what is happening with this code when it's deployed in the filed, specifically NGAP. I commented them out because things seemed ok, but I think we should leave them in just in case we have issues down the road. They're the right things in the right place and it's not always easy to to sort that out from scratch every time

modules/dmrpp_module/DmrppCommon.cc
```diff
@@ -91,11 +87,9 @@ void join_threads(pthread_t threads[], unsigned int num_threads)
...
             }
             else if (error != NULL) {
                 BESDEBUG(dmrpp_3, "Joined thread " << i << ", error exit: " << *error << endl);
-                // ERROR_LOG("Joined thread " << i << ", error exit" << *error << endl);
```
ndp-opendap, line 98: ibid

modules/dmrpp_module/DmrppCommon.cc
```diff
@@ -151,7 +146,7 @@ void DmrppCommon::parse_chunk_dimension_sizes(const string &chunk_dims_string)
         // Process space delimited content
         while ((strPos = chunk_dims.find(space)) != string::npos) {
             strVal = chunk_dims.substr(0, strPos);
-
+            // TODO stoull (CDS uses uint64_t) jhrg 5/2/22
```
ndp-opendap, line 149: Why not just do this??  Seems like it should be straight forward...

modules/dmrpp_module/DmrppCommon.cc
```diff
@@ -334,7 +288,20 @@ unsigned long DmrppCommon::add_chunk(
...
         const vector<unsigned long long> &position_in_array)
 {
-    std::shared_ptr<Chunk> chunk(new Chunk( byte_order, size, offset, position_in_array));
+    shared_ptr<Chunk> chunk(new Chunk( byte_order, size, offset, position_in_array));
+
+    d_chunks.push_back(chunk);
+    return d_chunks.size();
+}
+
+unsigned long DmrppCommon::add_chunk(
```
ndp-opendap, line 297: This method needs a doxygen (or whatever it is) header.

modules/dmrpp_module/DmrppCommon.h
```diff
@@ -61,6 +61,24 @@ class SuperChunk;
 
 void join_threads(pthread_t threads[], unsigned int num_threads);
 
+/**
+ * @brief hold the value used to fill empty chunks
+ */
+union fill_value_union {
```
ndp-opendap, line 67: Sweet! It's been a long time since I've seen/used a `union`.

modules/dmrpp_module/DmrppCommon.h
```diff
@@ -191,17 +207,32 @@ class DmrppCommon {
...
     void print_compact_element(libdap::XMLWriter &xml, const std::string &name_space = "", const std::string &encoded = "");
 
     void print_dmrpp(libdap::XMLWriter &writer, bool constrained = false);
 
-    // Replaced hsize_t with size_t. This eliminates a dependency on hdf5. jhrg 9/7/18
     /// @brief Set the value of the chunk dimension sizes given a vector of HDF5 hsize_t
-    void set_chunk_dimension_sizes(const std::vector<size_t> &chunk_dims) {
-        // tried using copy(chunk_dims.begin(), chunk_dims.end(), d_chunk_dimension_sizes.begin())
-        // it didn't work, maybe because of the differing element types?
+    void set_chunk_dimension_sizes(const std::vector<unsigned long long> &chunk_dims) {
```
ndp-opendap, line 235: Nice catch! That `size_t` business is trouble.

modules/dmrpp_module/SuperChunk.cc
```diff
@@ -74,6 +76,8 @@ void process_one_chunk(shared_ptr<Chunk> chunk, DmrppArray *array, const vector<
 {
     BESDEBUG(SUPER_CHUNK_MODULE, prolog << "BEGIN" << endl );
 
+    // TODO If this is a SuperChunk, hasn't the data been read by SuperChunk::Retrieve_data()?
```
ndp-opendap, line 79: I'm don't know what the answer is but it seems like we should resolve this question now or write a ticket to do so.

modules/dmrpp_module/SuperChunk.cc
```diff
@@ -388,12 +390,33 @@ void process_chunks_unconstrained_concurrent(
 //
 // = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = =
 
-
+// TODO There are (at least) two ways to handle 'fill value chunks.' The code can group
+//  them all together as one big SuperChunk or store each FV chunk in its own SuperChunk.
+//  (Of course, there are alternatives...) Using one SuperChunk is probably faster but
+//  will require more work on the SuperChunk code. I think we should postpone that for now
+//  to focus on getting the values correct (because that problem has yet to be solved).
+//  I will add a ticket to return to this code and make that modification. jhrg 5/7/22
```
ndp-opendap, line 398: +1

modules/dmrpp_module/SuperChunk.h
```diff
@@ -55,44 +52,47 @@ class SuperChunk {
...
+    bool d_uses_fill_value{false};
+
     bool is_contiguous(std::shared_ptr<Chunk> candidate_chunk);
     void map_chunks_to_buffer();
     void read_aggregate_bytes();
+    void read_fill_value_chunk();
 
 public:
-
+    // TODO Make the sc_id an uint64 and not a string - the code uses sstream to make the value. jhrg 5/7/22
```
ndp-opendap, line 63: but... What if we end up using a readable name? Will the be useful down the road?

ndp-opendap (approved):
That's a lot of work @jgallagher59701! Most of it looks great and most of my comments are nitpicky, but I see there are a lot of TODO items, many of which are important to sort out. It seems to me that  we need write follow on tickets for these before we merge this, because once it's back in the main branch it will be harder to remember what needs to be done and more likely that we'll just blow it off to the projects potential detriment.

modules/dmrpp_module/DmrppArray.cc
jgallagher59701: Yes. It's also the default for lots of automated tools.

modules/dmrpp_module/DmrppCommon.h
jgallagher59701: Yeah.

modules/dmrpp_module/DmrppCommon.h
jgallagher59701: Right. site_t is what vector::size() returns, so we need it, but we know our arrays can be bigger. We might switch from writing `unsigned long long` to `uint64_t` now that we have C++11 - it standardizes that while unsigned long long is not _guaranteed_ to be 64-bits.

modules/dmrpp_module/SuperChunk.cc
jgallagher59701: [HYRAX-713](https://bugs.earthdata.nasa.gov/browse/HYRAX-713)

modules/dmrpp_module/SuperChunk.h
jgallagher59701: I don't understand - my comment is that we don't need a string here and doing so uses more time than not. An int will be fine and if we need a string lagter on, we do that. 

I'm not changing the code now; this can go in a general TODO ticket and we can revisit the issue.

jgallagher59701:
All of the review comments have been addressed - if this passes the tests now that the #if 0 stuff is gone, I'll merge it and watch as the build progresses. There are a number of new tickets for the various TODO items - mostly in the 'dmr++' sprint (HYRAX-722 to 726).

ndp-opendap:
Thanks James!

PR #618, opened 2022-04-25
quiet down git

jgallagher59701:
This should be a trivial PR - If this passes, I'll merge without review. The only changes here are to .gitignore files.


ndp-opendap (approved):
[Shut. Up.](https://www.youtube.com/watch?v=EbSO7G2GT1w) Git...

PR #620, opened 2022-04-26
Hyrax 418

kyang2014:


modules/fileout_netcdf/FONcArray.cc
```diff
@@ -460,7 +460,15 @@ void FONcArray::define(int ncid) {
             // TODO Make this more adaptable to the Array's data type. Find out when it's
             //  best to use shuffle, et c. jhrg 7/22/18
             if (FONcRequestHandler::use_compression) {
+
                 int shuffle = 0;
+                // For integer, if the type size is >= 2, turn on the shuffle key always.
+                // For other types, turn on shuffle the key by default.
```
jgallagher59701, line 466: Do you mean 'For other types, turn **off** shuffle the key by default.'?

Also, with this logic, shuffle will be true for the integer types listed even if use_shuffle is false. Is that what we want? Maybe we do, but there is no way for users to get the old behavior. Is use_shuffle there for data providers that want floats and doubles shuffled (I assume that char is never shuffled).

kyang2014: > Do you mean 'For other types, turn **off** shuffle the key by default.'?

Yes, turn off, changed. 




> Also, with this logic, shuffle will be true for the integer types listed even if use_shuffle is false. Is that what we want? Maybe we do,

Yes, I think that's what we agreed. For integer type when the compression is on, shuffle is always turned on.  
Do we want to turn off shuffling when users don't want to use the shuffle at all compression?
I don't have strong preference although I prefer to always turn on the shuffle for integer.  Some users may want to turn off shuffle for floating but turn on shuffle for integer. 
Another way is: we can add two keys for shuffle. One for integer and one for float. To me that's too much for users.

jgallagher59701: We decided to merge this version and keep an ear out for user's comments.

PR #621, opened 2022-04-27
Fixing a prpblem with the way the DMZ processes trusted urls

ndp-opendap:
The code in the `ngap_module` marks CMR derived URLs as trusted using the XML attribute `dmrpp:trust`, but the DMZ code fails to identify this when a chunk element contains an `href` attribute as it only/incorrectly checks for a `trust` attribute and not `dmrpp:trust`. 

This branch corrects this.

Also of note, there are no tests, either in the unit-tests or in the regression tests that test this feature.


ndp-opendap:
**Also of note:**
Setting the define `USE_DMZ_TO_MANAGE_XML` to false causes failures because there are places in the code base where the `DMZ` is used unconditionally, and when `USE_DMZ_TO_MANAGE_XML` is not set to turn the `DMZ` object is a `nullptr` which causes a segmentation fault.

If the intention is to replace `DmrppSaxParser2` with `DMZ` then the `DmrppSaxParser2` should be retired and more thorough unit tests need to be developed to capture the functionality of `DmrppSaxParser2`.

jgallagher59701:
Merge

jgallagher59701 (approved):

PR #622, opened 2022-04-27
Replace sigalrm

dh-opendap:
Ready for review


dispatch/unit-tests/RequestTimerTest.cc
```diff
@@ -0,0 +1,285 @@
...
+
+    // Called at the end of the test
+    ~RequestTimerTest()
+    {
+    }
+
+    // Called before each test
+    void setUp()
+    {
+        if(debug) cerr << endl;
```
jgallagher59701, line 75: This could go in the constructor

dispatch/unit-tests/RequestTimerTest.cc
```diff
@@ -0,0 +1,285 @@
...
+        //CPPUNIT_TEST_EXCEPTION(test_ingest_chunk_dimension_sizes_4, BESError);
+
+    CPPUNIT_TEST_SUITE_END();
+};
+
+CPPUNIT_TEST_SUITE_REGISTRATION(RequestTimerTest);
+
+} // namespace timer
+
+int main(int argc, char*argv[])
```
jgallagher59701, line 246: This could be replaced by the template function in bes/modules/read_tests_cppunit.h

modules/asciival/BESAsciiTransmit.cc
```diff
@@ -98,6 +99,10 @@ void BESAsciiTransmit::send_basic_ascii(BESResponseObject *obj, BESDataHandlerIn
         // Send data values as CSV/ASCII
         unique_ptr<DDS> ascii_dds(datadds_to_ascii_datadds(loaded_dds));  // unique_ptr<> jhrg 9/6/16
 
+        // Verify the request hasn't exceeded bes_timeout, and disable timeout if allowed.
+        RequestServiceTimer::TheTimer()->throw_if_timeout_expired(prolog + " Ready to start streaming", __FILE__, __LINE__);
```
jgallagher59701, line 103: This is the hard part where to put the test(s). I wonder (but I don't know) if it would be better to test inside get_data_values_as_ascii(...); I don't know - maybe that's overkill. Here's the bulk of that function:

    DDS::Vars_iter i = dds->var_begin();
    while (i != dds->var_end()) {
        if ((*i)->send_p()) {
            dynamic_cast<AsciiOutput &>(**i).print_ascii(strm);
            strm << "\n";
        }
        ++i;
    }

If we tested the timer at the end of the loop, then we'd be testing once some data are sent. But then the time would include the time to send the data...

I think your solution is probably what we want, but we might need to revisit this later.

jgallagher59701 (approved):
I think we should start using the templated main<>() for our unit tests. That's the only change I think you need to make. It should be a easy change - if it is not, ignore this request, leave it as is and merge.

The spot in fileout_netcdf where the timeout is cancelled looks like the correct place to me.

modules/fileout_netcdf/FONcTransform.cc
```diff
@@ -433,6 +442,8 @@ void FONcTransform::transform_dap2(ostream &strm) {
             fbt->write(_ncid);
             nc_sync(_ncid);
 
+            RequestServiceTimer::TheTimer()->throw_if_timeout_expired(prolog + " bes-timeout: preparing to stream " + fbt->name() , __FILE__, __LINE__);
```
ndp-opendap, line 445: prolog + "**_ERROR: bes-timeout expired before transmit."_**

modules/fileout_netcdf/FONcTransform.cc
```diff
@@ -420,6 +422,13 @@ void FONcTransform::transform_dap2(ostream &strm) {
         uint64_t byteCount = 0;
 
         if (is_streamable()) {
+            // Verify the request hasn't exceeded bes_timeout.
+            RequestServiceTimer::TheTimer()->throw_if_timeout_expired("Ready to start streaming", __FILE__, __LINE__);
```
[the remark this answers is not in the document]
dh-opendap: I modified all the error message so that they start with bes-timeout: <message> in an attempt to tell the observer that the real cause was that the bes-timeout expired.

modules/fileout_netcdf/FONcTransform.cc
dh-opendap: OK. will add.

CMakeLists.txt
```diff
@@ -404,7 +404,8 @@ add_executable(bes
     dispatch/BESModuleApp.h
     dispatch/BESNames.h
     dispatch/BESNotFoundError.h
-    dispatch/BESObj.h
+		dispatch/BESObj.h
+		dispatch/BESObj.cc
```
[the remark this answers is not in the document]
dh-opendap: Yup

PR #625, opened 2022-05-03
Moved setContext XML command handling to build_data_request_plan

dh-opendap:
Moved handling of setContext XML commands into build_data_request_plan().   Still followed the previous approach of identifying the response_handler and calling its execute() method, which is somewhat opaque for this usage but works.  Only changes in xml_command/BESXMLInterface and xml_command/BESSetContextCommand.

### DIFF
--- a/xmlcommand/BESXMLInterface.cc
+++ b/xmlcommand/BESXMLInterface.cc
@@ -177,7 +177,6 @@ void BESXMLInterface::build_data_request_plan()
                 if (node_name == SET_CONTEXT_STR) {
                     current_cmd->parse_request(current_node);
                 } else {
-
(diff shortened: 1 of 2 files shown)


ndp-opendap (approved):
Well that didn't turn out to be so bad. Nice!

:shipit:

jgallagher59701 (approved):
I wonder if leaving the guts of the set-context command in BESXMLSetContextCommand will leave us scratching our heads down the road. I'm not sure how I would handle the implementation differently, however.
I think this should be merged.

dh-opendap:
Updated the code so that the only changes required were in build_data_request_plan(), not changes made to the BESXMLSetContextCommand class.

PR #626, opened 2022-05-04
Dmz issues

ndp-opendap:
This PR contains three fixes:

1. The DMZ was using `std::stoi()` to convert strings that (in reality) are representations of `unsigned long long` values.  I switched these calls to `std::stoull()`
2.  The `DmrppArray::find_needed_chunks()` method was coming up empty handed when a request was for data entirely inside of a missing chunk, and this was causing a nullptr exception downstream. While this should be addressed by the `FillValue` tickets I went ahed and added a sanity check so that if for some reason the `DmrppArray::find_needed_chunks()` comes up empty handed (because `FIllValue` or some future issue shows up) a `BESInternalError` is thrown with a useful message.
3. I added comment headers to the `DmrppCommon::add_chunk()` methods because I could understand what was happening without some minimal documentation.

### DIFF
--- a/modules/dmrpp_module/DmrppArray.cc
+++ b/modules/dmrpp_module/DmrppArray.cc
@@ -1260,10 +1260,8 @@ void DmrppArray::read_chunks()
             }
         }
     }
-
     BESDEBUG(dmrpp_3, prolog << "found_needed_chunks: " << (found_needed_chunks?"true":"false") << endl);
-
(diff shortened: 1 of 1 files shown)


jgallagher59701 (approved):
If this passes, it should be merged. NB: Probably should name the changed functions 'stoi,' etc., and not 'atoi' in the commit message. Change that during Squash and Merge.

PR #627, opened 2022-05-04
Changed TempFile to take a separate dir and file template as parameters

ndp-opendap:
Changed `TempFile` to take a separate directory name and file template as parameters to its constructor.
The `TempFile` will try to create the directory and if it fails it will throw an exception. 
If the directory exists already this is not a failure.
Now the code that uses TempFile is not limited to `/tmp` or some other prexisting directory, which allows use to be more discerning in where these temp files get placed.

Also in this PR I changed the way the filet_netcdf names its temples so that the temp file name contains the dataset name.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -405,7 +405,6 @@ add_executable(bes
     dispatch/BESNames.h
     dispatch/BESNotFoundError.h
 		dispatch/BESObj.h
-		dispatch/BESObj.cc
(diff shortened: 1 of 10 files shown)


jgallagher59701 (approved):
This looks ok, but you should look at the sonarscan issues, even if only to mark them as 'reviewed.'

dh-opendap (approved):
The changes all look good to me.

PR #628, opened 2022-05-05
New init() for TempFile static member

ndp-opendap:
Instead of initializing the static class member variable `TempFile::open_files` in the .cc file like this:
```
std::map<string, int> *TempFile::open_files = new std::map<string, int>;
```
Which always produces complaints from **CLion** about the possibilities of an exception being thrown, I tried moving the actual initialization to a static `TempFile::init()` method which is call from the constructor using `std::call_once()`

I think this might be a pattern we could use in a number of places.

### DIFF
--- a/dap/GlobalMetadataStore.cc
+++ b/dap/GlobalMetadataStore.cc
@@ -1254,7 +1254,7 @@ DDS *
 GlobalMetadataStore::get_dds_object(const string &name)
 {
     TempFile dds_tmp;
-    string dds_tmp_name = dds_tmp.create(get_cache_directory(), "mds_dds_XXXXXX");
+    string dds_tmp_name = dds_tmp.create(get_cache_directory(), "mds_dds");
 
     fstream dds_fs(dds_tmp_name.c_str(), std::fstream::out);
     try {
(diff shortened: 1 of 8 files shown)


ndp-opendap:
After looking at the [contract for mkstemp()](https://man7.org/linux/man-pages/man3/mkstemp.3.html) which states:
> The last six characters of template must be "XXXXXX"

I am think I should finish this refactor by modifying the `TempFile::create()` method to do hand adding the suffix. It seems silly to be checking for that and throwing exceptions if it's missing  (or ignoring it as the previous code did with a comment about the perils of getting it wrong in the method doc) when we know it's required. If the `TempFile::create()` method handles appending the six 'X' chars to the end of the template string then it will always work as expected, and the callers can just put together the prefix string and pass value into the method. I think this will be less confusing going forward.

ndp-opendap:
Also I noticed in the `TempFile::sigpipe_handler()` code we use the  ERROR_LOG macro and that in `TempFile::~TempFile()` we do as well. However, in `TempFile::~TempFile()` we wrap that call in a try catch block to ensure no exception can be thrown  - is that something that should also be added to `TempFile::sigpipe_handler()`? As in - make sure no exceptions get thrown in the signal handler?

ndp-opendap:
Made the template construction for `mkstemp()` part of the `TempFile::create()` method (i.e. the `XXXXXX` business is now baked in to the `TempFile::create()` method) Changed the semantics of the passed string from "template" to "prefix" and adjusted all callers accordingly.

ndp-opendap:
I think this is ready for final review and merge

jgallagher59701 (approved):
Looks good.

jgallagher59701 (approved):
Merge

jgallagher59701:
Odd - both times I clicked 'Approve'

Merge it.

PR #629, opened 2022-05-16
Fill value fix 1

jgallagher59701:
This PR fixes an issue where fill value chunks were being decompressed - which makes no sense and causes a hard failure. Fixed.


ndp-opendap (approved):
Make it so...

sawer297 (approved):
looks good to me

PR #632, opened 2022-05-20
Refined template substitution to resolve incorrect template matching

ndp-opendap:
Because the `invocation` attribute typically contains the template and we don't want to match it there.


jgallagher59701 (approved):
I think this is ready. Merge!

PR #636, opened 2022-05-25
Deflate fix

jgallagher59701:
Fixed and added tests.


kyang2014 (commented):
One question about using the inflate: Does the compression level need to be preserved? The testing file deliberately uses deflate_level =6 and deflate_level=1. So the DMRPP code handles this correctly?

kyang2014 (approved):

jgallagher59701:
—
James Gallagher
***@***.***
On May 25, 2022, 14:11 -0600, kyang2014 ***@***.***>, wrote:
	@kyang2014 commented on this pull request.
One question about using the inflate: Does the compression level need to be preserved? The testing file deliberately uses deflate_level =6 and deflate_level=1. So the DMRPP code handles this correctly?

Yes, inflate does not need to know the compression level - the level determines how much work the compressor does, but the resulting stream can be read without knowing the number.

PR #637, opened 2022-05-25
Get dmrpp tests

dh-opendap:
Updated testsuite.at to include tests for all get_dmrpp options, many currently commented out until get_dmrpp and/or build_dmrpp_macros.m4 are fixed.


modules/dmrpp_module/tests_build_dmrpp/build_dmrpp_macros.m4
```diff
@@ -250,38 +258,39 @@ AS_IF([test -z "$at_verbose"], [
...
 
 AS_IF([test -n "$baselines" -a x$baselines = xyes],
 [
     AS_IF([test -z "$at_verbose"], [echo "# get_dmrpp_baselines: Calling get_dmrpp application."])
     AT_CHECK([${TEST_CMD}], [], [stdout], [stderr])
-    NORMAILZE_EXEC_NAME([stdout])
-    REMOVE_PATH_COMPONENTS([stdout])
-    REMOVE_VERSIONS([stdout])
-    REMOVE_BUILD_DMRPP_INVOCATION_ATTR([stdout])
+    NORMAILZE_EXEC_NAME([${output_file}])
```
jgallagher59701, line 269: NORMALIZE_EXEC_NAME is misspelt (and in the original, too). I'm sure that's my handiwork, but it should be fixed.

jgallagher59701 (approved):
If you can, fix the spelling of that macro and then merge. Else, merge and we'll tackle it as a separate thing.

Thanks

ndp-opendap (approved):
Looks like you need to merge master into this branch resolve the conflicts, and then I think this is good to go.

PR #639, opened 2022-05-27
Added testsuite switch for S3 tests. Turned off always on verbose mode

ndp-opendap:


ndp-opendap:
This still needs a modification to .travis.yml to invoke the tests with the s3 tests turned on. I could use some advice about how to go about that.

modules/dmrpp_module/tests_build_dmrpp/Makefile.am
```diff
@@ -20,7 +20,8 @@ AUTOTEST = $(AUTOM4TE) --language=autotest
 TESTSUITE_DMRPP = $(srcdir)/testsuite
 
 check-local: atconfig atlocal  $(TESTSUITE_DMRPP) $(builddir)/test_bes.conf
-	$(SHELL) '$(TESTSUITE_DMRPP)' $(TESTSUITEFLAGS) && echo "bindir: $(bindir)"
+	if ./chkForAWS.sh; then export tsflags="$(TESTSUITEFLAGS) --s3tests=yes"; else export tsflags="$(TESTSUITEFLAGS)"; fi && \
```
jgallagher59701, line 23: What happens when the aws command is not installed (or found)?

I think you can decouple these so the '&&' is not required. 

You could simplify the line to 

if ./chkForAWS.sh; then export tsflags="--s3tests=yes"; fi

and then 

$(SHELL) '$(TESTSUITE_DMRPP)' $(TESTSUITEFLAGS) $${tsflags}

modules/dmrpp_module/tests_build_dmrpp/chkForAWS.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+#
+# Checks to see if the AWS CLI is present and configured with credentials.
+#
+aws configure list | grep "access_key" | grep -v "<not set>" > /dev/null
+ret=$?
+if test $ret != 0 ; then
+    echo "ERROR - The AWS_ACCESS_KEY_ID is not set."
```
jgallagher59701, line 8: Do we really want this output, esp the 'ERROR' part?

jgallagher59701:
Should we talk about this - I'm never 100% sure about the shell + make stuff, but I'm working on moving our Makefile.am files, especially the tests, to be more uniform. Thus, the part about not linking the two lines unless we have to.

PR #641, opened 2022-05-27
Moved the call to AT_XFAIL_IF to the top of the macros

jgallagher59701:
This makes it more clear what is really happening - that
AT_XFAIL_IF is run before any other part of the macro and
which parameter is used. Also, this fixed one bug.


ndp-opendap (approved):
xfail for the wins

:shipit:

PR #643, opened 2022-06-06
Functions refactor

jgallagher59701:


jgallagher59701:
I've replaced this with the new-gdal branch and merged the best parts of it into that branch.

PR #645, opened 2022-06-10
Hyrax 639

kyang2014:
Make the fileout coverage json support the CF geographic projection only.


jgallagher59701 (approved):
Looks good. 
Merge!
The plan to remove the #if 0 .. #ndif debugging in the next few commits is good, too.
Thanks!

PR #646, opened 2022-06-10
New gdal

jgallagher59701:
new gdal handler consolidates the gdal code in one place I removed the gdal-based functions from the functions module). This should eliminate run-time issues with GDAL on RHEL 8.


kneumiller (commented):
Looks good. Does the new gdal_module introduce any significant changes compared to the old fileout_gdal and gdal_handler?

jgallagher59701:
> Looks good. Does the new gdal_module introduce any significant changes compared to the old fileout_gdal and gdal_handler?

There is no new functionality. I realized that I could combine these and thus link only one module to liberal.a instead of two, and thus address the issue we were seeing at run time during server exit. NB: I also removed some software from the functions module since I had to reduce the use of GDAL to only one place. We _could_ move those functions to this new 'gdal_module.' I didn't do that because it seemed too much for a single PR.

PR #647, opened 2022-06-15
H4 clean csmell

kyang2014:
Just check with sonar cloud.


kyang2014:
This is to clean up code smells for the HDF4 handler. Mostly use auto, range-for, override, const to replace the old ways. Also clean up/remove some old code.

kyang2014:
valgrind tests are good. merge it.

PR #648, opened 2022-06-17
Refactored std::vector memory access patterns in NCArray

ndp-opendap:
Replaced usage of the `std::vector` access pattern `&v[0]` with the c++11 method `std::vector::.data()` in the hope that this change will address rh8 SIGABRT issues that seemed tied to the old pattern.

The c++11 access method `std::vector::data()` is discussed here:

- https://cplusplus.com/reference/vector/vector/data/

This PR is an experiment to see if James' assertion that the implementation of `std::vector` may have changed in RH8 could be at the root of the zero-length array SIGABRT problems.  When I saw that the class had this "new" c++11 method I figured it would be worth a shot to see if this would address the issue.

If this does produce a fix I think we will want to write a ticket to make a similar refactor throughout both the libdap code and BES code. Once bitten, etc.

A quick search of *.cc and *.h files in **`libdap4`** and **`bes`** for things matching **`&[a-zA-Z0-9]+\[0\]`** shows ~1100 occurrences of the access pattern that we may need to address.

### DIFF
--- a/modules/netcdf_handler/NCArray.cc
+++ b/modules/netcdf_handler/NCArray.cc
@@ -1,4 +1,3 @@
-
 // -*- mode: c++; c-basic-offset:4 -*-
 
 // This file is part of nc_handler, a data handler for the OPeNDAP data
--- a/modules/netcdf_handler/NCArray.cc
+++ b/modules/netcdf_handler/NCArray.cc
@@ -175,9 +175,9 @@ void NCArray::do_cardinal_array_read(int ncid, int varid, nc_type datatype,
             if (!has_values) {
                 values.resize(nels * size);
                 if (has_stride)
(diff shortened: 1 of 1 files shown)


ndp-opendap:
I was able to build a the `opendap/hyrax-rh8:1.16.5-274` docker image using RPMs from this branch and the docker image passed all of the regression tests.

I think we need to propagate these changes throughout our code base!

jgallagher59701:
Take a look at the code smells and one low security issue. My take on it is to make the edits and merge this - I'll review the code 'formally' so that's more obvious.

As for the 1,000 +/- other places, I suggest that we make a ticket and take that on for the next PI. My _guess_ is that CLion's regex replace, or emacs' or sed are the tools to use for that. Based on what I know, it's a straight replacement.

jgallagher59701 (approved):
These changes are good. Merge.

PR #649, opened 2022-06-20
Changed from &vector[0] to vector.data() for direct access

jgallagher59701:


ndp-opendap (approved):
Gotta love the almighty regex.

:shipit:

PR #657, opened 2022-07-12
Hyrax 745

kyang2014:
For DMR++, add the handling of the scalar variable reading when the contiguous storage and the data value is the fill value. The testsuite is also added to cover this case.


jgallagher59701 (approved):
Merge

PR #658, opened 2022-07-13
Generic timeout errmsg

dh-opendap:
Generic error message added to RequestServiceTimer::throw_if_timeout_expired(), that's the only change.


ndp-opendap (commented):
I made some changes to the error message, and I added a unit test for RequestServiceTimer::throw_if_timeout_expired(). If the unit test is run in debug mode it will show the complete error message.

ndp-opendap (approved):
Let's :shipit:

PR #659, opened 2022-07-14
Hyrax 772

ndp-opendap:
Deal with "It's too big" errors in fileout_netcdf by improving detection and error messaging."


dh-opendap (approved):
Looks good.   Always suspicious of type changes from size_t to unsigned long long (in FoncRequestHandler.h) but assume that type change is warranted.

ndp-opendap:
> Looks good. Always suspicious of type changes from size_t to unsigned long long (in FoncRequestHandler.h) but assume that type change is warranted.

Depending on the system, size_t may not be big enough, so what I did is added another overload instance to `FoncRequestHandler::read_key_value` for the specific `unsigned long long` type:  `FoncRequestHandler::read_key_value(const string &key_name, unsigned long long &key, const unsigned long long default_value)` which uses `std::stoull()` to convert the key-value to an unsigned long long

PR #662, opened 2022-08-02
Modified BESDapTransmit so that it uses the libdap::Error file/line info

jgallagher59701:
This is an improvement over using __FILE__ and __LINE__ for the location
where the libdap::Error object was rethrown. This makes tracking down
errors and setting debugger breakpoints faster.


ndp-opendap (approved):
:shipit:

PR #663, opened 2022-08-07
Hyrax 771

kyang2014:
Check if the tests get passed.


modules/dmrpp_module/Chunk.cc
```diff
@@ -209,17 +209,20 @@ size_t chunk_write_data(void *buffer, size_t size, size_t nmemb, void *data) {
  *
  * @note Stolen from the HDF5 library and hacked to fit.
  *
- * @param dest Write the 'inflated' data here
+ * @param destp The pointer to the pointer of the 'inflated' data
```
jgallagher59701, line 212: You could also describe this as a 'value-result' parameter - that is, the parameter is passed in as a value and the code modifies that in the course of execution and then uses the parameter to return the result.

modules/dmrpp_module/Chunk.cc
```diff
@@ -612,23 +617,127 @@ void Chunk::filter_chunk(const string &filters, unsigned long long chunk_size, u
...
+
         string filter = *i;
 
         if (filter == "deflate") {
-            char *dest = new char[chunk_size];
-            try {
-                inflate(dest, chunk_size, get_rbuf(), get_rbuf_size());
-                // This replaces (and deletes) the original read_buffer with dest.
+
+            // Here we find that the deflate filter is applied twice.
```
jgallagher59701, line 664: Is this because the premise of the DMR++ document that all chunks for a given variable/dataset have the same filters is wrong? If so, we should probably fix that. Let me know if that's the case and we can make tickets for it, et cetera.

jgallagher59701 (approved):
This can be merged. I made one comment and I have one question about the DMR++ design and if it should be changed.

PR #667, opened 2022-08-17
Length refactor

jgallagher59701:
This branch takes all the calls to string::length in the BES and changes them so they use string::size.
This will make using libdap::BaseType::length easier to spot and edit so it uses ...length_ll().


jgallagher59701:
Woo Hoo. It works!

ndp-opendap (approved):
Wow. The scope is so big for this change. CLion must have been very busy...

:shipit:

PR #668, opened 2022-08-17
Dmrpp string arrays

ndp-opendap:
This PR includes the code in support of [HYRAX-653](https://bugs.earthdata.nasa.gov/browse/HYRAX-653) and includes:

* Code to detect fixed length string arrays and add the relevant information to the dmr++ by adding the element:
`<dmrpp:FixedLengthStringArray string_length="###" pad="null | space"/>`
* Support  for arrays of fixed length strings stored with either the Contiguous or Chunked layouts.
* Support for arrays  of fixed length strings stored in the Compact layout is not yet supported. (But will be soon)
* Arrays of variable ;length strings are not yet supported.


modules/dmrpp_module/DmrppArray.cc
```diff
@@ -1803,6 +1803,54 @@ void DmrppArray::print_dap4(XMLWriter &xml, bool constrained /*false*/)
...
+
+            if (xmlTextWriterEndElement(xml.get_writer()) < 0)
+                throw InternalErr(__FILE__, __LINE__, "Could not end " + type_name() + " element");
+        }
+        else if(is_vlsa()){
+            string element_name("dmrpp:VariableLengthStringArray");
+            if (xmlTextWriterStartElement(xml.get_writer(), (const xmlChar *) element_name.c_str()) < 0)
+                throw InternalErr(__FILE__, __LINE__, "Could not write " + element_name + " element");
+
+            if (xmlTextWriterWriteString(xml.get_writer(), (const xmlChar *) "offset[0]:size[0],...,offset[n]:size[n]") < 0)
```
jgallagher59701, line 1841: Should this be "...offset[n-1]:size[n-1]" for an N-element String Array?
But, I don't understand this - I'm sure it makes sense, but I was expecting a loop here.

modules/dmrpp_module/DmrppArray.cc
```diff
@@ -1803,6 +1803,54 @@ void DmrppArray::print_dap4(XMLWriter &xml, bool constrained /*false*/)
...
+                throw InternalErr(__FILE__, __LINE__, "Could not write text into element");
+
+            if (xmlTextWriterEndElement(xml.get_writer()) < 0)
+                throw InternalErr(__FILE__, __LINE__, "Could not end " + type_name() + " element");
+
+        }
+        else {
+            //throw InternalErr(__FILE__, __LINE__, "ERROR: Internal State Failure. The array does not "
```
jgallagher59701, line 1849: Is this a comment to remind you of more work to be done?

modules/dmrpp_module/DmrppArray.h
```diff
@@ -119,16 +144,41 @@ class DmrppArray : public libdap::Array, public dmrpp::DmrppCommon {
             const vector<unsigned long long> &constrained_array_shape);
 
 public:
-    DmrppArray(const std::string &n, libdap::BaseType *v) : libdap::Array(n, v, true /*is dap4*/), DmrppCommon() { }
+    DmrppArray(const std::string &n, libdap::BaseType *v) :
+            libdap::Array(n, v, true /*is dap4*/),
+            DmrppCommon(),
+            is_variable_length_string_array(false),
```
jgallagher59701, line 150: It's probably better to initialize these where they are defined in the class and not here in the constructors since that will eliminate duplication and provide the proverbial single source of truth. I bet these are some of the 21 'code smells' that sonar is reporting.

modules/dmrpp_module/DmrppArray.h
```diff
@@ -70,6 +79,22 @@ class DmrppArray : public libdap::Array, public dmrpp::DmrppCommon {
...
     //void _duplicate(const DmrppArray &ts);
 
+    // In the dmr++ XML:
+    //     <dmrpp:vStringArray>
+    //         <v>0:1084,1025:653,65523:8746,9750:100,84660:122, ... ,98466:12</v>
+    //     </dmrpp:vStringArray
+    std::string d_vlen_string_addrs;
+
+
+    bool is_variable_length_string_array;
```
jgallagher59701, line 89: These bools can be initialized here - see my comment below.

jgallagher59701 (approved):
I made some suggestions, but this is a work in progress, so make them now or wait later. Plus I know some of the things I said are probably about place holders that will be changed by the time 654 is done.

modules/dmrpp_module/DmrppArray.cc
ndp-opendap: It's pseudo data. It might as well have been a string like "LOOK HERE FOR THE STUFF YOU NEED"

modules/dmrpp_module/DmrppArray.cc
ndp-opendap: Throwing an exception there broke a lot of tests. Don't know why that would be the case...

PR #674, opened 2022-09-08
Hyrax 811

kyang2014:
The CF option to handle the case when a non-coordinate netCDF variable shares the same name as a dimension name in a netCDF-4 file.


jgallagher59701:
Thanks, This is ready to be merged.

jgallagher59701 (approved):
Ready for merge. 

The SonarScan analysis shows that some of the code is 'too' complex. Obviously, that's a judgement call and it seems to be a pretty arbitrary rule in the system. I'm pretty sure that they are using some variation of cyclomatic complexity (aka McCabe). Here's a link to some information on that: https://en.wikipedia.org/wiki/Cyclomatic_complexity. I'm on a plane and the internet is really slow, so it's hard to say why because looking at the code in detail will have to wait until later. Let's merge this and refactor the code, if we indeed decide to do that, on a separate branch later.

But I generally find it's better to do any refactoring after the PR is merged,

PR #675, opened 2022-09-08
Build dmrpp refactor

ndp-opendap:
This is a simple refactor PR:

- Simplified the most complex function, breaking it into smaller blocks of code
- Improved the naming of some functions to reflect their actual activity.
- Added new helper functions for tasks that happen frequently.
- Cleaned up the instances of the hdf5 plist that were not being closed.

### DIFF
--- a/modules/dmrpp_module/build_dmrpp_util.cc
+++ b/modules/dmrpp_module/build_dmrpp_util.cc
@@ -142,13 +142,26 @@ string h5_filter_name(int type) {
 }
 
 /**
+ * Safely creates an hdf5 plist for the dataset.
+ * @param dataset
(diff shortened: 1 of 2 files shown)


jgallagher59701 (approved):
Looks like it's ready to merge!

I'm on a plane, and like I told Kent, the network is really slow so I can't do much. SonarScan is complaining about complexity in a few places, but it sounds like you've improved things. When SS say that thing about a method being at '31' and that's over the threshold of '25', I think it's talking about Cyclomatic complexity. Here's a link FWIW: https://en.wikipedia.org/wiki/Cyclomatic_complexity.

Anyway, thanks and merge!

PR #676, opened 2022-09-09
More refactoring, no work on compact yet

ndp-opendap:
I think this completes the refactoring of build_dmrpp, at least for now.

I want to merge this and then begin work on implementing the arrays of fixed length strings with compact storage use case.

A new function, `proto_flsa_compact()` is the beginning of this implementation, but it is not being utilized at this point.


jgallagher59701 (approved):
Merge. Thanks!

PR #677, opened 2022-09-11
HYRAX-810, make build_dmrpp correct generate chunk info for the case …

kyang2014:
…when anetCDF-4 variable shares its name with a dimension.


modules/dmrpp_module/build_dmrpp_util.cc
```diff
@@ -1034,6 +1035,44 @@ string get_type_decl(BaseType *btp){
...
+
+    if (dimname_list.empty() == false) {
+        // Then find the nc4_non_coord candidate variables,
+        for (auto btp = group->var_begin(), ve = group->var_end(); btp != ve; ++btp) {
+            if (dimname_list.find((*btp)->name())!=dimname_list.end())   
+                nc4_non_coord_candidate.insert((*btp)->name());
+        }
+    }
+
+#if 0
```
jgallagher59701, line 1067: Let's remove this now.

modules/dmrpp_module/build_dmrpp_util.cc
```diff
@@ -1080,12 +1119,29 @@ void get_chunks_for_all_variables(hid_t file, D4Group *group) {
             // that variable was synthesized (likely for CF compliance)
             H5Eset_auto2(H5E_DEFAULT, nullptr, nullptr);
             string FQN = (*btp)->FQN();
+            string name = (*btp)->name();
+            if (nc4_non_coord_candidate.empty() == false) {
```
jgallagher59701, line 1123: You don't need to test if the set is empty(). If it is, find() will return end().

modules/dmrpp_module/build_dmrpp_util.cc
```diff
@@ -1080,12 +1119,29 @@ void get_chunks_for_all_variables(hid_t file, D4Group *group) {
             // that variable was synthesized (likely for CF compliance)
             H5Eset_auto2(H5E_DEFAULT, nullptr, nullptr);
             string FQN = (*btp)->FQN();
+            string name = (*btp)->name();
```
jgallagher59701, line 1122: Is this (name) and FQN above used inside this if? I can't find it, but in this code-remove mode, it's hard to tell. I see FQN used below, but it would be better to move the symbol down there or just skip it (since it only is used once - I think).

jgallagher59701 (changes requested):
Take a look at these suggestions and make them if they make sense. Merge this after with or without the suggestions depending.

Thanks!

modules/dmrpp_module/build_dmrpp_util.cc
kyang2014: Done

modules/dmrpp_module/build_dmrpp_util.cc
kyang2014: Remove the empty

modules/dmrpp_module/build_dmrpp_util.cc
kyang2014: Good catch. This is for debugging. Thanks,

kyang2014: Also add a test for this feature.

kyang2014:
All the suggestions are addressed and a test is added.

PR #678, opened 2022-09-12
Dap4 grid functions

dh-opendap:
Changes are relegated to modules/functions; notably the grid and geogrid functions and their helper functions.


modules/functions/GeoGridFunction.cc
```diff
@@ -212,24 +218,211 @@ bool GeoGridFunction::canOperateOn(DDS &dds)
...
-    	// See if any one of them looks like suitable GeoGrid
-    	vector<Grid *>::iterator git;
-    	for(git=grids.begin(); !usable && git!=grids.end() ; git++){
-    		Grid *grid = *git;
-    		usable = is_geo_grid(grid);
-    	}
+        // See if any one of them looks like suitable GeoGrid
+        vector<Grid *>::iterator git;
+        for(git=grids.begin(); !usable && git!=grids.end() ; git++){
+            Grid *grid = *git;
```
jgallagher59701, line 232: This is a place - the '.begin()' and '.end()' methods are a hint - that the new range-based for-loop can be used. This can be rewritten:

for (auto grid: grids) {
    if (is_geo_grid(grid) {
         return true;
    }
}

No need to make the change this time, but the pattern can save time,

jgallagher59701 (approved):
Merge!

Thanks,

PR #686, opened 2022-10-05
Fonc string arrays

jgallagher59701:
This branch contains an optimization for the fileout_netcdf handler when it is working with an array of strings. For a test dataset with a large string array, processing time dropped significantly and the time spent in FONcArray::wrote() dropped from 70% to 7.6%. The optimization build the data in memory and make a single call to nc_put_vara_text() instead of one call for each string.


kyang2014 (approved):
Looks fine to me. One suggestion is:  don't forget remove the 
`#define STRING_ARRAY_OPT 1`
in the future.

kyang2014 (approved):
See my previous comment.

PR #688, opened 2022-10-13
HYRAX-842, adding code and tests to change the special characters of …

kyang2014:
…HDF-EOS5 object names.


jgallagher59701 (approved):
Ready to merge - you might want to try the range-based for loop.

PR #689, opened 2022-10-16
HYRAX-846, fix the attribute _fillvalue datatype mismatch issue for t…

kyang2014:
…he netCDF classic model. signed 8-bit integer is supported, so it should not be changed.


jgallagher59701 (approved):
Thanks!

PR #692, opened 2022-10-26
Initial commit of this branch - moved credential mgmt to bes/http

jgallagher59701:
I have moved the AccessCredentials, CredentialManager, NGAP Credentials and 
awsv4 signing code from the DMR++ module into the http convenience library.
This provides a way to use those classes in other parts of the server, like the 
nascent s3_reader module being developed for the automated S3-based tests.


http/CredentialsManager.cc
```diff
@@ -168,23 +158,24 @@ CredentialsManager::add(const std::string &key, AccessCredentials *ac){
...
-    AccessCredentials *best_match = NULL;
-    std::string best_key("");
+    AccessCredentials *best_match = nullptr;
+    std::string best_key;
 
-    if(url->protocol() == HTTP_PROTOCOL || url->protocol() == HTTPS_PROTOCOL) {
-        for (std::map<std::string, AccessCredentials *>::iterator it = creds.begin(); it != creds.end(); ++it) {
-            std::string key = it->first;
+    if (url->protocol() == HTTP_PROTOCOL || url->protocol() == HTTPS_PROTOCOL) {
+        //for (std::map<std::string, AccessCredentials *>::iterator it = creds.begin(); it != creds.end(); ++it) {
```
ndp-opendap, line 170: wtf -just delete the old way

http/NgapS3Credentials.cc
```diff
@@ -43,19 +43,21 @@ using std::endl;
...
+const char *NgapS3Credentials::BES_CONF_REFRESH_KEY = "NGAP.S3.refresh.margin";
+const char *NgapS3Credentials::BES_CONF_URL_BASE = "NGAP.s3.url.base";
 
+// TODO Move to the header. jhrg 10/27/22
 bool NgapS3Credentials::is_s3_cred() { return true; }
 
 string NgapS3Credentials::get(const std::string &key) {
     if (needs_refresh()) {
-        this->get_temporary_credentials();
+        this->get_temporary_credentials();  // TODO Why use 'this->'? jhrg 10/27/22
```
ndp-opendap, line 60: BTFOM. I don't understand why you didn't change it along with all the other stuff.

http/NgapS3Credentials.h
```diff
@@ -24,30 +24,32 @@
 #ifndef HYRAX_GIT_S3CREDENTIALS_H
 #define HYRAX_GIT_S3CREDENTIALS_H
 
-#include "AccessCredentials.h"
 #include <string>
 
+#include "AccessCredentials.h"
```
ndp-opendap, line 29: This reordering of includes makes no sense to me. I thought we always included the language stuff before our own includes.

http/NgapS3Credentials.h
```diff
@@ -24,30 +24,32 @@
...
 class NgapS3Credentials : public AccessCredentials {
 public:
     // These are the string keys used to express the normative key names
     // for the credentials components.
-    static const std::string AWS_SESSION_TOKEN;
-    static const std::string AWS_TOKEN_EXPIRATION;
-    static const std::string BES_CONF_S3_ENDPOINT_KEY;
-    static const std::string BES_CONF_REFRESH_KEY;
-    static const std::string BES_CONF_URL_BASE;
+    static const char *AWS_SESSION_TOKEN;
```
ndp-opendap, line 37: All of these key names should end in `_KEY` 
And we really need a better way to handle this - maybe char* is the right way but that seems like a step backwards to me.

http/NgapS3Credentials.h
```diff
@@ -24,30 +24,32 @@
...
+    static const char *AWS_TOKEN_EXPIRATION;
```
ndp-opendap, line 38: ibid

http/NgapS3Credentials.h
```diff
@@ -24,30 +24,32 @@
...
+    static const char *BES_CONF_S3_ENDPOINT_KEY;
+    static const char *BES_CONF_REFRESH_KEY;
+    static const char *BES_CONF_URL_BASE;
```
ndp-opendap, line 41: ibid

http/unit-tests/CredentialsManagerTest.cc
```diff
@@ -125,17 +120,14 @@ class CredentialsManagerTest: public CppUnit::TestFixture {
             TheBESKeys::TheKeys()->set_key(CATALOG_MANAGER_CREDENTIALS, cm_config);
             CredentialsManager::theCM()->load_credentials();
         }
-        catch (BESError &e) {
+        catch (const BESError &e) {
```
ndp-opendap, line 123: Probably there's a ticket here if `catch(const SomeError &e`) is the correct pattern. I doubt this happens in the BES's top level error handler or elsewhere in our code base.

http/unit-tests/bes_ngap_s3_creds.conf
```diff
@@ -0,0 +1,13 @@
+#
```
ndp-opendap, line 1: Looks like this is a generated file and should not be checked into git.

modules/dmrpp_module/unit-tests/Makefile.am
```diff
@@ -81,35 +72,17 @@ bes.conf: $(srcdir)/$(BES_CONF_IN) $(top_srcdir)/configure.ac
...
-             DmrppCommonTest \
-             CredentialsManagerTest \
-             awsv4_test \
-             CurlHandlePoolTest \
-             DMZTest \
-             build_dmrpp_util_test \
-             DmrppChunkOdometerTest
-
-#              DmrppMetadataStoreTest
+UNIT_TESTS = DmrppArrayTest SuperChunkTest ChunkTest DmrppParserTest DmrppCommonTest CurlHandlePoolTest \
```
ndp-opendap, line 82: I prefer the other formatting.

modules/dmrpp_module/unit-tests/awsv4_old/get-unreserved/get-unreserved.authz
```diff
@@ -0,0 +1 @@
+AWS4-HMAC-SHA256 Credential=AKIDEXAMPLE/20150830/us-east-1/service/aws4_request, SignedHeaders=host;x-amz-date, Signature=07ef7494c76fa4850883e2b006601f940f8a34d404d0cfa977f52a65bbf5f24f
```
ndp-opendap, line 1: Why is this in `awsv4_old`? We have been using `retried` and `unused`  as the directory names for things we move aside. I think this will be super confusing a few months from now.

ndp-opendap (changes requested):
There seems to be a bunch of whitespace only changes from reformatting which increases the scope and the difficulty of  finding the things that actually matter. I think, in general, if we are going to reformat a bunch code that we should do that in a separate PR - it makes doing this kind of review mind numbingly tedious.

I think it would have been a lot more manageable do this in several much smaller and more focused PRs:
- Move the classes with only the minimum changes (like namespaces etc) 
- White space changes
- Refactor/rename variable names, string expression etc.

The order might be changed, ymmv.

This kind review is very hard for me - with the scope as large as it is I feel that I am unlikely to notice an **actual** problem.

http/CredentialsManager.cc
```diff
@@ -43,18 +43,18 @@
...
 
 using namespace std;
 
 #define prolog std::string("CredentialsManager::").append(__func__).append("() - ")
 
+namespace http {
+
 // Class vocabulary
-const char *CredentialsManager::ENV_ID_KEY     = "CMAC_ID";
+const char *CredentialsManager::ENV_ID_KEY = "CMAC_ID";
```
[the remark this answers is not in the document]
jgallagher59701: We can look this up.

http/NgapS3Credentials.h
jgallagher59701: In this case, there is a #define in NgapS3Credentials.cc named AWS_SESSION_TOKEN_KEY and it has a different string value. I'm going to leave this one as it is.

http/NgapS3Credentials.h
jgallagher59701: These I changed.

http/unit-tests/CredentialsManagerTest.cc
jgallagher59701: It is the correct pattern - or a 'better' pattern.

modules/dmrpp_module/unit-tests/awsv4_old/get-unreserved/get-unreserved.authz
jgallagher59701: done

jgallagher59701:
I fixed a number of 'code smells' esp in the awsv4.cc/h files.

jgallagher59701:
So, I'll run those through before merging.

PR #693, opened 2022-10-31
Hyrax 858

kyang2014:


modules/hdf5_handler/h5cfdaputil.cc
```diff
@@ -328,35 +328,43 @@ HDF5CFDAPUtil:: print_attr(H5DataType type, int loc, void *vals)
...
-                rep<<".";
+            if (isnan(attr_val)) 
+                rep<<"NaN";
+            else {
+                bool is_a_fin = isfinite(attr_val);
+                gp.fp = (float *) vals;
+                rep << showpoint;
+                rep << setprecision(10);
+                rep << *(gp.fp+loc);
+                string tmp_rep_str = rep.str();
```
jgallagher59701, line 339: I'm not sure about the rep << showpoint, ..., because I need more context, but the repeated calls to find() can be collapsed into 'rep.str().find_first_of(".eE") == string::npos' and I think the 'isfinite(attr_val)' can be moved dow into the if, so that it can be rewritten as:

if ((rep.str().find_first_of(".eE") == string::npos) && isfinite(attr_val)) {
    rep << ".";
}

It might not make that big a difference, but it might be something that would ease future work if you need to do this in the future.

modules/hdf5_handler/h5get.cc
```diff
@@ -981,38 +981,58 @@ string print_attr(hid_t type, int loc, void *sm_buf) {
...
                 
                 float attr_val = *(float*)sm_buf;
-                bool is_a_fin = isfinite(attr_val);
-                // Represent the float number.
-                // Some space may be wasted. But it is okay.
-                gp.tfp = (float *) sm_buf;
-                int ll = snprintf(gps, 30, "%.10g", *(gp.tfp + loc));
+                // Note: this comparsion is the same as isnan.
+                // However, on CentOS 7, isnan() is declared in two headers and causes conflicts.
+                if (attr_val!=attr_val) {
```
jgallagher59701, line 986: Interesting!

modules/hdf5_handler/h5get.cc
```diff
@@ -981,38 +981,58 @@ string print_attr(hid_t type, int loc, void *sm_buf) {
...
+                    rep.resize(3);
+                    rep[0]='N';
```
jgallagher59701, line 988: You could use something like 'strncpy(rep.data(), "NaN", sizeof(NaN));' but it's probably 6 to one and half a dozen to the other. But... are you going to expect a null character at the end? The snprintf() function does append nulls.

jgallagher59701 (changes requested):
Can you double check about the terminating null for the two NaN strings? The other comments are just that - try those changes if you want. Once you look at the NaNs and see that it's OK for the strings to not have terminating nulls, it's fine to merge.

modules/hdf5_handler/h5cfdaputil.cc
kyang2014: All these are originally codes. The showpoint and the setprecision are necessary but I cannot remember why. I will trim the code with your suggestions. Thanks,

modules/hdf5_handler/h5get.cc
kyang2014: This again is the original code. I don't want to spend time changing the original code with this ticket. The null character doesn't show up in the final DMR and fileout netCDF output. So let's keep this code.

kyang2014:
>

kyang2014:
The final DMR that contains NaN doesn't have NULL.  For example:
https://github.com/OPENDAP/bes/blob/hyrax-858/modules/hdf5_handler/bes-testsuite/h5.nasa.default.dap4/S-MODE_PFC_OC2108A_adcp_os75nb.h5.dmr.bescmd.baseline
`        <Attribute name="_FillValue" type="Float64">
            <Value>NaN</Value>
        </Attribute>`

PR #698, opened 2022-11-07
Reformatting only

jgallagher59701:
Reformatting only


jgallagher59701:
Literally option + command + l (lower case L). Nothing elese.

PR #701, opened 2022-11-16
Remote Resources bug 2

jgallagher59701:
Trial #2


jgallagher59701:
Comment about the code review: The main changes are in RemoteResource.cc - all the other changes derive from those and most of those are code removal. There is one #if 0... #endif block I removed that was already wrapped in the #if 0... #endif preprocessor directives. So that is not really a change from this PR. I maybe should have left that in...

Also, there are some changes to how retrieveResource works - those are the only major _changes_. The majority of the other edits are me removing stuff that had to do with the headers.

Once we approve this, the code should be refactored and we should make tickets for BESFileLockingCache to include support for cache entry updates and for an entry to have two or more parts (and have those treated atomically).

PR #704, opened 2022-11-17
Dap2 bad request error

dh-opendap:
Changes to HDF5_Handler to throw BESSyntaxUserError when variables are determined to be incompatible with requested encoding response.  That is, when CFEnable=false and a DAP2 response is requested, and the DDS contains incompatible types which for its present incarnation this handler only identifies int64 types as incompatible


modules/hdf5_handler/h5get.cc
```diff
@@ -2546,3 +2552,14 @@ string handle_string_special_characters_in_path(const string &instr) {
     return outstr;
 }
 
+string invalid_type_error_msg(
+        string var_type
+){
+    stringstream msg;
+
+    msg << "Your request was for a (DAP2 data model response). ";
```
jgallagher59701, line 2560: Let's change this -> "Your request was for a (DAP2 data model response). "
To  "Your request was for a response that uses the DAP2 data mode. "

jgallagher59701 (approved):
Make that change and merge. Thanks.

modules/hdf5_handler/h5get.cc
dh-opendap: Good suggestion.  Made it '... response that uses the DAP2 data model.'

jgallagher59701:
I'm going to merge this.

PR #705, opened 2022-11-18
Hyrax 730

sawer297:
added :
- check for dap4 projected variables
- error message if variables are found
- unit-tests for d4_tools fcts
- regression tests for over all functionality


modules/fileout_netcdf/d4_tools.cc
```diff
@@ -0,0 +1,193 @@
...
+            }
+        }
+        else if (var->send_p() && var->type() == libdap::dods_group_c) {
+            try {
+                D4Group *d4g = dynamic_cast<D4Group *>(var);
+                if (is_dap4_projected(d4g, projected_dap4_variable_inventory)) {
+                    has_dap4_var = true;
+                }
+            }
+            catch (exception e){
```
jgallagher59701, line 172: Change 'catch(exception e)' to 'catch(const exception &e)'

modules/fileout_netcdf/d4_tools.cc
```diff
@@ -0,0 +1,193 @@
...
+
+    //check if group has any subgroups and check them for dap4 projected variables
+    for (auto gb = group->grp_begin(), ge = group->grp_end(); gb != ge; ++gb){
+        try {
+            D4Group *d4g = *gb;
+            if (is_dap4_projected(d4g, projected_dap4_variable_inventory)) {
+                has_dap4_var = true;
+            }
+        }
+        catch (exception e){
```
jgallagher59701, line 186: Same as above.

jgallagher59701 (changes requested):
Please make these changes plus look a the 'code smells' and see if any are easy to address. Be wary since SonarScan can flag things that are not that big a deal and/or suggest things that require C++-14 which we cannot yet use.

modules/fileout_netcdf/d4_tools.cc
jgallagher59701: One of the flagged 'code smells' is to change the '#define d4_tools_h_ 1' line to a constexpr. Ignore that suggestion. The other suggestions are actually pretty good.

PR #707, opened 2022-11-23
H4 clean csmell

kyang2014:
Just an experiment to see how the removal of code smells goes.


kyang2014:
Not ready to merge, just check if I clean up code smells I want to clean up.

modules/hdf4_handler/HDF4Module.cc
```diff
@@ -42,12 +41,13 @@
 #include <BESDebug.h>
 
 using namespace std;
+const string HDF4_CATALOG="catalog";
```
jgallagher59701, line 44: In this case I'm not so sure the change is an improvement. We have had mixed results with 'string' objects in the global namespace. This should be OK, but be wary about advice to do this with class members. SonarScan does not seem to like the preprocessor. I don't agree with it on that point. However, if this works and does not cause an issue during program exit, it's fine.

modules/hdf4_handler/hc2dap.cc
```diff
@@ -359,7 +401,7 @@ HDFGrid *NewGridFromSDS(const hdf_sds & sds, const string &dataset)
 // return 0 if the HDF Type is invalid or not supported.
 BaseType *NewDAPVar(const string &varname,
 		    const string &dataset,
-		    int32 hdf_type)
+		    const int32 hdf_type)
```
jgallagher59701, line 404: Curious: Did SonarScan suggest this?

modules/hdf4_handler/hdfutil.cc
```diff
@@ -107,7 +107,7 @@ void *ExportDataForDODS(const hdf_genvec & v, int i)
 {
     switch (v.number_type()) {
     case DFNT_INT16:{
-            int16 *temp = new int16;
+            auto temp = new int16;
             *temp = v.elt_int16(i);
             return (void *) temp;
```
jgallagher59701, line 112: This is another 'curious' question: If you used `return temp;` instead of the cast, would C++-11 make the cast for you given the return type of the function is `void*`? Again, more of a question than anything else.

jgallagher59701 (approved):
Just a note and two questions. But I think is ready for a merge.

modules/hdf4_handler/HDF4Module.cc
kyang2014: I agree with you. I won't change any MACRO like this in the future.

modules/hdf4_handler/hc2dap.cc
kyang2014: No, this is a mistake I introduced. I will retract the change.

PR #708, opened 2022-11-28
Dap2 bad request error

dh-opendap:
The error message generated by the hdf5_handler when a DAP2 response is requested for a dataset containing DAP4 types that cannot be represented by DAP2 is the following:

"Your request was for a response that uses the DAP2 data model. This dataset contains variables whose data type ( Int64 ) is not compatible with that data model, causing this request to FAIL. You may try constraining your request to elide the problematic data type or ask for a different encoding such as NetCDF4 or DAP4 binary response encodings.";

but in reality you cannot use a CE to 'elide the programmatic data type' because the hdf5_handler generates a DDS based on the entire contents of the dataset, prior to libdap applying any CE, and this error is generated at the time the initial DDS is being constructed.    Should I remove the wording suggesting a CE can be used to workaround these variables?


kyang2014 (changes requested):
> "Your request was for a response that uses the DAP2 data model. This dataset contains variables whose data type ( Int64 ) is not compatible with that data model, causing this request to FAIL. You may try constraining your request to elide the problematic data type or ask for a different encoding such as NetCDF4 or DAP4 binary response encodings.";

but in reality you cannot use a CE to 'elide the programmatic data type' because the hdf5_handler generates a DDS based on the entire contents of the dataset, prior to libdap applying any CE, and this error is generated at the time the initial DDS is being constructed. Should I remove the wording suggesting a CE can be used to workaround these variables?

It is correct that CE(Constraint Expression) has no effect to the DDS. But your wording "a different encoding netCDF4 or DAP4"  has nothing to do with CE. But in this case(64-bit), the only correct encoding is to change the "response from DAP2 to DAP4".  Then DAP4 DMR instead of DDS response is requested. In this place, the encoding has nothing to do with netCDF4.  So I suggest to change this sentence 
`You may try constraining your request to elide the problematic data type or ask for a different encoding such as NetCDF4 or DAP4 binary response encodings.` to something like `You can elide the problematic data type by asking for the DAP4 binary response encoding.`

dh-opendap:
Changed message to the following:

 "Your request was for a response that uses the DAP2 data model. This dataset contains variables whose data type ( Int64 ) is not compatible with that data model, causing this request to FAIL. To access this dataset ask for the DAP4 binary response encoding.";

Thoughts?   Removed use of 'elide' since there is no way to omit these variables.  Remove NetCDF4 as part of the message, as per Kent's suggestion.

In reality, use of the IFH to request a NetCDF4 would negate any error being generated, or a NetCDF3 response requested thru the IFH would also catch this issue and also negate this error being generated.   The highest likelihood case for this error being generated would be for a direct DAP2 request so removing the NetCDF4 suggestion shouldn't present much of an issue.

kyang2014 (approved):
Looks good to me.

dh-opendap:
Merging branch, only affects HDF5_Handler.

dh-opendap:
Only HDF5_Handler affected by this branch.

PR #709, opened 2022-12-01
Hyrax 880

kyang2014:


modules/hdf5_handler/HDF5RequestHandler.cc
```diff
@@ -623,6 +623,13 @@ temp_table->print(cerr);
 
         bdas->clear_container() ;
     }
+    catch(BESSyntaxUserError & e) {
```
jgallagher59701, line 626: I think while this gets fixed (the order of these exceptions in the catch clauses) we should add `const` to the caught exception object declaration. So, "BESSyntaxError &e" becomes "const BESSyntaxUserError &e". The same change should be applied to the BESError &e declarations, too.

jgallagher59701 (changes requested):
Add the 'const' qualifier to the exception declarations in the catch clauses and merge. Thanks.

modules/hdf5_handler/HDF5RequestHandler.cc
kyang2014: Done.

jgallagher59701:
Great. Once the tests pass - assuming they do - then merge. No need to ask for a second review.

Thanks.

PR #710, opened 2022-12-07
Httpd catalog update

ndp-opendap:
Fixes the scraper so it doesn't null ptr exception on empty/missing strings, Added debugging to tests. Fixes baseline files affected by changes to REPLACE_VERSION in the modules m4 macros

### DIFF
--- a/modules/httpd_catalog_module/unit-tests/HttpdDirScraperTest.cc
+++ b/modules/httpd_catalog_module/unit-tests/HttpdDirScraperTest.cc
@@ -142,6 +142,7 @@ public:
     // Called before each test
     void setUp()
     {
+        if(debug) cerr << endl;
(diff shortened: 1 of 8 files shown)


modules/httpd_catalog_module/unit-tests/HttpdDirScraperTest.cc
```diff
@@ -142,6 +142,7 @@ class HttpdDirScraperTest: public CppUnit::TestFixture {
     // Called before each test
     void setUp()
     {
+        if(debug) cerr << endl;
```
jgallagher59701, line 145: You can also use DBG(...) in place of `if (debug) ...` for any unit tests that use the header in bes/modules

jgallagher59701 (approved):
Merge!

PR #711, opened 2022-12-10
Cmr catalog update

ndp-opendap:


modules/cmr_module/GranuleUMM.cc
```diff
@@ -0,0 +1,242 @@
...
+
+
+void GranuleUMM::setSize(const nlohmann::json& granule_obj)
+{
+    CmrApi cmrApi;
+    const auto &umm_obj = cmrApi.qc_get_object(CMR_UMM_UMM_KEY, granule_obj);
+    const auto &data_granule_obj = cmrApi.qc_get_object(CMR_UMM_DATA_GRANULE_KEY, umm_obj);
+    BESDEBUG(MODULE, prolog << CMR_UMM_DATA_GRANULE_KEY << data_granule_obj.dump(2) << endl );
+    const auto &arch_and_info_array = cmrApi.qc_get_array(CMR_UMM_ARCHIVE_AND_DIST_INFO_KEY, data_granule_obj);
+    for(const auto &entry : arch_and_info_array){
```
jgallagher59701, line 96: This should be re-written as an if since it will only be run once.

jgallagher59701 (changes requested):

modules/cmr_module/GranuleUMM.cc
ndp-opendap: It _might_ be run once. 
There is a need to process an array of objects and an "if" is not really optimal for that

ndp-opendap: See the comment I added to the code. Although in this incarnation it does get runonce I have counter examples where it may have to run more than once. At this point I would say the determination of size is a bit of a punt - the CMR records are both unreliably populated and difficult to discriminate. 🤷‍♂️

jgallagher59701 (approved):

PR #714, opened 2022-12-22
Refactor e url

jgallagher59701:
Refactor of code in bes/http. The affected classes are url, EffectiveUrl and EffectiveUrlCache.

This started as a refactor of EffectiveUrlCache ahead of work on RemoteResources, but spread.


jgallagher59701:
We should go over this for the review and not just use the diffs, which will likely be really heard to decipher.

ndp-opendap (approved):
Seems like we need some tickets:
* Remove shared_ptr usage.
* Consolidate the idea of isExpired into RemoteResource by migrating the expires code from url_impl to Remote Resource - and then of course there's EffectiveUrl needs similar treatment.

:shipit:

PR #715, opened 2022-12-27
Hyrax 888,Hyrax 882, Hyrax 891

kyang2014:
This is an attempt to fix the followings:
1)  The dimension size is not updated if the dimension is unlimited and a variable is filled with fillvalue (HYRAX-882)
2) The HDF5 fixed-size string(both  scalar and array) fill value needs to be supported(HYRAX-888)
3) Clear error messages shoulde be generated when dmrpp:ref is not specified in the dmrpp file and a data read error before it passes to the data read() function (HYRAX-891).


modules/dmrpp_module/Chunk.cc
```diff
@@ -874,14 +876,22 @@ const char *get_value_ptr(fill_value &fv, libdap::Type type, const string &v)
...
+
     const char *value = get_value_ptr(fv, d_fill_value_type, d_fill_value);
-    unsigned int value_size = get_value_size(d_fill_value_type);
 
     unsigned long long num_values = get_rbuf_size() / value_size;
+
     char *buffer = get_rbuf();
 
-    for (int i = 0; i < num_values; ++i, buffer += value_size) {
+    for (unsigned long long  i = 0; i < num_values; ++i, buffer += value_size) {
```
jgallagher59701, line 894: The type unsigned long long is guaranteed to be at least 64-bit, but it might be longer. That's not an issue here, but in the places where we want exactly 64-bits, use uint64t or int64_t. No need to change this code.

jgallagher59701 (approved):
Ready to merge. Thanks!

PR #717, opened 2022-12-31
Refactor the regex code

jgallagher59701:
Refactor the regex code. If we are moving to C++14, we might as well use the (possibly) better regex library. Right now this builds using std=c++11, but it should only build with the new code on the machines where the c++ compiler correctly implements the library.


jgallagher59701:
I made this refactor because I saw the regex code was still full of #if 0 stuff and because we have decided to move onward to C++ 2014, where the regex code is better implemented. Note that the modifications to the build I made support C++-11 compilers that correctly implement regex as well as those that don't.

jgallagher59701:
Also, the unit tests other than regexT that I changed do not have substantive edits but simple warning fixes. The PR really covers changes to BESRegex.cc/.h and unit-tests/Makefile.am and regexT.cc. The other changes are trivial.

ndp-opendap (approved):
:shipit:

PR #719, opened 2023-01-09
Migrates the is_dap4_projected() method from BES to libdap4

ndp-opendap:
This PR completes the change in which the `is_dap4_projected()` function in fileout_netcdf handler is moved into the libdap type tree beginning with `BaseType::is_dap4_projected()`


jgallagher59701 (approved):
Merge!

PR #723, opened 2023-01-20
HYRAX-927, fix the issue that causes the wrong attribute values in 1A…

kyang2014:
… and 2A GPM files

Just one line of code introduced in another ticket. Tested both hdf5 handler and fileout-netCDF for all NASA testsuites.


jgallagher59701 (approved):
Merge!

Thanks!

PR #726, opened 2023-01-24
Hyrax 859

kyang2014:
1. Make HDF5 handler default option handle the large array size.
2. The testsuite at the HDF5 handler and the fileout netCDF are updated.


jgallagher59701 (approved):
Question: Do you think the handler should use uint64_t? I know libdap4 uses int64_t because of the places where -1 is used as a flag, but I'm just thinking about the that maybe we could go right to uint64_t here in the HDF5 handler. This is more of a question than a suggestion, since it might be easy to change later and these changes pass tests now.

Let me know what you think.

It's ready to merge and thanks for all this!

kyang2014:
I think we should keep hdf5 handler consistent with libdap4 regarding the 64-bit datatype. If libdap4 uses signed 64-bit integer, HDF5 handler should also use signed 64-bit integer.

PR #729, opened 2023-01-30
Hyrax 934

kyang2014:
Just check the sonar cloud/travis. Need to clean up comments/debugging information later.


modules/fileout_netcdf/FONcArray.cc
```diff
@@ -217,6 +242,125 @@ void FONcArray::convert(vector<string> embed, bool _dap4, bool is_dap4_group) {
...
+
+        // The chunk sizes of the two fastest changing dimensions are fixed(not exceeding the MAX_CHUNK_SIZE). So set them first.
+        size_t two_fastest_chunk_dim_sizes=1;
+        auto d_chunksize_it = d_chunksizes.begin();
+
+        for (size_t i = d_a->dimensions();i>d_a->dimensions()-2; i--) {
+
+            size_t size = d_dim_sizes[i-1];
+            BESDEBUG("fonc", "FONcArray::CHUNK - dim size backward: " << size << endl);
+            d_chunksize_it = d_chunksizes.insert(d_chunksize_it, size <= MAX_CHUNK_SIZE ? size : MAX_CHUNK_SIZE);
```
jgallagher59701, line 287: For things like "size <= MAX_CHUNK_SIZE ? size : MAX_CHUNK_SIZE" you could use "min(size, MAX_CHUNK_SIZE)" or std::min(...) although I'm not sure there's a huge difference in performance. I think min() & max() are a bit easier to read, but it's probably not worth too much effort.

modules/fileout_netcdf/FONcArray.cc
```diff
@@ -61,6 +61,12 @@ vector<FONcDim *> FONcArray::Dimensions;
 
 const int MAX_CHUNK_SIZE = 1024;
 
+// Set general total maximum chunk sizes to 1M(1024x1024)
+const int GENERAL_MAX_CHUNK_SIZES = 1048576;
```
jgallagher59701, line 65: I was wondering, do we ever want to make chunks as small as 1K elements?

jgallagher59701 (approved):
I think this is ready for merge. I do have a question about 1K chunks, but it's more of a general question than a change that needs to be made. My comment about min/max is more for the future. They are defined in the header <algorithm>.

modules/fileout_netcdf/FONcArray.cc
kyang2014: We've done this all the time for 1-D array, even smaller than 1K if the array size is small. It doesn't matter too much for me since it doesn't hurt performance too much. The real issue we solved in this pull request is when the 1-D array is very big , we need to increase the chunk size.

PR #731, opened 2023-01-30
Use C++14 for the builds by default

jgallagher59701:
Use C++14 for the builds by default

Also removed the OSX_Resources stuff


ndp-opendap (approved):
WooHoo! C++14!

:shipit:

PR #733, opened 2023-02-02
Bes keys fix

jgallagher59701:
TheBESKeys now uses a unique_ptr and passes tests.
    
Some other code had to be changed because the singleton
was 'messed with' by some of the BES framework destructors
and some of the unit-tests. I added a reload_keys() method
to enable unit-tests to switch key files and/or reload the
original key file, effectively resetting changes they might
have made during the tests.


kyang2014:
Does this fix the memory leak caused by the BES key operations?

dispatch/unit-tests/old/keysT.cc
```diff
@@ -279,9 +243,6 @@ class keysT : public TestFixture {
         if (debug) cout << __func__ << END << endl;
     }
 
-    /**
```
ndp-opendap, line 282: What's the deal with the method comment headers? It helps me find each method... :(

ndp-opendap (approved):

dispatch/unit-tests/old/keysT.cc
jgallagher59701: Oh, sorry. I thought those were just auto generated and not filled in. Should I add them back in?

jgallagher59701: Wait, that's the keysT.cc code - it's now in 'old' (aka unused).

PR #744, opened 2023-02-23
modules/hdf5_handler/gctp: C99 compatibility improvements

fweimer-rh:
This change adds additional prototypes to avoid implicit function declarations, which will not be supported by future compilers by default, resulting in build failures.  Also add a missing int type, which future compilers will require, too.

The additional prototypes revealed that the wrapper function gctp_ has the wrong signature, which is also fixed in this commit.

Related to:

* https://fedoraproject.org/wiki/Changes/PortingToModernC
* https://fedoraproject.org/wiki/Toolchain/PortingToModernC


jgallagher59701:
I made a branch and submitted a PR so our tests would run/pass. Merged. Sorry for the delay. 
Thanks for the fix!

PR #747, opened 2023-03-07
Hyrax 956

kyang2014:
1. Add the DAP4 coverage support of the netCDF-CF files for HDF5 handler's default option. The grid/swath and DSG HDF5 files that follow the netCDF model and CF conventions will have the coverage maps added to the corresponding variables.
2. The tests are added both to the general and NASA testsuite.


jgallagher59701 (approved):
Merge! Thanks.

PR #748, opened 2023-03-09
Curl util refactor

jgallagher59701:


jgallagher59701:
I fixed the issue with the temporary directory creation. Short version: static mutex.

Also the temporary directory has a BES key now (Http.RemoteResource.TmpDir) with a default value of /tmp/bes_rr_tmp.

The code includes some multi-threaded tests (RemoteResourceTest.cc). 

There is still the issue with gateway and NcML, but I added two new tests in modules/gateway that read data from fnoc1.nc after it is pulled over the net using the gateway (and thus RemoteResources) and that works fine with the netCDF handler. I think the issue with the NcML code is that it does some odd things with the container.

To fix the NcML issue, we might modify RemoteResource so that it can 'release' the managed temporary file in the same way that unique_ptr can be told to stop managing a pointer. That would enable the NcML code to assume control of the temporary file. But that is a different PR.

http/RemoteResource.cc
```diff
@@ -52,362 +49,182 @@
...
+
 namespace http {
 
-RemoteResource::RemoteResource(shared_ptr<http::url> target_url, string uid, unsigned long expiredInterval)
-                               : d_remoteResourceUrl(std::move(target_url)), d_uid(std::move(uid)),
-                               d_expires_interval(expiredInterval) {
-
-    d_resourceCacheFileName.clear();
-
-    if (d_remoteResourceUrl->protocol() == FILE_PROTOCOL) {
```
ndp-opendap, line 70: These checks need to be replaced before this code can be merged to master.

jgallagher59701: We did...

PR #749, opened 2023-03-09
HYRAX-601, The default option adds DAP4 maps correctly. However, the …

kyang2014:
…CF option doesn't include time because the coordinates attribute doesn't include time. However, the time is a dimension scale and should be included as the map. Revise the CF option code and make sure these kind of dimensions are handled properly. Update the testsuite etc.


jgallagher59701 (approved):
Looks good. Merge. Thanks!

PR #753, opened 2023-04-10
 HYRAX-974-1, the map variables can contain >1 dimension scales. So r…

kyang2014:
…emove the wrong sanity check.

Note: another pull request will be created to fix the map variable orders. This ticket will remove the fake error message.


jgallagher59701 (approved):
OK. let's merge this. Thanks.

PR #755, opened 2023-04-13
Hyrax 974

kyang2014:
Make sure a map variable is ahead of all the variables that use the map variable for the DAP4 coverage support.


jgallagher59701 (approved):
I think this can be merged. I looked at some of the code smells, and the 'critical' ones mostly have to do with complexity and I think those can be ignored. You could address the 'smells' that have to do with maps and the std::less<> transparent operator, since that will improve performance a bit. That's mostly you call, but it's worth noting. 

If you want to look into the complexity, then I think the only way is to factor out the body of one of the outer 'if' or 'for'statements.

PR #757, opened 2023-04-14
Debug log refactor

jgallagher59701:
Refactor of the BES Debug, Log and StopWatch classes. No changes to behavior.


jgallagher59701:
If this passes the tests, I'll remove the #if ... #endif stuff.

dh-opendap (approved):
Look good to me, definitely need to start using std::for_each myself in place of citer interation.

PR #762, opened 2023-04-24
Found more places where AWS tokens were leaked - fixed

jgallagher59701:
Found more places where AWS tokens were leaked - fixed

Also removed code coverage altogether. Will look into other, faster, ways to compute code coverage.


dh-opendap (approved):
Looks good to me.

PR #763, opened 2023-04-26
I modified the get_dmrpp CLI tests to include two tests for -r

jgallagher59701:
The -r option is the 'reduce' option where a single file for missing values can be used for more than one data file. One of two of the new tests work. I also added a very verbose option so that the versbose option by itself is easier to use while retaining the more complete output of the orignal version.


kyang2014 (changes requested):
I think one of your tests failed and also could you provide the command-line either in the comment of github or the comment of the  jira ticket on how -r option works?  Thanks,

PR #764, opened 2023-04-27
Hyrax 781

kyang2014:
Support HDF-EOS5 grid files for the default option.
1. Currently projection GEO, SINSOID, LAMAZ and PS are supported.
2. Latitude, longitude, CF grid mapping and some CF special attributes are provided. 
3. Both hdf5 handler and fileout netCDF tests are updated. Real file tests are also updated.


jhrg (approved):
I think this is ready to be merged. Thanks.

PR #766, opened 2023-04-29
Cutil refactor 04.26.23

jgallagher59701:
Various refactoring changes in CurlUtils. The main fix in the PR is that
the http_get() function that uses a 1MB buffer has been replaced with
a version that uses a vector<char> that is sized as needed. Also, tests
for http_get_as_json() were added.


ndp-opendap (approved):
I hate to see us invest another minute of effort into `rapidjson`. I get that 
`rapidjson::Document http_get_as_json(const std::string &target_url)` needs tests but it really should be marked deprecated.

PR #767, opened 2023-05-02
Patch for three failing tests in the CMR API code.

jgallagher59701:
Patch for three failing tests in the CMR API code.

We will need to address the underlying issue, but this gets the master branch building again, albeit with a broken CMR module


ndp-opendap (approved):
Like I said earlier, without additional visibility (tickets? Something...) This module will get old and die silently s the world changes out from underneath it.

PR #768, opened 2023-05-03
Hackery so that we can build against the new hyrax-deps code

jgallagher59701:
Hackery so that we can build against the new hyrax-deps code

Once the hyrax dependencies are pushed through the CI/CD system and everyone has them on their development machines, back these changes out of the code and update the baselines based on the output of the new libraries. Look for tests (testuite.at) that include the comment GDAL3-6 jhrg 4/19/23. Remove those comments and rebuild the baselines.


ndp-opendap (approved):
one can only hope this works...

:shipit:

PR #772, opened 2023-05-16
HDF5 file check

kneumiller:
Compare the provided file with the HDF5 signature before reading.


ndp-opendap (changes requested):

modules/dmrpp_module/build_dmrpp.cc
```diff
@@ -244,6 +244,26 @@ int main(int argc, char *argv[]) {
...
+            // Read the first 8 bytes (file signature) from the file
+            char signature[8];
+            file.read(signature, 8);
+
+            // Compare the signature with the HDF5 file signature
+            const char hdf5Signature[] = { '\211', 'H', 'D', 'F', '\r', '\n', '\032', '\n' };
+
+            bool isHDF5 = std::memcmp(signature, hdf5Signature, sizeof(hdf5Signature)) == 0;
+            if (!isHDF5) {
+                cerr << "The file provided is not HDF5, currently only HDF5 files are supported" << endl;
```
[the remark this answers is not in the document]
kneumiller: Yeah, I can fix that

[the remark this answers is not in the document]
ndp-opendap: 👍

modules/dmrpp_module/build_dmrpp.cc
```diff
@@ -244,6 +244,26 @@ int main(int argc, char *argv[]) {
             // is fragile. - ndp 6/6/18
             string h5_file_path = BESUtil::assemblePath(bes_data_root, h5_file_name);
 
+            //Check to see if the file being used is a HDF5 file
+            std::ifstream file(h5_file_path, std::ios::binary);
+            if (!file) {
+                std::cerr << "Error opening file: " << h5_file_path << std::endl;
```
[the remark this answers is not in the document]
ndp-opendap: According to the contract for the `std::ifstream` _initialization constructor_ (With a filename string  parameter) , a failure to open will set the streams `failbit`.

You can see the ifstream constructor here:
- https://cplusplus.com/reference/fstream/ifstream/ifstream/

And the `failbit` stuff here
- https://cplusplus.com/reference/ios/ios_base/iostate/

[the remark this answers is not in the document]
kneumiller: I looked it up and I think calling `strerror(errno);` should work with the failbit. I added it to the code

modules/dmrpp_module/build_dmrpp.cc
```diff
@@ -244,6 +244,45 @@ int main(int argc, char *argv[]) {
             // is fragile. - ndp 6/6/18
             string h5_file_path = BESUtil::assemblePath(bes_data_root, h5_file_name);
 
+            //Use this ifstream file to run a check on the provided file's signature
+            // to see if it is an HDF5 file
+            ifstream file(h5_file_path, ios::binary);
+            if (!file) {
+                cerr << "Error opening file: " << h5_file_path << endl;
+                cerr << "Cause of error: " << strerror(errno) << endl;
```
ndp-opendap, line 252: In my first comment, I mentioned error. 
But then I did some reading and discovered that `std:ifstream` does not set errno, like ever.
This prompted me to add my second comment in which I provided some links about `std::ifstream`
 
This:
```
cerr << "Cause of error: " << strerror(errno) << endl;
```
Is not a solution for `std::ifstream`

kneumiller: I updated the error checking based on the links you posted and what I found here: [dealing correctly with badbit, failbit, eofbit, and perror()](https://gehrcke.de/2011/06/reading-files-in-c-using-ifstream-dealing-correctly-with-badbit-failbit-eofbit-and-perror/)

ndp-opendap: Well, I got something completely from that article different and I think this code still needs to be rewritten. 

* Don't  use `perror` that just writing to stderr and dow nothing at handle to handle the problem.
* Throw exceptions when things go wrong, otherwise the code just moves on as if everything is fine, and it's not, so more bad things will happen.
*

modules/dmrpp_module/DmrppD4Group.h
```diff
@@ -49,15 +49,16 @@ class DmrppD4Group: public libdap::D4Group, public DmrppCommon {
 
     DmrppD4Group &operator=(const DmrppD4Group &rhs);
 
-    virtual libdap::BaseType *ptr_duplicate(){
+    virtual libdap::BaseType *ptr_duplicate() override {
```
jgallagher59701, line 52: 'virtual' with 'override' is redundant and should be removed.

modules/dmrpp_module/DmrppD4Group.h
```diff
@@ -49,15 +49,16 @@ class DmrppD4Group: public libdap::D4Group, public DmrppCommon {
...
         return new DmrppD4Group(*this);
     }
 
 #if 1
     void set_send_p(bool state) override;
 #endif
 
-    virtual void dump(ostream & strm) const;
+
+    virtual void dump(ostream & strm) const override;
```
jgallagher59701, line 61: Same comment about 'virtual' here.

modules/dmrpp_module/build_dmrpp_util.cc
```diff
@@ -1161,50 +1165,243 @@ void add_chunk_information(const string &h5_file_name, DMRpp *dmrpp)
...
-        auto version = new D4Attribute("build_dmrpp_metadata", StringToD4AttributeType("container"));
+    //HDF5 and NetCDF3 signatures:
+    const char hdf5Signature[] = {'\211', 'H', 'D', 'F', '\r', '\n', '\032', '\n'};
+    const char netcdf3Signature[] = {'C', 'D', 'F'};
 
-        auto build_dmrpp_version = new D4Attribute("build_dmrpp", StringToD4AttributeType("string"));
-        build_dmrpp_version->add_value(CVER);
-        version->attributes()->add_attribute_nocopy(build_dmrpp_version);
+    //Read the first 8 bytes (file signature) from the file
+    char signature[8];
```
jgallagher59701, line 1215: You can use std::string here instead of a c-style array.  I'm pretty sure you can say:
string signature;
signature.reserve(8);

file.read(signature.data(), 8)

Is this better? Well, code scanners will gripe about C-style arrays, and knowing how to mix the C++ strings with code that was designed with C-arrays in mind is probably a good thing. I don't know if it's really an improvement... Your call on this change.

You can also do this using:
string signature(8);

But, it's important to use ()'s there and not braces.

modules/dmrpp_module/build_dmrpp_util.cc
```diff
@@ -1161,50 +1165,243 @@ void add_chunk_information(const string &h5_file_name, DMRpp *dmrpp)
...
+        //Reset the file stream to read from the beginning
+        file.clear();
+        file.seekg(0);
 
-        stringstream ldv;
-        ldv << libdap_name() << "-" << libdap_version();
-        auto libdap4_version =  new D4Attribute("libdap", StringToD4AttributeType("string"));
-        libdap4_version->add_value(ldv.str());
-        version->attributes()->add_attribute_nocopy(libdap4_version);
+        char newSignature[3];
```
jgallagher59701, line 1225: Same deal here.

modules/dmrpp_module/build_dmrpp_util.cc
```diff
@@ -1161,50 +1165,243 @@ void add_chunk_information(const string &h5_file_name, DMRpp *dmrpp)
...
+    //HDF5 and NetCDF3 signatures:
+    const char hdf5Signature[] = {'\211', 'H', 'D', 'F', '\r', '\n', '\032', '\n'};
+    const char netcdf3Signature[] = {'C', 'D', 'F'};
 
-        auto build_dmrpp_version = new D4Attribute("build_dmrpp", StringToD4AttributeType("string"));
-        build_dmrpp_version->add_value(CVER);
-        version->attributes()->add_attribute_nocopy(build_dmrpp_version);
+    //Read the first 8 bytes (file signature) from the file
+    char signature[8];
+    file.read(signature, 8);
```
jgallagher59701, line 1216: However, for this bit, the '8' should not be hardcoded. Do this:
file.read(signature, sizeof(signature));

That way, if the size is changed, the code won't do a bad thing, because read() is probably pretty dumb and will put those bytes into memory without checking if it's really ok to do that.

If you go the string route, use:
file.read(signature.data(), signature.size());

jgallagher59701 (approved):
You should probably make the changes I describe. The second one, with the strings, the key change is make sure you call file.read() using _either_ sizeof(signature) or signature.size(), depending on whether you keep char[] or change to string.

jgallagher59701 (approved):
You should probably make the changes I describe. The second one, with the strings, the key change is make sure you call file.read() using _either_ sizeof(signature) or signature.size(), depending on whether you keep char[] or change to string.

jgallagher59701 (approved):
You should probably make the changes I describe. The second one, with the strings, the key change is make sure you call file.read() using _either_ sizeof(signature) or signature.size(), depending on whether you keep char[] or change to string.

jgallagher59701 (approved):
You should probably make the changes I describe. The second one, with the strings, the key change is make sure you call file.read() using _either_ sizeof(signature) or signature.size(), depending on whether you keep char[] or change to string.

PR #773, opened 2023-05-16
S3 module

jgallagher59701:
Fix for the CredentialsManager. 

Misc. refactorings of related code.


jgallagher59701:
NB: This is still a draft PR - there are 26 issues I'd like to look into.

jgallagher59701:
FYI - this code has an issue in the http_get_as_json_file_test and/or http_get_as_json_http_test.

http/CredentialsManager.cc
```diff
@@ -125,19 +112,8 @@ CredentialsManager::~CredentialsManager() {
     for (auto &item: creds) {
         delete item.second;
     }
-    creds.clear();
```
ndp-opendap, line 128: Why remove this call to clear? The previous for loop deletes each AccessCredentials object  in the map {{creds}}, but it does not change the contents of cress. Thus crews is left full of broken pointers.

ndp-opendap (commented):
Some of this seems troublesome to me - I made a comment regarding one concern above. 
I couldn't sort out changes to CredentialsManager - too much churn. But I did notice that it looks to me like  thread safety stuff had been removed. Not sure how that's going to work out...

jgallagher59701:
The use of locks on the various load credentials methods was not removed. In the refactored code, load_credentials() is private, the only caller of the other two (also private) load methods and is only called from inside std::call_once(). So those are thread safe - there is no point in calling the class' mutex. That mutex is used for the get() and add() methods.

ndp-opendap (approved):

jgallagher59701:
This is still failing http::CurlUtilsTest::http_get_as_json_file_test.

ndp-opendap:
> This is still failing http::CurlUtilsTest::http_get_as_json_file_test.

I have no idea why - it's a very simply function. It should be deprecated tho - and replaced with `nlohmann` json guts. But that's a big job.

PR #774, opened 2023-05-18
Changed sonarscan properties files based on feedback for SS

jgallagher59701:
Changed sonarscan properties files based on feedback for SS


ndp-opendap (approved):
When it passes :shipit:

PR #775, opened 2023-05-19
modules/hdf5_handler/gctp: C99 compatibility improvements

jgallagher59701:
This change adds additional prototypes to avoid implicit function declarations, which will not be supported by future compilers by default, resulting in build failures.  Also add a missing int type, which future compilers will require, too.

The additional prototypes revealed that the wrapper function gctp_ has the wrong signature, which is also fixed in this commit.

### DIFF
--- a/modules/hdf5_handler/gctp/include/gctp_prototypes.h
+++ b/modules/hdf5_handler/gctp/include/gctp_prototypes.h
@@ -280,5 +280,13 @@ int bceainvint(double r_maj, double r_min, double center_lon,
 	       double center_lat, double false_east, double false_north);
 int bceainv(double x, double y, double *lon, double *lat);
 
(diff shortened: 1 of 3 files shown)


jgallagher59701:
@kyang2014 This was a PR from someone in the Fedora project. Does it look OK to you?

kyang2014 (changes requested):
I think these changes are OK. Could you add your comments somewhere to the code so that we know why we change them?
```
This change adds additional prototypes to avoid implicit function declarations, which will not be supported by future compilers by default, resulting in build failures. Also add a missing int type, which future compilers will require, too.

The additional prototypes revealed that the wrapper function gctp_ has the wrong signature, which is also fixed in this commit.
```

PR #776, opened 2023-05-24
Hyrax 1032

kyang2014:
Test the travis as well as collecting info from sonar cloud.


modules/dmrpp_module/reduce_mdf.cc
```diff
@@ -124,8 +124,10 @@ short update_sha256_file(char* m_dmrpp_fname,char* m_h5_fname,char* m_sha256_fna
      * checks if stream was opened correctly and if not creates the file and opens again
      */
     if (!sha_fstream.is_open()){
-        write_sha256_file(m_dmrpp_fname,m_h5_fname,m_sha256_fname,sha256_buf);
-        sha_fstream.open(m_sha256_fname,ifstream::in);
+        return write_sha256_file(m_dmrpp_fname,m_h5_fname,m_sha256_fname,sha256_buf);
+#if 0
```
jgallagher59701, line 128: I think that if we don't need this it should be removed. Or maybe explain why it should be left in but commented out.

PR #777, opened 2023-05-25
Small changes associated with timeout processing

ndp-opendap:


dh-opendap (approved):
Changes look fine to me.  Not sure the TODO listed in BESXmlInterface.cc is necessary.  If any/all the response_handlers would have written to the output_stream during execute() than in all likelihood the timeout would be disabled, or have already been thrown.

xmlcommand/BESXMLInterface.cc
```diff
@@ -336,7 +335,7 @@ void BESXMLInterface::execute_data_request_plan()
             __LINE__);
 
         d_dhi_ptr->response_handler->execute(*d_dhi_ptr);
-
+        // @TODO Check RequestServiceTimer here?? Might be a good idea
```
jgallagher59701, line 338: Would this be hard to do? 

It if so, then initial and date the comment.

xmlcommand/BESXMLInterface.cc
```diff
@@ -315,8 +314,8 @@ void BESXMLInterface::log_the_command()
  */
 void BESXMLInterface::execute_data_request_plan()
 {
-    vector<BESXMLCommand *>::iterator i = d_xml_cmd_list.begin();
-    vector<BESXMLCommand *>::iterator e = d_xml_cmd_list.end();
+    auto i = d_xml_cmd_list.begin();
```
jgallagher59701, line 317: This might be pedantic, but the loop could be rewritten as

for (const auto &cmd: d_xml_cmd_list) {
     cmd.prep_request();
      ...
}

jgallagher59701 (approved):
Some suggestions. But, it's ready.

xmlcommand/BESXMLInterface.cc
ndp-opendap: It would not be hard to. 
But I missed the comment before I merged.

PR #778, opened 2023-05-25
Remove pthread

jgallagher59701:
This completes the removal of any use of pthread in the BES
    
There are some source files in old or unused directories
that still have the include, and there one place in BESInterface
where it's part of a large comment. But pthreads are out AFAIK.


ndp-opendap (approved):
Yay! no more pthreads

PR #782, opened 2023-06-01
Hyrax 1063

kyang2014:
draft pull request to check if it can pass the travis test.


modules/fileout_covjson/FoDapCovJsonTransform.h
```diff
@@ -602,6 +610,7 @@ class FoDapCovJsonTransform: public BESObj {
      * @throw BESInternalError if the DDS* is null or if localfile is empty.
      */
     FoDapCovJsonTransform(libdap::DDS *dds);
+    FoDapCovJsonTransform(libdap::DMR *dmr);
```
jgallagher59701, line 613: This constructor, and the one above, should be marked with the 'explicit' keyword. That will prevent the compiler from using these in type casts.

modules/fileout_covjson/FoDapCovJsonTransform.cc
```diff
@@ -1648,6 +1662,80 @@ cout <<"grid map name: "<<(*i)->name() <<endl;
     printCoverageJSON(strm, indent, testOverride);
 }
 
+void FoDapCovJsonTransform::transform(ostream *strm, libdap::DMR *dmr, const string& indent, bool sendData, bool testOverride)
+{
+
+#if 0
```
jgallagher59701, line 1668: If you want to keep this because it's needed, that's OK, but otherwise it will be easier if it is removed.

jgallagher59701 (approved):
I made two suggestions, neither are critical, but adding the 'explicit' keyword could prevent problems in the future that are hard to find.

This can be merged. Thanks.

modules/fileout_covjson/FoDapCovJsonTransform.h
kyang2014: I will add the explicit to these two methds.

modules/fileout_covjson/FoDapCovJsonTransform.cc
kyang2014: Will remove this  #if 0  block.

PR #783, opened 2023-06-02
Fix for Python on rocky8 and start of fix for openssl

jgallagher59701:


jgallagher59701:
Lots of changes - ~20 Makefile.am files - but it's all the same change. 

This PR fixes the OPENSSL include issue(s) on Linux and OSX, but it's not perfect. If openssl is not somewhere the compiler looks, is needs to be in '/usr/local/opt/{lib,include}'. 

Also part of this PR is a fix for Rocky8 which has an old(er) python3 install and lacks ThreadedHTTPServer.

And I toggled the CmrApiTest patch since now the provider API code fails again. I took the opportunity to switch that code to the new run_test_cppunit template.

ndp-opendap (changes requested):
Just a debug change - I made the relevant comment in the code.

PR #784, opened 2023-06-06
Re-added the d4 tests

kneumiller:
More testing may be needed. May need to modify the libdap4/Array::transform_to_dap4() as well.


sawer297 (approved):
good to go

ndp-opendap (approved):
Looks like you need to :shipit: , :shipit: good...

PR #787, opened 2023-06-11
DRAFT: HYRAX-1076,HTYRAX-1061, add the check for the cases to generate DMRs …

kyang2014:
…when the coordinates attribute doesn't conform to the CF. Tests are added.


modules/hdf5_handler/h5dmr.cc
```diff
@@ -2361,9 +2368,11 @@ void handle_relative_path_cv(const D4Group *d4_grp, string &coord_name) {
     msg +="must contain at least one ../ string but this coordinate with the value <";
     msg +=coord_name +'>'+" doesn't contain one.";
 
+#if 0
```
jgallagher59701, line 2371: Remove this

jgallagher59701 (approved):
Ready to merge.

modules/hdf5_handler/h5dmr.cc
kyang2014: Done.

PR #788, opened 2023-06-12
Changes - fixed leak (unverified as of now)

jgallagher59701:
Also refactored to adopt C++14 features.


kyang2014 (approved):
The change looks fine except that I find the memory is still possible leaking with valgrind. I will create another ticket for the possible leaking.

jgallagher59701:
> The change looks fine except that I find the memory is still possible leaking with valgrind. I will create another ticket for the possible leaking.

Thanks for making that ticket.

PR #789, opened 2023-06-16
Segfault fix

jgallagher59701:
This branch contains a fix for the segmentation fault on exit we see from the CredentialsManager.

The 'fix' was to back the unique_ptr singleton class out and return to the 'at_exit' version.
I reverted just the Cred Mngr code to the version from 5/18/23 (commit 2d61577d) except
that I removed the load_ngap_s3_credentials() call and the NgapS3Credentials.h header
include.


ndp-opendap (approved):
Looks good... :shipit: !

PR #790, opened 2023-06-20
Draft: HYRAX-1082, suport NULL space variables. If this variable is dimensio…

kyang2014:
…n scale, it will map to a DAP4 dimension. Add tests for both hdf5 handler and fileout netcdf..


jgallagher59701 (approved):
The code changes look fine. Approved. Merge. Thanks!

I wonder if there is some way we can automate the valgrind test - like make 'valgrind' an 
option to the 'testsuite' program. If we could do that it could be a huge plus overall. Do you
have any suggestions on that?

PR #791, opened 2023-06-23
Hyrax 1083 -first round

kyang2014:
Mostly reduce code complexity and use unique_ptr  for the key files that handle the DAP4 support in the default option.


kyang2014:
Will check later.

modules/hdf5_handler/h5get.cc
```diff
@@ -129,76 +122,33 @@ hid_t get_attr_info(hid_t dset, int index, bool is_dap4, DSattr_t * attr_inst_pt
...
-    // used in other location, we still keep the original code.
+    // used in other locations, we still keep the original code.
     // KY 2011-11-16
 
     int ndims = H5Sget_simple_extent_ndims(aspace_id);
     if (ndims < 0) {
         string msg = "cannot get hdf5 dataspace number of dimension for attribute ";
         string attrnamestr(attr_name.begin(),attr_name.end());
         msg += attrnamestr;
+        H5Tclose(ty_id);
```
jgallagher59701, line 151: One thing you can do - not a change you need to make, but a technique that can be useful - is to make a class with a constructor and destructor that will manage the need to call multiple 'close()' functions before an exception is thrown. Those can be classes local to a file and they can help ensure that all the things are closed every time that is needed.

jgallagher59701 (approved):
I think you can replace these:

   static ObjMemCache *das_cache;
    static ObjMemCache *dds_cache;
    static ObjMemCache *datadds_cache;
    static ObjMemCache *dmr_cache;
    static ObjMemCache *lrdata_mem_cache;
    static ObjMemCache *srdata_mem_cache;

in the HDF5RequestHandler class definition with unique_ptrs. Then the handler dtor can be the default.

ll my comments are just comments, not things you have to change. The simplifications are the most important part of this. 

One suggestion for the future: we you see a bunch of places like copyright changes, make those in one PR and just merge with a message that they are changes to comments only. That makes it much easier to see what's most important in the PR.

modules/hdf5_handler/h5dmr.h
```diff
@@ -33,7 +33,7 @@
 /// HDF5 files. 
 /// 
 /// \author Hyo-Kyung Lee <hyoklee@hdfgroup.org>
-/// \author Muqun Yang <ymuqun@hdfgroup.org>
+/// \author Keng Yang <myang6@hdfgroup.org>
```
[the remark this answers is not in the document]
kyang2014: Yes, corrected.

modules/hdf5_handler/h5get.cc
kyang2014: Will discuss with you on this later.

PR #792, opened 2023-07-05
Flc thread test

jgallagher59701:
File Locking Cache thread test: Made a RAII lock class for the advisory locking 
that is applied to the whole cache (not a specific item).


jgallagher59701:
The key thing to review here is, do the tests that are in place test this new feature? 
I made a choice to _not_ throw exceptions in the AdvisoryLockGuard class' constructor and destructor but instead log errors. That might be a source of controversy...

ndp-opendap (approved):
So many stitches to create thread safety...

dh-opendap (approved):
From what I can follow, the addition of the AdvisoryLockGuard class seems better than the previous implementation, as well as noting that the cache names are, or will be, used in the cache looks to a better implementation as well.

PR #793, opened 2023-07-05
Cmr providers patch

ndp-opendap:
This PR migrates the catalog code away from the providers legacy api to the new, slower and perfectly pedantic providers api.


modules/cmr_module/CmrApi.cc
```diff
@@ -764,42 +769,85 @@ const {
...
- */
-Provider CmrApi::get_provider(const std::string &provider_id) const
+void CmrApi::get_providers_list(std::vector<std::string> &provider_ids) const
 {
     JsonUtils json;
 
-    string cmr_query_url = BESUtil::pathConcat(d_cmr_providers_search_endpoint_url,provider_id);
-    cmr_query_url += ".json";
-    BESDEBUG(MODULE, prolog << "CMR Providers Search Request Url: : " << cmr_query_url << endl);
+    stringstream cmr_query_url;
```
jgallagher59701, line 776: It's hard to see from the 'diff view' in github so maybe stringstream is needed, but in general, string's + operator is a much faster way to combine strings. It might not be the fastest (append might be faster still, but with the compiler's optimizer on, it seems to be really close). There are various sources that quote measurements of 25-30 times faster when code is optimized. FWIW.

jgallagher59701 (approved):
This is good, but take a look at the use of stringstream and also look over the code smells to see if some of those can be removed.

PR #796, opened 2023-08-07
Removed exceptions thrown in CredentialsManager::load_credemtials().

ndp-opendap:
Removed exceptions thrown in `CredentialsManager::load_credentials()` because `load_credentials() `is called from inside a `std::call_once()` and throwing exceptions in there is a bad bad plan. This was making many bad things happen.
In particular, the no more curl handles business.

Instead of throwing the exception I just write the message to the `ERROR_LOG` and to `BESDEBUG`. Which will only happen one time.

### DIFF
--- a/http/CredentialsManager.cc
+++ b/http/CredentialsManager.cc
@@ -37,6 +37,7 @@
 #include "kvp_utils.h"
 #include "BESInternalError.h"
 #include "BESDebug.h"
+#include "BESLog.h"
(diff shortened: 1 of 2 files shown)


sawer297 (approved):

kyang2014 (approved):
I am not familiar with the logic. The code change looks fine to me.

dh-opendap (commented):
Changes look good.

PR #797, opened 2023-08-07
Add try block for cURL handle release in CurlHandlePool

ndp-opendap:


modules/dmrpp_module/CurlHandlePool.cc
```diff
@@ -384,7 +394,7 @@ CurlHandlePool::get_easy_handle(Chunk *chunk) {
  *
  * @param handle
```
jgallagher59701, line 395: We should document 'replace' here. 

I have questions about what this is doing, so I think I need to checkout the branch and look at all the code. The diffs don't show enough.

modules/dmrpp_module/CurlHandlePool.cc
```diff
@@ -395,23 +405,26 @@ void CurlHandlePool::release_handle(dmrpp_easy_handle *handle) {
...
-    // Find the handle; erase from the vector; delete; allocate a new handle and push it back on
-    for (std::vector<dmrpp_easy_handle *>::iterator i = d_easy_handles.begin(), e = d_easy_handles.end(); i != e; ++i) {
-        if (*i == handle) {
-            BESDEBUG("dmrpp:5", "Found a handle match for the " << i - d_easy_handles.begin() << "th easy handle." << endl);
-            delete handle;
-            *i = new dmrpp_easy_handle();
-            break;
+    //   I stuck it in the "replace" block below. ndp 08/07/23
+
+    if(replace || !KEEP_ALIVE) {
```
jgallagher59701, line 410: I think we should talk about the KEEP_ALIVE compile-time constant, since it could make the 'replace' parameter meaningless (if KEEP_ALIVE is 0, then the if is always true no matter what replace is).

jgallagher59701 (commented):
I need to look over the real code. doing that now...

jgallagher59701:
Approved

PR #800, opened 2023-08-18
Adding build_dmrpp rejection of arrays of variable length strings.

ndp-opendap:
Because currently we just make bogus entries in the dmr++ file like this:
```
 <dmrpp:chunks compressionType="shuffle deflate" fillValue="unsupported-variable-length-string" byteOrder="LE">
    <dmrpp:chunkDimensionSizes>7</dmrpp:chunkDimensionSizes>
    <dmrpp:chunk offset="101363854" nBytes="39" chunkPositionInArray="[0]"/>
</dmrpp:chunks>
<dmrpp:VariableLengthStringArray>offset[0]:size[0],...,offset[n]:size[n]</dmrpp:VariableLengthStringArray>
```
Note the value of `dmrpp:chunks@fillValue` is `unsupported-variable-length-string`,  and presence of the aspirational `dmrpp:VariableLengthStringArray` element. 

These dmr++ entries have no software in the `dmrpp_module` to deal with these fantasy elements at runtime. 

This PR is to stop the `build_dmrpp` app from allowing this to happen by refusing to build a dmr++ for a dataset with variable length strings.

I believe an option to elide the AVLS variable from the dataset will take additional work.


ndp-opendap:
This is failing two tests in the modules/dmrpp_module/tests_build_dmrpp tests, #17 and #54, both of which attempt to build a dmr++ from a granule (`modules/dmrpp_module/data/dmrpp/chunked_scalar_string.h5`) that contains a previously  un-noticed array of variable length strings.

modules/dmrpp_module/UnsupportedTypeException.h
```diff
@@ -0,0 +1,33 @@
...
+namespace dmrpp {
+
+/**
+ * @brief Used to prevent the production of dmr++ files that contain unsupported data types.
+ */
+class UnsupportedTypeException : public std::exception  {
+
+private:
+    std::string d_msg;
+    UnsupportedTypeException() = delete;
```
jgallagher59701, line 23: Not sure what the best practice behavior for children of std::exception is, but you might want to declare/define the default versions for the copy constructor and destructor. Also, I think you should move the default ctor to the public section. I'm pretty sure leaving it private will show up as a 'code smell.'

jgallagher59701 (changes requested):
Those changes are pretty simple and probably not strictly necessary, but sonar will compain without them and it's probably good (if not 'best') practice. Overall, we need to trap this now and think about how best to deal with these strings in particular (Kent mentioned to me he has some ideas) and types we cannot handle in general.

modules/dmrpp_module/UnsupportedTypeException.h
ndp-opendap: sonar cloud could care less

I'm trying to keep the class sup simple, an d by eliminating the default constructor we also eliminate the need for setter functions for the message.

ndp-opendap: Added these as public:
    UnsupportedTypeException(UnsupportedTypeException &) = default;
    ~UnsupportedTypeException() = default;

PR #805, opened 2023-08-28
Improved handling/rejection of granules containing unsupported data types.

ndp-opendap:
Moves detection of unsupported types from the inappropriate determination of the value of the fillValue to the main variable processing loop. Instead of making bad dmr++ files exceptions are thrown and we'll hear about it.


jgallagher59701 (approved):
:shipit:

jgallagher59701 (approved):

PR #809, opened 2023-08-31
Enhance dmrpp file to support direct chunk IO 

kyang2014:
Add deflate_level and filter_mask to the dmrpp file to support the direct chunk IO.


kyang2014:
> Question: what happens if there is a Chunk in the DMR++ XML that has compression type "deflate" but no deflateLevel XML attribute?

I recollect that I informed this already.  In this case, the vector size of std::vector<unsigned int> deflate_levels is 0. No deflate levels can be retrieved.

jgallagher59701:
OK, That's good.

jgallagher59701 (approved):
Approved. Please merge. Thanks.

PR #811, opened 2023-09-06
Dmrpp new reorg

jgallagher59701:
Dmrpp new reorg. I'm floating this as an idea more than a real PR. Is this a good/better/useful way to organize our handlers that employ specializations of the libdap type classes? This came about while trying to get the casting to work for Kent's code.

I also fixed a number of bugs in the Makefile.am files in DMR++.


jgallagher59701:
Nathan, Kent,

This PR changes how the DMR++ code is organized so that it's possible to use the DmrppArray, etc., classes in other handlers. At least in theory, I have not tried that yet. I will, but before I jump in that I'd like to know how much disruption will cause for work you have going.

kyang2014 (approved):
I merged this branch to the current direct IO branch and tested it. It can compile and all the tests get passed. So this merge is OK for us.

jgallagher59701:
At this point, I'm dropping this as a PR.

PR #819, opened 2023-09-19
Hyrax 1161 - draft 

kyang2014:
May be worth for code review and merging to the master.


modules/fileout_netcdf/FONcArray.cc
```diff
@@ -668,6 +668,32 @@ void FONcArray::write_nc_variable(int ncid, nc_type var_type) {
     else
         d_a->intern_data(*get_eval(), *get_dds());
 
+    bool d_io_flag = d_a->get_dio_flag();
```
jgallagher59701, line 671: For this it is better to say BESDEBUG("fonc", "d_io_flag is " << d_a->get_dio_flag() << endl); Since BESDEBUG(x); reduces to an empty statement when building for production use. When configure is used with --enable-developer.

modules/fileout_netcdf/FONcArray.cc
```diff
@@ -668,6 +668,32 @@ void FONcArray::write_nc_variable(int ncid, nc_type var_type) {
...
+    if (d_io_flag)
+        BESDEBUG("fonc", "d_io_flag is true"<<endl);
+    else
+        BESDEBUG("fonc", "d_io_flag is false"<<endl);
+    
+    if (d_io_flag) {
+        Array::var_storage_info dmrpp_vs_info = d_a->get_var_storage_info();
+        for (const auto& def_lev:dmrpp_vs_info.deflate_levels) 
+            BESDEBUG("fonc", "deflate level: "<<def_lev<<endl);
```
jgallagher59701, line 680: For these for loops, it is probably better to wrap the loop with #ifndef NDEBUG ... #endif. The code will only be compiled when NDEBUG is not defined. The double negative logic is annoying, but using NDEBUG is used by the assert.h header, etc., so I used it without changing its definition.

modules/dmrpp_module/DmrppRequestHandler.cc
```diff
@@ -413,6 +421,24 @@ bool DmrppRequestHandler::dap_build_dap4data(BESDataHandlerInterface &dhi)
...
+
+        bool is_netcdf4_response =(dhi.data["return_command"]=="netcdf-4");
+
+        DmrppRequestHandler::is_netcdf4_enhanced_response = is_netcdf4_response;
+        if (DmrppRequestHandler::is_netcdf4_enhanced_response &&  
+                DmrppRequestHandler::is_netcdf4_classic_response)
+            DmrppRequestHandler::is_netcdf4_enhanced_response = false;
+
+        if (DmrppRequestHandler::is_netcdf4_enhanced_response) 
+            BESDEBUG(MODULE, prolog << "netcdf-4 enhanced response" << endl);
```
jgallagher59701, line 433: See my comment in the FONC code.

jgallagher59701 (approved):
We can merge this, but it would be better to change how the BESDEBUG macro is used in three places I found.

modules/fileout_netcdf/FONcArray.cc
kyang2014: Yes, will do.

modules/dmrpp_module/DmrppRequestHandler.cc
kyang2014: Code changed.

modules/fileout_netcdf/FONcArray.cc
kyang2014: Done

PR #821, opened 2023-09-21
Ngap cache

jgallagher59701:


ndp-opendap (changes requested):
There is a systemic swapping of real_name in the parameters.  
for example:
```
bool NgapContainer::get_dmrpp_cache(const string &url_key, string &real_name) 
```
But then it's used is like:
```
dmrpp_cache(get_real_name(), dmrpp_string)
```
Which is very confusing to me, a human. I think the semantics of these caching methods would be greatly improved if they stuck to consistent generic variable naming like:
```
bool NgapContainer::get_dmrpp_cache(const string &key, string &value)
```
Or simply got the associations correct:
```
bool NgapContainer::get_dmrpp_cache(const string &real_name, string &dmrpp)
```

ndp-opendap (approved):
A dizzying array of updates. I'm glad this is working! 
:shipit:

PR #822, opened 2023-09-22
Elide unsupported

ndp-opendap:
Drops variables whose `dmrpp:chunks/@fillValue` attribute has a value of
* `unsupported-string`
* `unsupported-array`
* `unsupported-compound`
* `unsupported-variable-length-string`

The `unsupported-string` case is more complex as it is an indiscriminate label that was applied to both the ArraysOfFixedLengthStrings and ArraysOfVariableLengthStrings prior to the implementation of the ArraysOfFixedLengthStrings solution.

### DIFF
--- a/modules/dmrpp_module/DMZ.cc
+++ b/modules/dmrpp_module/DMZ.cc
@@ -220,20 +220,14 @@ bool flagged_as_unsupported_type(xml_node var_node, string &unsupported_flag) {
         return is_unsupported_type;
     }
 
-    xml_attribute *fillValue = nullptr;
-    for (xml_attribute attr = chunks.first_attribute(); attr; attr = attr.next_attribute()) {
-        if (is_eq(attr.name(), "fillValue")) {
-            fillValue = &attr;
-        }
-    }
-
-    if(!fillValue) {
+    xml_attribute fillValue_attr  = chunks.attribute("fillValue");
+    if(!fillValue_attr) {
         // No fillValue attribute? Then we can be done, it's supported.
         return is_unsupported_type;
     }
 
     // We found th fillValue attribute, So now we have to deal with its various tragic values...
-    if(is_eq(fillValue->value(),UNSUPPORTED_STRING)){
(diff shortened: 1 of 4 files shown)


modules/dmrpp_module/DMZ.cc
```diff
@@ -84,16 +84,34 @@ using namespace libdap;
...
+
+
+
+
 namespace dmrpp {
 
 using shape = std::vector<unsigned long long>;
 
+// The original unsupported fillValue flags from 4/22
+constexpr static const auto UNSUPPORTED_STRING = "unsupported-string";
```
jgallagher59701, line 96: Interesting...

modules/dmrpp_module/DMZ.cc
```diff
@@ -132,13 +150,27 @@ static inline DmrppCommon *dc(BaseType *btp)
...
+}
+
 /**
  * @brief Build a DMZ object and initialize it using a DMR++ XML document
  * @param file_name The DMR++ XML document to parse.
  * @exception BESInternalError if file_name cannot be parsed
  */
 DMZ::DMZ(const string &file_name)
 {
+    load_config_from_keys();
```
jgallagher59701, line 173: I'm in a quandary about constructors that do things that can throw, esp. when those exceptions are not fatal. Actually, more and more, I think we should rarely have ctors other than the defaults or sones that just pass strings, ints, etc., onto members.

We could drop this ctor and say that to use this class, you have to load the keys or get undefined behavior (or we could put calling load...() inside an init() method. Because...

In my PR, I add another way to parse, so if the DMZ is made using the default ctor and then parse_xml_string(...) is called, the keys will never be loaded. But, while there are ways around this particular problem, it's a general issue. Code that throws should not be in a ctor. I did this all over the place in the past.

jgallagher59701 (approved):
This can be merged, but once I merge my caching stuff, we're going to have to deal with the constructor that takes a filename (and calls the parser). I think we can improve the design; that was my goof.

PR #828, opened 2023-10-12
Clean up HDF4 handler code: remove extra spaces, update comments, rem…

kyang2014:
…ove most old debugging info with #if 0 #endif blocks.

Just check if there is any issue with travis.


kyang2014:
Just changing author names, updating comments, removing #if 0 #endif blocks, also updating the download script for the HDF4 handler. No need to review.

PR #831, opened 2023-10-16
Memory cache class

jgallagher59701:


jgallagher59701:
This branch hold a general refactor of the NGAP module code so that it's updated to C++11 and so that it uses a new version of the 'memory cache' that is instantiated for each of the CMR and DMR++ caches (instead of a collection six functions and some fields in NgapRequestHandler. The caches are still local to the handler, but the new MemoryCache<V> template could be used more generally. I didn't test it with non-string values and I didn't move it out of the NGAP module, but if that seems useful in the future it should be simple enough.

ndp-opendap (approved):
Wooo! CachingPants!

:shipit:

PR #832, opened 2023-10-18
Fixed the bug where a token is leaked when a Chunk gets an error from S3

jgallagher59701:
Fixed the bug where a token is leaked when a Chunk gets an error from S3

Also found and noted that the http::url::is_expired() code is duplicated by the unused method:
NgapApi::signed_url_is_expired(const http::url &signed_url).


ndp-opendap (approved):
:shipit:

PR #833, opened 2023-10-19
Vls scalar

ndp-opendap:
Adds an implementation for the variable length string scalar (vlss)  types.
This implementation reads the value from the hdf5 file, converts the associated DAP variable to "compact" and then sets the variables value and read_p flags.
Thus when the dmr++ is serialized the scale thing valued is carried to the runtime engine.

### DIFF
--- a/modules/dmrpp_module/build_dmrpp_util.cc
+++ b/modules/dmrpp_module/build_dmrpp_util.cc
@@ -1103,34 +1103,43 @@ bool is_unsupported_type(hid_t dataset_id, BaseType *btp, string &msg){
 }
 
 /**
+ * @brief Identifies, reads, and then stores a vlss in a DAP dmr++ variable using the compact representation
  *
  * @param dataset The HDF5 dataset id.
  * @param btp The BaseType pointer to the sister DAP class
  * @return Returns true if the variable was processed, false otherwise.
  */
 bool process_variable_length_string_scalar(const hid_t dataset, BaseType *btp){
-    // Do the vlss case as a compact variable.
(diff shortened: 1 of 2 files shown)


kyang2014 (commented):
1. Before merging to the master, definitely need to have tests to verify that a scalar variable length string  is written to a dmrpp file and that the scalar variable length string can be retrieved successfully. In addition to a test in the dmrpp module, suggest to add a test in the fileout netcdf module.
2. Can generally follow the flow of the coding, but it is hard to just use change shown in the github to carefully check the code.

modules/dmrpp_module/build_dmrpp_util.cc
```diff
@@ -1117,109 +1226,72 @@ void get_chunks_for_all_variables(hid_t file, D4Group *group) {
...
-            throw;
+        else {
+            VERBOSE(cerr << prolog << "Unable to open " << btp->FQN()
+            << " with the hdf5 api. Skipping chunk production. "
+            << "Everything will be ok." << endl);
         }
     }
 
     // all groups in the group
+    // for(auto g:group->groups())
```
jgallagher59701, line 1292: One thing that sonar scan complains about is commented out code. Maybe remove this?

jgallagher59701 (approved):
One tiny suggestion... But it can go as it is.

modules/dmrpp_module/build_dmrpp_util.cc
ndp-opendap: done

PR #834, opened 2023-10-19
Removed unused ctor from MemoryCache<V>

jgallagher59701:


modules/ngap_module/MemoryCache.h
```diff
@@ -104,7 +104,11 @@ class MemoryCache {
 public:
     MemoryCache() = default;    // this makes testing easier
     MemoryCache(const MemoryCache *src) = delete;
+
+#if 0
```
ndp-opendap, line 108: Is this the unused actor? Why not just delete this instead of `#if 0`

ndp-opendap (approved):
One crufty comment

PR #835, opened 2023-10-19
Now the DMR++ is only in a string and not stored in a temp file.

jgallagher59701:
This PR removes the DMR++ --> file --> string --> filter --> file  chain that was not actually needed. I used the BESContainer attributes to communicate whether the access() method returns a string with the information or a string with the name of a file that contains the information.


ndp-opendap (approved):
:shipit:

PR #838, opened 2023-10-30
Response limits

ndp-opendap:


dap/DapUtils.cc
```diff
@@ -292,82 +299,38 @@ uint64_t crsaibv_process_variable(
...
 
     uint64_t response_size = 0;
-    if (var->send_p()) {
-        uint64_t vsize = var->width_ll(true);
-        response_size += vsize;
 
-        BESDEBUG(MODULE_VERBOSE, prolog << "  " << get_dap_decl(var) << "(" << vsize << " bytes)" << endl);
-        if (vsize > max_var_size) {
-            too_big.emplace(pair<string, uint64_t>(get_dap_decl(var), vsize));
+    if (var->is_constructor_type()) {
```
jgallagher59701, line 303: I don't know what calls this code, but it looks like the CE has already been parsed/evaluated (below there's a test for send_p()) so to be most efficient, this should only be called on variables that have send_p() set. For a Constructor, that property will be true if any descendants have it set (not just direct children).

jgallagher59701 (approved):
I added one comment about how the crsaibv_process_variable() function gets called. Look at that and edit if needed. Otherwise, it's ready for merging.

dap/DapUtils.cc
ndp-opendap: Is that also true for Groups? 

Does a Group's send_p() return true iff a child data variable also has send_p() return true???

PR #840, opened 2023-11-01
First pass at controlling max variable size and max response size for DAP2 requests

ndp-opendap:
This PR adds code to inspect DAP2 requests and determine if:
- The response will be too larger than the configured maximum. 
- If any of the requested and constrained variable are larger than the configured maximum.
And if one or both are true it will write an error message explaining the situation and then throw `BESUserSyntaxError`

### DIFF
--- a/dap/DapUtils.cc
+++ b/dap/DapUtils.cc
@@ -27,6 +27,7 @@
 #include <iostream>
 #include <sstream>
 #include <unordered_map>
+#include <cmath>
 
 #include <libdap/DDS.h>
 #include <libdap/DMR.h>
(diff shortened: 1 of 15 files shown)


dap/DapUtils.cc
```diff
@@ -484,64 +520,155 @@ std::string too_big_error_prolog(uint64_t max_response_size_bytes, uint64_t max_
...
+{
+    uint64_t max_var_size_bytes=0;
+    uint64_t max_response_size_bytes=0;
+    std::vector<std::string> too_big_vars;
+
+    get_max_sizes_bytes(max_response_size_bytes, max_var_size_bytes, true);
+    BESDEBUG(MODULE, prolog << "max_var_size_bytes: " << max_var_size_bytes << "\n");
+    BESDEBUG(MODULE, prolog << "max_response_size_bytes: " << max_response_size_bytes << "\n");
+
+    auto response_size_bytes = compute_response_size_and_inv_big_vars(dds, max_var_size_bytes, too_big_vars);
```
jgallagher59701, line 658: Lets merge this because the 'too big' requests are a problem that can crash the server. But, let's also profile this, after it is merged, to see if building the too_big_vars should be postponed until/if that info is needed.

jgallagher59701 (approved):
:shipit: 

But then profile...

dap/DapUtils.cc
ndp-opendap: I have no idea what you mean, please explain.

ndp-opendap: `1699007614|&|3363|&|timing|&|elapsed_us|&|23|&|start_us|&|1699007614778245|&|stop_us|&|1699007614778268|&|-|&|dap_utils::throw_if_too_big() - 
`

PR #843, opened 2023-11-06
Switched BESContextManager to unordered_map

ndp-opendap:
Because std:map is much slower than std::unordered_map


jgallagher59701 (approved):
:shipit: 

Thanks!!

PR #847, opened 2023-11-17
Ngap file cache

jgallagher59701:
This is the culmination of the work on DMR++ caching started at the behest of NSIDC.

A memory cache is used by each process to hold DMR++ documents for fast reuse. 
Timing data shows the memory cache returns hits in ~100uS. This is backed up by a
file cache that can be accessed across all the processes on a given host (and across
hosts if the file system supports NFS  - which AWS's EFS does). Timing data shows 
cache hits for the file cache return items in ~1mS. Network access has been refactored
so that the DMR++ is read from the wire directly into a C++ string object and this takes
about 500mS without the overhead of authentication (i.e., in tests where we use OPeNDAP's
S3 bucket and do not rely on a web service for credentials or 'URL signing').


dispatch/FileCache.h
```diff
@@ -333,23 +355,33 @@ class FileCache {
...
     /**
      * @brief Initialize the cache.
-     * @param cache_dir Directory on some filesystem where the cache will be stored.
+     * @param cache_dir Directory on some filesystem where the cache will be stored. This
+     * directory is made if it does not exist.
      * @param size Allow this many bytes in the cache
      * @param target_size When purging, remove items until this many bytes remain.
      * @return False if the cache object could not be initialized, true otherwise.
      */
     virtual bool initialize(const std::string &cache_dir, long long size, long long purge_size) {
```
ndp-opendap, line 364: Why a signed int for size and purge size? Why not `uint64_t` ??

modules/ngap_module/NgapNames.h
```diff
@@ -58,6 +58,10 @@ constexpr static auto const USE_DMRPP_CACHE = "NGAP.UseDMRppCache";
 constexpr static auto const DMRPP_CACHE_THRESHOLD = "NGAP.DMRppCacheSize";
 constexpr static auto const DMRPP_CACHE_SPACE = "NGAP.DMRppCachePurge";
 
+constexpr static auto const DMRPP_FILE_CACHE_THRESHOLD = "NGAP.DMRppFileCacheSize"; // in MB
```
ndp-opendap, line 61: At this point the server is littered with BESKeys for size with unit that are not contained in the key name but rather in a comment. And we have bytes, kb. mb, etc. 

I think we should start adding the units to the key name to remove the confusion:
```
NGAP.DMRppFileCacheSize.MB
NGAP.DMRppFileCachePurge.MB
```

modules/ngap_module/NgapRequestHandler.cc
```diff
@@ -47,18 +47,24 @@ using namespace libdap;
 using namespace ngap;
 
 // CMR caching
-unsigned int NgapRequestHandler::d_cmr_cache_size = 100;
+unsigned int NgapRequestHandler::d_cmr_cache_size = 100;    // Entries, not size in bytes, MB, etc.
 unsigned int NgapRequestHandler::d_cmr_cache_purge = 20;
```
ndp-opendap, line 51: Is this in MB? Hw would we know?

ndp-opendap (approved):

dispatch/FileCache.h
jgallagher59701: I made these sizes signed because they are used in comparisons where a negative value is possible and casting would cause that to 'flip' over to a huge positive value (so the cache would never be purged, etc.). I figured that was OK since signed 64 bit int is still really large.

modules/ngap_module/NgapNames.h
jgallagher59701: Yes. I'll do that.

modules/ngap_module/NgapRequestHandler.cc
jgallagher59701: I'll add the units there, too.

modules/ngap_module/NgapRequestHandler.cc
```diff
@@ -47,18 +47,24 @@ using namespace libdap;
...
 
 bool NgapRequestHandler::d_use_cmr_cache = false;
-MemoryCache<std::string> NgapRequestHandler::d_new_cmr_cache;
+MemoryCache<std::string> NgapRequestHandler::d_cmr_mem_cache;
 
 // DMR++ caching
-unsigned int NgapRequestHandler::d_dmrpp_cache_size = 100;
-unsigned int NgapRequestHandler::d_dmrpp_cache_purge = 20;
+unsigned int NgapRequestHandler::d_dmrpp_mem_cache_size = 100;
+unsigned int NgapRequestHandler::d_dmrpp_mem_cache_purge = 20;
```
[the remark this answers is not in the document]
jgallagher59701: res ipsa loquitur... Well, it will once I change its name.

PR #849, opened 2023-11-21
HYRAX-1241, handle the fill value of the compound datatype. Handle th…

kyang2014:
…e case when the fill value is the default fillvalue defined by the HDF5 library.


jgallagher59701 (approved):
Thanks. This is ready to be merged.

PR #850, opened 2023-11-27
Implmenting a probe to get just the redirect url

ndp-opendap:
Instead of blindly following all redirects there are times, like when talking to a signing service like TEA, the we just want the redirect URL. In the TEA case that's a signed AWS S3 URL associated with the requesting users EDL id. It takes time to follow the redirect, and since this is done as a "set-up" step we really only need the redirect URL.

This PR/branch now contains the new function `curl::get_redirect_url()` which:
- Retrieves the redirect url from the origin service, but does not follow the redirect.
- Is a drop in replacement for the previously used `curl::retrieve_effective_url()` that achieves the same result by actually following the redirects.
- Has replaced the use of `curl::retrieve_effective_url()` in `EffectiveUrlCache`


http/CurlUtils.cc
```diff
@@ -1538,17 +1542,11 @@ std::shared_ptr<http::EffectiveUrl> retrieve_effective_url(const std::shared_ptr
...
-    if( BESDebug::IsSet(CURL_TIMING) ) {
-        moo.start(prolog + "Retrieve effective url for starting_point_url: " + starting_point_url->str());
-    }
-#endif
-
     // Add the authorization headers
     request_headers = add_edl_auth_headers(request_headers);
 
     // TODO Is this really something that we might need to sign for S3 access? jhrg 11/3/22
+    //  No, its not. But this call is part of how the business get
```
jgallagher59701, line 1549: If we don't need to make this call, why not drop it?

jgallagher59701 (approved):
Looks ready to merge - I think I follow the new code you wrote.

:shipit:

PR #851, opened 2023-11-30
Hyrax 1247

kyang2014:
direct IO code ready to be reviewed.  

Add the direct IO feature to the BES. dmrpp and fileout netCDF modules are modified. DmrppArray.cc, FoNCArray.cc, Chunk.cc and SuperChunk.cc are largely updated. I try to separate direct IO code from other parts of code even though it may seem redundant for some functions.  Also implement a short-term solution to reduce the memory print of the case the direct IO is not used.


jgallagher59701 (approved):
I think this is ready. Once merged, it would be a good idea to look into some of the uses of unique_ptr. Using the make_unique<class>() utility is faster at run time than using new. Also, there are places where it would be OK to pass in the raw point (use unique_prt::get()) than pass a const reference to the unique_ptr since that is easy to confuse with passing the unique_ptr itself. But, those are minor improvements that won't change how the code works. 

I think this is ready to be merged.

PR #853, opened 2023-12-08
Hyrax 1250

kyang2014:
The allocation of data buffer is moved out of the define mode in the fileout netcdf module.


kyang2014:
This is replaced by https://github.com/OPENDAP/bes/pull/855

modules/fileout_netcdf/FONcTransform.cc
```diff
@@ -1390,6 +1451,54 @@ void FONcTransform::check_and_obtain_dimensions_internal(D4Group *grp) {
     }
 
 }
+void FONcTransform::set_constraint_var_dio_flag(libdap::Array* t_a) const {
```
jgallagher59701, line 1454: It might be the odd way GitHub formats the diffs, but please take a look at the formatting of this  function and adjust if needed.

jgallagher59701 (commented):
Ready to merge but with a note that formatting of one function should checked.

Thanks!

PR #854, opened 2023-12-08
H4 mapper

jgallagher59701:
This version of the HDF4 DMR++ Builder can process at least one MODIS file and produce values that appear correct. 
WIP.


ndp-opendap (approved):
This is big, but there's seems to be little change, beyond white space phaffing, in the older files beyond wiring this up to the regular build
🤷‍♂️ 
:shipit:

PR #855, opened 2023-12-12
Hyrax 1254

kyang2014:
The allocation of data buffer is moved out of the define mode in the fileout netcdf module, also clear memory after each variable for the direct IO case.


kyang2014:
James already reviewed 853, which is essentially this one. So merge it.

PR #857, opened 2023-12-19
Hyrax 1257

kyang2014:
generate dmrpp info for scalar structure data.


kyang2014:
Find some issues. Close the pull request for the time being.

jgallagher59701 (approved):
Looks ready to merge. Thanks!

But one thing to keep in mind: We have `const vector<BaseType*> &variables() const` in libdap::Constructor so loops that use var_begin() and var_end() can be replaced by range-based for loops using 
```
for (var: constructor->variables()) {
...
}
```

PR #860, opened 2023-12-29
HYRAX-1241, add get_dmrpp tests for HDF5 compound datatypes. The defa…

kyang2014:
…ult fill value is included.


jgallagher59701 (approved):
Thanks! Ready to merge.

PR #861, opened 2023-12-29
Keys singleton fix

jgallagher59701:
Fix for the TheBESKeys singleton.


ndp-opendap (approved):
It passes the tests, so :shipit:

But really, do any of these tests test concurrency?

PR #862, opened 2024-01-02
Build dmrpp resource log

ndp-opendap:
This PR contains:

1. A new function `log_response_and_memory_size()` used by the **build_dmrpp** service.
2. A consolidation of the `BESUtil::file_to_stream()` methods. There were three and now there is one.
3. Repairs to the lifecycle of `NgapBuildDmrppContainer`, including using a unique_ptr for its `RemoteResource`.
4. New, more detailed error messages in `FONcTransform` and in `FONcUtils::handle_error()`


dh-opendap (approved):
Changes all seem correct to me.

PR #864, opened 2024-01-02
EffectiveUrlCache singleton pattern update

ndp-opendap:


http/EffectiveUrlCache.cc
```diff
@@ -168,12 +140,12 @@ shared_ptr<EffectiveUrl> EffectiveUrlCache::get_effective_url(shared_ptr<url> so
         // the instance we placed in the cache - it can be modified and the one in the cache
         // is unchanged. Trusted state was established from source_url when effective_url was
         // created in curl::retrieve_effective_url()
-        effective_url = shared_ptr<EffectiveUrl>(new EffectiveUrl(effective_url));
+        effective_url = make_shared<EffectiveUrl>(effective_url); // shared_ptr<EffectiveUrl>(new EffectiveUrl(effective_url));
```
jgallagher59701, line 143: It's a minor point, but commented out code is going to be flagged by sonarscan as a code smell. And it probably improves the code if we remove it.

jgallagher59701: I know, I did this... my bad.

http/EffectiveUrlCache.h
```diff
@@ -83,7 +96,13 @@ class EffectiveUrlCache : public BESObj {
     std::shared_ptr<EffectiveUrl> get_effective_url(std::shared_ptr<url> source_url);
 
     void dump(std::ostream &strm) const override;
-    virtual std::string dump() const;
+
+private:
```
jgallagher59701, line 100: A comment only - not a suggestion for a change because I have no idea how this would affect the BES code base, but this version of dump could be defined in the BESObj class once.

jgallagher59701: We could try that and see how much angst it causes... ;-) (but not with this PR).

jgallagher59701 (approved):
This can be merged. Thanks for separating my stuff from the rest.

jgallagher59701:
NB: Since this was originally my thing, I'll make the one change and merge.

http/EffectiveUrlCache.cc
jgallagher59701: Done

PR #867, opened 2024-01-03
NgapBuildDmrppContainer lifecycle changes from lcfo-2023

ndp-opendap:
This is an attempt to stop the container from hanging onto a shared pointer to an object which keeps a file open until the next request.

Imported lifecycle changes for `NgapBuildDmrppContainer` from `lcfo-2023`

It is entirely possible that this attempt is just wrong. At the point I wonder if we couldn't just get the container at the end of the transmission and call  release() on the shared_ptr

### DIFF
--- a/modules/builddmrpp_module/NgapBuildDmrppContainer.cc
+++ b/modules/builddmrpp_module/NgapBuildDmrppContainer.cc
@@ -94,11 +94,11 @@ void NgapBuildDmrppContainer::initialize()
 }
 
 NgapBuildDmrppContainer::NgapBuildDmrppContainer(const NgapBuildDmrppContainer &copy_from) :
-        BESContainer(copy_from), d_data_rresource(copy_from.d_data_rresource), d_real_name(copy_from.d_real_name) {
+        BESContainer(copy_from), d_real_name(copy_from.d_real_name) {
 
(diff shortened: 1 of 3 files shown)


jgallagher59701 (approved):
I think this can be merged. It seems to have languished, but if it's still valid, then lets merge it. Otherwise, let's close this PR.

jgallagher59701:
@ndp-opendap What is the status of this PR? Should it be merged or dropped?

ndp-opendap:
It was never review. It's possible this idea is worthy, or it may be junk. Given the c++ situation in general this would require careful review

jgallagher59701 (commented):
Lets drop this or review it, merge master to it, etc.

jgallagher59701:
I'm going to close this - the changes are all in the NgapBuildDmrppContainer which has since been retired.

```bash
hyrax/bes % find . -name NgapBuildDmrpp\*
./retired/modules/builddmrpp_module/NgapBuildDmrppContainerStorage.h
./retired/modules/builddmrpp_module/NgapBuildDmrppContainer.cc
./retired/modules/builddmrpp_module/NgapBuildDmrppContainerStorage.cc
./retired/modules/builddmrpp_module/NgapBuildDmrppContainer.h
```

PR #870, opened 2024-01-11
Superchunky issue

ndp-opendap:
This small PR changes three things in SuperChunk

- Removed `SuperChunk::get_chunks()` which was used only by internal `SuperChunk` methods.
- Fixed a read accounting error in the method `SuperChunk::read_fill_value_chunk()` by setting `d_is_read=true`
- Dropped redundant calls to `SuperChunk::retrieve_data()` in `SuperChunk::read()` and  `SuperChunk::read_unconstrained()`

I doubt that they solve the issue the thread is named for, but these changes fix problems and I would like merge this and move on.


jgallagher59701 (approved):
Ready to merge.

PR #872, opened 2024-01-12
Hyrax 1280

kyang2014:
support the simple array of structure in fileout netcdf.


modules/fileout_netcdf/FONcArrayStructure.h
```diff
@@ -0,0 +1,66 @@
...
+// You can contact The HDF Group, Inc. at 410 E University Ave,
+// Suite 200, Champaign, IL 61820
+
+
+#ifndef FONcArrayStructure_h_
+#define FONcArrayStructure_h_ 1
+
+#include <libdap/Structure.h>
+#include "FONcBaseType.h"
+#include "FONcArrayStructureField.h"
```
jgallagher59701, line 31: Since FONcArrayStructureField is used as a pointer here, the header does not don't need to be included. You can replace the include with a forward declaration like:

class FONcArrayStructureField;

modules/fileout_netcdf/FONcArrayStructure.cc
```diff
@@ -0,0 +1,220 @@
...
+ * This constructor takes a DAP BaseType and makes sure that it is a DAP Array
+ * Structure instance. If not, it throws an exception
+ *
+ * @param b A DAP BaseType that should be an array of Structure
+ * @throws BESInternalError if the BaseType is not a an array
+ */
+FONcArrayStructure::FONcArrayStructure(BaseType *b) :
+    FONcBaseType()
+{
+    _as = dynamic_cast<Array *>(b);
```
jgallagher59701, line 43: I *think* you can move this to the initializer list:

FONcBaseType(), _as(dynamic_cast<Array *>(b))

but I'm not 100% sure.

modules/fileout_netcdf/FONcArrayStructure.cc
```diff
@@ -0,0 +1,220 @@
...
+ *
+ * Delete each of the FONcBaseType instances that is a part of this
+ * structure.
+ */
+FONcArrayStructure::~FONcArrayStructure()
+{
+
+    // Adapt the code from FONcStructure.cc
+    bool done = false;
+    while (!done) {
```
jgallagher59701, line 64: I think this while loop can be replaced by

for (auto &var: _vars) {
    delete var;
}

I don't thiink the code needs to call erase() because this is the destructor for the FONcArrayStructure object so its fields won't exist when this returns.

modules/fileout_netcdf/FONcArrayStructure.cc
```diff
@@ -0,0 +1,220 @@
...
+ * structure
+ */
+void FONcArrayStructure::convert(vector<string> embed, bool _dap4, bool is_dap4_group){
+    set_is_dap4(_dap4);
+    FONcBaseType::convert(embed,_dap4,is_dap4_group);
+    embed.push_back(name());
+
+    auto as_v = dynamic_cast<Structure *>(_as->var());
+    if (!as_v) {
+        string s = (string) "File out netcdf, write an array of structure was passed a " + "variable that is not a structure";
```
jgallagher59701, line 105: You can drop the '+' since it's all a constant.

modules/fileout_netcdf/FONcArrayStructure.cc
```diff
@@ -0,0 +1,220 @@
...
+ *
+ * @param b A DAP BaseType that should be an array of Structure
+ * @throws BESInternalError if the BaseType is not a an array
+ */
+FONcArrayStructure::FONcArrayStructure(BaseType *b) :
+    FONcBaseType()
+{
+    _as = dynamic_cast<Array *>(b);
+    if (!_as) {
+        auto s = (string) "File out netcdf, array of structure was passed a " + "variable that is not an array ";
```
jgallagher59701, line 45: These two strings don't need the '+' since it's all a constant.

modules/fileout_netcdf/FONcArrayStructure.cc
```diff
@@ -0,0 +1,220 @@
...
+ * @throws BESInternalError if there is a problem defining the structure
+ */
+void FONcArrayStructure::define(int ncid)
+{
+
+    if (!d_defined) {
+        BESDEBUG("fonc", "FONcArrayStructure::define - defining " << d_varname << endl);
+        vector<FONcArrayStructureField *>::const_iterator i = _vars.begin();
+        vector<FONcArrayStructureField *>::const_iterator e = _vars.end();
+        for (; i != e; i++) {
```
jgallagher59701, line 151: for (auto &fbt: _vars) {
    BESDEBUG("fonc", "defining " << fbt->name() << endl);
    fbt->define(ncid);
}

modules/fileout_netcdf/FONcArrayStructure.cc
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    if (false == isNetCDF4_ENHANCED()) {
+        throw BESInternalError("Fileout netcdf: We don't support array of structure for the classical model now. ",
+                               __FILE__, __LINE__);
+    }
+
+    Constructor::Vars_iter vi = as_v->var_begin();
+    Constructor::Vars_iter ve = as_v->var_end();
+    for (; vi != ve; vi++) {
```
jgallagher59701, line 116: I think Constructor has a method variables() that will enable th erange-base for loop, so you can write

for (auto &bt: variables()) {
    BESDEBUG("fonc", "FONcArrayStructure::convert - converting " << bt->name() << endl);
    auto fsf = new FONcArrayStructureField(bt, _as);
    vars.push_back(fsf);
    fsf->convert(embed,_dap4,is_dap4_group);
}

I think the variables() method returns a vector<BaseType*> so the type of 'bt' will be BaseType*.

modules/fileout_netcdf/FONcArrayStructure.cc
```diff
@@ -0,0 +1,220 @@
...
+        var->write(ncid);
+    }
+    BESDEBUG("fonc", "FONcArrayStructure::define - done writing " << d_varname << endl);
+}
+
+/** @brief Returns the name of the structure
+ *
+ * @returns The name of the structure
+ */
+string FONcArrayStructure::name()
```
jgallagher59701, line 188: This method can be const

modules/fileout_netcdf/FONcArrayStructure.cc
```diff
@@ -0,0 +1,220 @@
...
+ * Displays the pointer value of this instance plus instance data,
+ * including the members of the structure by calling dump on those FONc
+ * objects.
+ *
+ * @param strm C++ i/o stream to dump the information to
+ */
+void FONcArrayStructure::dump(ostream &strm) const
+{
+    BESInternalError("FONcArrayStructure::dump is not supported yet.", __FILE__, __LINE__);
+#if 0
```
jgallagher59701, line 204: Why is it not supported?

modules/fileout_netcdf/FONcArrayStructureField.cc
```diff
@@ -0,0 +1,366 @@
...
+FONcArrayStructureField::define( int ncid )
+{
+    BESDEBUG("fonc", "FONcArray::define() - defining array '" << d_varname << "'" << endl);
+
+    if (!d_defined) {
+
+        BESDEBUG("fonc", "FONcArray::define() - defining array of structure field: " << d_varname << "'" << endl);
+        auto i = struct_dims.begin();
+        auto e = struct_dims.end();
+        for (; i != e; i++) {
```
jgallagher59701, line 137: You could use the range-based for here, too.

modules/fileout_netcdf/FONcArrayStructureField.cc
```diff
@@ -0,0 +1,366 @@
...
+        }
+        auto structure_elem = dynamic_cast<Structure *>(cb);
+        if (!structure_elem)
+            throw BESInternalError("Fileout netcdf: This is not array of structure", __FILE__, __LINE__);
+
+        //loop through the structure
+        Constructor::Vars_iter vi = structure_elem->var_begin();
+        Constructor::Vars_iter ve = structure_elem->var_end();
+
+        for (; vi != ve; vi++) {
```
jgallagher59701, line 191: I think you can use structure_elem->variables() and the range-based for loop here.

modules/fileout_netcdf/FONcArrayStructureField.cc
```diff
@@ -0,0 +1,366 @@
...
+            throw BESInternalError(err, __FILE__, __LINE__);
+    }
+}
+
+/** @brief returns the name of the array structure field
+ *
+ * @returns The name of the array structure field
+ */
+string
+FONcArrayStructureField::name()
```
jgallagher59701, line 303: This method can be const

modules/fileout_netcdf/FONcArrayStructureField.cc
```diff
@@ -0,0 +1,366 @@
...
+{
+    return var_name ;
+}
+
+/** @brief returns the netcdf type of the DAP object
+ *
+ * @returns The nc_type of this field.
+ */
+nc_type
+FONcArrayStructureField::type()
```
jgallagher59701, line 313: This methods can be const

modules/fileout_netcdf/FONcArrayStructureField.cc
```diff
@@ -0,0 +1,366 @@
...
+}
+
+// The following method is adapted from FONcArray.cc.
+FONcDim *
+FONcArrayStructureField::find_sdim(const string &name, int64_t size) {
+
+    FONcDim *ret_dim = nullptr;
+    vector<FONcDim *>::iterator i = FONcArrayStructureField::SDimensions.begin();
+    vector<FONcDim *>::iterator e = FONcArrayStructureField::SDimensions.end();
+    for (; i != e && !ret_dim; i++) {
```
jgallagher59701, line 325: Range-baed for.

Also, I don't think you need to test for !empty().

modules/fileout_netcdf/FONcDim.cc
```diff
@@ -101,6 +106,30 @@ void FONcDim::define(int ncid)
     }
 }
 
+void FONcDim::define_struct(int ncid)
+{
+    if (!_defined) {
+        if (_name.empty()) {
+            ostringstream dimname_strm;
```
jgallagher59701, line 113: The stringstream classes are pretty slow. Another way to do this (with C++-14) is: _name = "sdim" + to_string(++FONcDim::StructDimNameNum);

jgallagher59701 (commented):
Look over these ideas and see if they make sense.

jgallagher59701 (commented):
Look over these ideas and see if they make sense.

modules/fileout_netcdf/FONcArrayStructure.cc
kyang2014: Your way is much elegant. I simply copied the destructor from FONcStructure.cc. I change the code to use your way.

modules/fileout_netcdf/FONcArrayStructure.cc
kyang2014: Yes. change it.

modules/fileout_netcdf/FONcArrayStructure.cc
kyang2014: Your way is more elegant. Adopt your suggestion. Change the code. Thanks

modules/fileout_netcdf/FONcArrayStructure.cc
kyang2014: Yes,  just copied the old way from FONcStructure.cc. Use the concise way.  Change the code.

modules/fileout_netcdf/FONcArrayStructure.cc
kyang2014: Thought it was not essential. Since you asked, just add the code to support this method.

modules/fileout_netcdf/FONcArrayStructure.cc
kyang2014: This is virtual method that inherits from FONcBase.h. I need to change a lot of header files to make it compile. So skip it until we want to update the whole module.

modules/fileout_netcdf/FONcArrayStructureField.cc
kyang2014: This cannot be done with range-based and the !empty() is a key check to ensure the generation of dimension names .

modules/fileout_netcdf/FONcArrayStructureField.cc
kyang2014: Will not change since I need to change many header files.

modules/fileout_netcdf/FONcArrayStructureField.cc
kyang2014: Will not change because I need to change many headers.

modules/fileout_netcdf/FONcArrayStructureField.cc
kyang2014: Yes, change the code according to your suggestions.

modules/fileout_netcdf/FONcArrayStructureField.cc
kyang2014: Yes, change the code.

modules/fileout_netcdf/FONcDim.cc
kyang2014: Yes, the code is changed by following your suggestion.

jgallagher59701 (approved):
Looks good. Thanks!

PR #873, opened 2024-01-16
Switching BES Travis to jdk17

ndp-opendap:


jgallagher59701:
I think we should close this PR. Do you agree @ndp-opendap ?

PR #875, opened 2024-01-23
Fixing variable representation

ndp-opendap:
Making the TooBig message show the ce instead of the actual unconstrained dimension sizes when displaying the names of the variables that are too big.


jgallagher59701 (approved):
Ready to merge.

Issue #883, opened 2024-02-23
Missing #include <cstring> in http/ directory


rwmjones:
```
diff -ur bes-3.21.0-46.old/http/CurlUtils.cc bes-3.21.0-46.new/http/CurlUtils.cc
--- bes-3.21.0-46.old/http/CurlUtils.cc	2024-01-23 23:28:51.000000000 +0000
+++ bes-3.21.0-46.new/http/CurlUtils.cc	2024-02-23 13:19:11.556704797 +0000
@@ -27,6 +27,7 @@
 #include <unistd.h>
 #include <fcntl.h>
 #include <ctime>
+#include <cstring>
 
 #include <curl/curl.h>
 
diff -ur bes-3.21.0-46.old/http/RemoteResource.cc bes-3.21.0-46.new/http/RemoteResource.cc
--- bes-3.21.0-46.old/http/RemoteResource.cc	2024-01-23 23:28:51.000000000 +0000
+++ bes-3.21.0-46.new/http/RemoteResource.cc	2024-02-23 13:19:39.630855718 +0000
@@ -28,6 +28,7 @@
 
 #include <cstdio>
 #include <unistd.h>
+#include <cstring>
 
 #include <sstream>
 #include <string>
diff -ur bes-3.21.0-46.old/http/tests/remote_resource_tester.cc bes-3.21.0-46.new/http/tests/remote_resource_tester.cc
--- bes-3.21.0-46.old/http/tests/remote_resource_tester.cc	2024-01-23 23:28:51.000000000 +0000
+++ bes-3.21.0-46.new/http/tests/remote_resource_tester.cc	2024-02-23 13:20:12.830034191 +0000
@@ -6,6 +6,7 @@
 
 #include <iostream>
 #include <string>
+#include <cstring>
 
 #include <sys/stat.h>
 #include <unistd.h>
```

rwmjones:
With GCC 14 you get errors about missing functions like `strlen` and `strerror`.  The patch above fixes it.

jgallagher59701:
Until we can get out from under using AWS/S3, these PRs will fail when Travis tries to get the binary dependency packages from S3. We binaries because travis will timeout if we build the whole source collection.

I can copy this PR over to OPeNDAP's github and resubmit to get it to go through CI/CD. Would you like me to do that?

rwmjones:
Sure, although it should also be pretty easy to test if you just build locally with a sufficiently recent GCC.

PR #886, opened 2024-02-23
Fix missing includes

rwmjones:
Fixes: https://github.com/OPENDAP/bes/issues/883


jgallagher59701 (approved):

jgallagher59701:
I merged this - it seems reasonable and getting our BES CI/CD working for PRs from forks seems like it will take too long, leaving this to hang for weeks.

Thanks for the fix.

Issue #877, opened 2024-01-28
build failed of bes 3.21.0-46 on Fedora


topazus:
```bash
Making all in xmlcommand
make[2]: Entering directory '/home/ruby/rpmbuild/SOURCES/bes-3.21.0/xmlcommand'
Making all in .
make[3]: Entering directory '/home/ruby/rpmbuild/SOURCES/bes-3.21.0/xmlcommand'
depbase=`echo BESXMLInterface.lo | sed 's|[^/]*$|.deps/&|;s|\.lo$||'`;\
/bin/sh ../libtool  --tag=CXX   --mode=compile g++ -DHAVE_CONFIG_H -I. -I..  -I../dap -I../dispatch -I/usr/include/libxml2 -DWITH_GZFILEOP  -I/include  -g -O2  --pedantic --std=c++14 -MT BESXMLInterface.lo -MD -MP -MF $depbase.Tpo -c -o BESXMLInterface.lo BESXMLInterface.cc &&\
mv -f $depbase.Tpo $depbase.Plo
depbase=`echo BESXMLCommand.lo | sed 's|[^/]*$|.deps/&|;s|\.lo$||'`;\
/bin/sh ../libtool  --tag=CXX   --mode=compile g++ -DHAVE_CONFIG_H -I. -I..  -I../dap -I../dispatch -I/usr/include/libxml2 -DWITH_GZFILEOP  -I/include  -g -O2  --pedantic --std=c++14 -MT BESXMLCommand.lo -MD -MP -MF $depbase.Tpo -c -o BESXMLCommand.lo BESXMLCommand.cc &&\
mv -f $depbase.Tpo $depbase.Plo
depbase=`echo BESXMLUtils.lo | sed 's|[^/]*$|.deps/&|;s|\.lo$||'`;\
/bin/sh ../libtool  --tag=CXX   --mode=compile g++ -DHAVE_CONFIG_H -I. -I..  -I../dap -I../dispatch -I/usr/include/libxml2 -DWITH_GZFILEOP  -I/include  -g -O2  --pedantic --std=c++14 -MT BESXMLUtils.lo -MD -MP -MF $depbase.Tpo -c -o BESXMLUtils.lo BESXMLUtils.cc &&\
mv -f $depbase.Tpo $depbase.Plo
depbase=`echo BESXMLDefaultCommands.lo | sed 's|[^/]*$|.deps/&|;s|\.lo$||'`;\
[...]

PR #887, opened 2024-02-23
Fix missing libxml2 headers

rwmjones:
Required by changes in libxml2 >= 2.12

Fixes: https://github.com/OPENDAP/bes/issues/877

### DIFF
--- a/cmdln/CmdTranslation.cc
+++ b/cmdln/CmdTranslation.cc
@@ -34,6 +34,8 @@
 #include <list>
 #include <map>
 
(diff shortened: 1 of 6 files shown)


jgallagher59701:
I see you've run into the S3 credential morass. In the past I have run these PRs myself, but I think that is becoming less practical. I'll look into storing the needed dependency files on our web site so we can have open read access. Likely Monday.

jgallagher59701 (approved):

PR #888, opened 2024-02-26
HYRAX-1337, add the support of IEEE big-endian floating-point data.

kyang2014:
Add IEEE big-endian floating-point data support to dmrpp module. Both array and scalar 32-bit and 64-bit floating-point types are supported. Tests are added to dmrpp and fileout netcdf.


modules/dmrpp_module/DmrppArray.cc
```diff
@@ -2272,6 +2272,24 @@ bool DmrppArray::read()
                 }
                 break;
             }
+            case dods_float32_c: {
```
jgallagher59701, line 2275: One thing I've been thinking about is making it easier to write unit tests for our code. For this section, if these two case statement blocks called a functions to swap the bytes, then those functions would be easier to test in isolation. That might be one side effect of following the sonar scan advice to reduce the complexity.

The first block could be something like:

case dods_float32_c:
    swap_float32(get_buf());
    break;

kyang2014: Make the swap separate functions.

jgallagher59701 (approved):
Thanks. Ready to merge.

PR #892, opened 2024-02-28
HYRAX-1338, 1) Add the support to handle the HDF4 SDS fill value code…

kyang2014:
…. 2) Fix the bug that generates the wrong value when the datatype is big-endian 3) Add corresponding tests to the HDF4 handler, dmrpp and fileout netCDF.


jgallagher59701 (approved):
Ready to merge. Thanks!

PR #894, opened 2024-03-01
Refactor to remove use of unique_ptr as singleton.

jgallagher59701:
Refactor to remove use of unique_ptr as singleton.

Also includes more tests, but no change to functionality. 
Question: Should we include other signals besides SIGPIPE?


ndp-opendap (approved):
This is a good thing.

:shipit:

PR #895, opened 2024-03-01
HYRAX-1302, add the code that generates dmrpp that contains linked bl…

kyang2014:
…ocks. A test is also added.


kyang2014:
No need to check travis now. Will open the pull request later.

PR #897, opened 2024-03-06
Removed the last vestige of the unique_ptr singleton in TheBESKeys.

jgallagher59701:
Removed the last vestige of the unique_ptr singleton in TheBESKeys.


ndp-opendap (approved):
I think of you might dump the `#if 0` code blocks (since their removal is the whole point here) and then this is good to go.

:shipit:

PR #898, opened 2024-03-06
Hyrax 1302 -draft

kyang2014:
Support HDF4 SDS in the dmrpp. It includes both generating dmrpp and read the data from the dmrpp. Highlights are: 1) Object references are used to retrieve the data. 2) linked blocks are supported. 3) big-endian floating-point conversions are supported. 4) fill values are handled. 5) Tests are added and one NASA file is also tested.


modules/dmrpp_module/DmrppArray.cc
```diff
@@ -1336,6 +1337,203 @@ void DmrppArray::read_chunks_dio_unconstrained()
...
+    string filters_string = this->get_filters();
+    if (filters_string.find("deflate")!=string::npos) {
+        char *in_buf = get_buf();
+
+        char **destp = nullptr;
+        char *dest_deflate = nullptr;
+        unsigned long long out_buf_size = 0;
+        unsigned long long dest_len = get_var_chunks_storage_size();
+        unsigned long long src_len = get_var_chunks_storage_size();
+        dest_deflate = new char[dest_len];
```
jgallagher59701, line 1387: You could use a vector<char> here and then would not need the delete[] and would not need to guard against exceptions. vector has a data() method that returns a pointer to the vector.

modules/dmrpp_module/DmrppArray.cc
```diff
@@ -1336,6 +1337,203 @@ void DmrppArray::read_chunks_dio_unconstrained()
...
+        }
+        else {
+            // If we're not done and just ran out of buffer space, we need to extend the buffer.
+            // We may encounter this case when the deflate filter is used twice. KY 2022-08-03
+            if (0 == z_strm.avail_out) {
+
+                /* Allocate a buffer twice as big */
+                size_t outbuf_size = nalloc;
+                nalloc *= 2;
+                char *new_outbuf = new char[nalloc];
```
jgallagher59701, line 1513: Same comment about vector here.

modules/dmrpp_module/DMZ.cc
```diff
@@ -1544,6 +1557,44 @@ void DMZ::process_chunk(DmrppCommon *dc, const xml_node &chunk) const
...
+        }
+        else if (is_eq(attr.name(), "nBytes")) {
+            size = attr.value();
+        }
+    }
+
+    if (offset.empty() || size.empty())
+        throw BESInternalError("Both size and offset are required for a block node.", __FILE__, __LINE__);
+    if (!href.empty()) {
+        shared_ptr<http::url> data_url(new http::url(href, href_trusted));
```
jgallagher59701, line 1588: Now that we have C++14 we can use make_shared<http::url>(href, href_trusted) here. The make_shared<>() special form produces a shared_ptr that is quite a bit more efficient than one made using new.

modules/dmrpp_module/DmrppArray.cc
```diff
@@ -1336,6 +1337,203 @@ void DmrppArray::read_chunks_dio_unconstrained()
...
+        char *in_buf = get_buf();
+
+        char **destp = nullptr;
+        char *dest_deflate = nullptr;
+        unsigned long long out_buf_size = 0;
+        unsigned long long dest_len = get_var_chunks_storage_size();
+        unsigned long long src_len = get_var_chunks_storage_size();
+        dest_deflate = new char[dest_len];
+        destp = &dest_deflate;
+        out_buf_size = inflate_simple(destp, dest_len, in_buf, src_len);
```
jgallagher59701, line 1389: Sonar reports that out_but_size is never used.

jgallagher59701 (approved):
My suggestions are pretty minor - make them if you have time. 

Thanks for making these changes.

Ready to merge.

PR #901, opened 2024-03-12
Hyrax 1303

kyang2014:
The following features are added:
1) Map HDF4 vdata to dmrpp
2) Make dmrpp successfully access HDF4 vdata objects
3) Various tests are added.


modules/hdf4_handler/HDF4RequestHandler.cc
```diff
@@ -103,7 +103,7 @@ void close_hdf4_fileid(const int sdfd,const int fileid,HDFSP::File*h4file);
 bool rw_das_cache_file(const string & filename, DAS *das_ptr,bool rw_flag);
 bool r_dds_cache_file(const string & cache_filename, DDS *dds_ptr,const string & hdf4_filename);
 
-bool HDF4RequestHandler::_direct_dmr               = false;
+bool HDF4RequestHandler::_direct_dmr               = true;
```
jgallagher59701, line 106: Does this mean that the HDF4 handler now builds DMR responses without first building a DDS (and DAS)?

modules/dmrpp_module/DmrppArray.cc
```diff
@@ -1346,7 +1346,7 @@ void DmrppArray::read_linked_blocks(){
         throw BESInternalError("The number of linked blocks must be >1 to read the data.", __FILE__, __LINE__);
 
     //Change to the total storage buffer size to just the compressed buffer size.
-    reserve_value_capacity_ll_byte(get_var_chunks_storage_size());
+    //reserve_value_capacity_ll_byte(get_var_chunks_storage_size());
```
jgallagher59701, line 1349: Sonar will probably ask for the commented out code to be removed.

modules/dmrpp_module/DmrppArray.cc
```diff
@@ -1364,6 +1364,7 @@ void DmrppArray::read_linked_blocks(){
     for (unsigned int i = 1; i < num_linked_blocks; i++)
         accumulated_lengths[i] = individual_lengths[i-1] + accumulated_lengths[i-1];
 
+#if 0
```
jgallagher59701, line 1367: I think these #if 0 ... #endif blocks should be removed unless we are pretty sure they will be used later on.

modules/dmrpp_module/DmrppArray.cc
```diff
@@ -1373,25 +1374,74 @@ void DmrppArray::read_linked_blocks(){
...
+        if (filters_string.find("deflate")!=string::npos) {
+    
+            char *in_buf = get_buf();
+    
+            char **destp = nullptr;
+            char *dest_deflate = nullptr;
+            unsigned long long out_buf_size = 0;
+            unsigned long long dest_len = get_var_chunks_storage_size();
+            unsigned long long src_len = get_var_chunks_storage_size();
+            dest_deflate = new char[dest_len];
```
jgallagher59701, line 1435: For situations like this, you could use a unique_ptr<char[]> or a vector<char>. The benefit is that you do not have to delete (or delete[]) the memory yourself and thus have no leak if other code (e.g., reserve_value_capacity_ll_byte()) throws an exception.

For a vector '.data()' returns a pointer to the block of memory. Use .get() with a unique_ptr.

modules/dmrpp_module/DmrppArray.cc
```diff
@@ -2965,12 +3021,14 @@ void DmrppArray::read_array_of_structure(vector<char> &values) {
 
     size_t values_offset = 0;
     int64_t nelms = this->length_ll();
+if (this->twiddle_bytes()) 
+    BESDEBUG(dmrpp_3, prolog << "swap bytes " << endl);
 
     for (int64_t element = 0; element < nelms; ++element) {
 
         auto dmrpp_s = dynamic_cast<DmrppStructure*>(var()->ptr_duplicate());
```
jgallagher59701, line 3029: I wonder if we should be checking if dmrpp_s is null here?

jgallagher59701 (approved):
Some suggestions - take a look at them, but it's your call.

modules/hdf4_handler/HDF4RequestHandler.cc
kyang2014: Yes.

kyang2014: The issue remains is that I haven't had the data access implemented. It is inside a ticket and hopefully it will be in before the next release.

modules/dmrpp_module/DmrppArray.cc
kyang2014: Have another pull request, I made a note to remove this in that pull request.

modules/dmrpp_module/DmrppArray.cc
kyang2014: Remove it or leave a comment in the next pull request.

modules/dmrpp_module/DmrppArray.cc
kyang2014: This is the from the existing way and I recall vector was not working. I will leave it now and will check this in the future.

modules/dmrpp_module/DmrppArray.cc
kyang2014: Yes, add it in the new pull request.

PR #903, opened 2024-03-20
Fixing issues identified by asan

ndp-opendap:
Problems discovered by ASAN in the dmrpp_module are remedied here. These changes alleviate several ASAN problems:

1.  Heap overflow issues in DmrppArray and DmrppStr caused by the order of evaluation in loop conditions used to read string values.
2. Variable alignment problems in the retrieval of the Chunk checksum from the data stream. In this case the casting of the computed point location:
`            auto f_checksum = *(uint32_t *)(get_rbuf() + get_rbuf_size() - FLETCHER32_CHECKSUM);`
Was attempting to assign `f_checksum` to a value that did not start on a 4 byte boundary. 
My patch uses a temporary stack variable and memcpy to get the checksum value into it. This produces the correct result and resolves the ASAN error. 
3. Heap overflow problems in `build_dmrpp_util.cc` that I fixed by normalizing the way the code declares and initializes `vector<>` objects.
4. Indexing issues caused by setting the initial value of a count-down index to the size() of a vector and not to size()-1 in `merge_dmrpp.cc` and `check_dmrpp.cc`

### DIFF
--- a/modules/dmrpp_module/Chunk.cc
+++ b/modules/dmrpp_module/Chunk.cc
@@ -803,14 +803,12 @@ void Chunk::filter_chunk(const string &filters, unsigned long long chunk_size, u
                 throw BESInternalError("fletcher32 filter: buffer size is less than the size of the checksum", __FILE__, __LINE__);
             }
 
-            auto f_chksum_ptr = get_rbuf() + get_rbuf_size() - FLETCHER32_CHECKSUM;
-
+            // Where is the checksum value?
+            auto data_ptr = get_rbuf() + get_rbuf_size() - FLETCHER32_CHECKSUM;
+            // Using a temporary variable ensures that the value is correctly positioned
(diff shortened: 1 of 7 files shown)


modules/dmrpp_module/Chunk.cc
```diff
@@ -803,7 +803,12 @@ void Chunk::filter_chunk(const string &filters, unsigned long long chunk_size, u
                 throw BESInternalError("fletcher32 filter: buffer size is less than the size of the checksum", __FILE__, __LINE__);
             }
 
-            auto f_checksum = *(uint32_t *)(get_rbuf() + get_rbuf_size() - FLETCHER32_CHECKSUM);
+            // Where is the checksum value?
```
jgallagher59701, line 806: Good fix

jgallagher59701 (approved):
Woo Hoo! Merge. 

Thanks!

PR #904, opened 2024-03-21
Hyrax 1356

kyang2014:
Add some tests to handle HDF4 to dmrpp mapping and the related modules.


kyang2014:
No need to review since only tests are added. Merge it.

PR #906, opened 2024-03-22
Hyrax 1336

kyang2014:
Handle the case when the dimension size is 0 for an HDF4 SDS. This is legal for an SDS that has an unlimited dimension. Now we simple remove that dimension and the corresponding SDS objects since there are no data inside.
Tests are added.


jgallagher59701 (approved):
Ready to merge.

Thanks.

PR #907, opened 2024-03-27
HYRAX-1361, install build_dmrpp_h4 and make sure that the direct DMR …

kyang2014:
…is generated by default.


jgallagher59701 (approved):
Ready to merge. Thanks!

PR #908, opened 2024-03-28
Hyrax 1361 - add get_dmrpp_h4 python script

kyang2014:
A python script is provided for get_dmrpp_h4. Users just need to provide an HDF4 file name to generate the corresponding dmrpp file if they have Hyrax installed.


kyang2014:
Will change the default behavior of the HDF4 handler. The python script needs to be changed.

PR #909, opened 2024-04-01
Hyrax 1361

kyang2014:
Revert the default behavior to the original one. Also provide a python script to generate a dmrpp file just based on the HDF4 file name.


jgallagher59701 (approved):
Ready to merge. Thanks very much for this!

PR #912, opened 2024-04-10
Hyrax 1369

kyang2014:
1. Add HDF-EOS2 latitude/longitude support in the dmrpp module.
2. Add HDF-EOS2 latitude/longitude support in the hdf4 handler.


modules/dmrpp_module/build_dmrpp_h4/build_dmrpp_util_h4.cc
```diff
@@ -767,14 +767,236 @@ cout <<"length is: "<<length <<endl;
...
+    string grid_name = eos_ll_attr_value.substr(0,space_pos);
+    string ll_name = eos_ll_attr_value.substr(space_pos+1);
+
+    VERBOSE(cerr<<"grid_name: "<<grid_name <<endl);
+    VERBOSE(cerr<<"ll_name: "<<ll_name<<endl);
+    bool is_lat = true;
+    if (ll_name =="lon")
+        is_lat = false;
+
+    int32 gridfd = GDopen(const_cast < char *>(filename.c_str()), DFACC_READ);
```
jgallagher59701, line 813: These const_cast<>() uses could be a problem. Are you sure the GDopen() and attach() calls don't try to write to the strings? If they do, we could change this to:

std::vector<char> name(str.begin(), str.end());
data.push_back('\0');

then 

GDopen((char *)name.data(), DFACC_READ);

With the caveat that I've not tested this, so it might contain mistakes. 

This is not needed, however, if we know the strings won't be modified.

jgallagher59701 (approved):
One question about const_cast<char*>. Thank you!

modules/dmrpp_module/build_dmrpp_h4/build_dmrpp_util_h4.cc
kyang2014: We have to use const_cast. I haven't tried  the way you listed. Will do that later. The file name won't be modified.

PR #914, opened 2024-04-16
Hyrax 1400

kyang2014:
1. Add TRMM lat/lon variables and the corresponding attributes to the dmr file.
2.  Add TRMM lat/lon values to the dmrpp file.
3. Add tests.


jgallagher59701 (approved):
Ready to merge. Thanks!

PR #916, opened 2024-04-22
HYRAX-1363, for the dmrpp module, add big-endian scalar HDF5 compound…

kyang2014:
… datatype tests for both build_dmrpp and dmrpp access.


kyang2014:
There is no source code for the BES. Only tests are added to the dmrpp and build_dmrpp modules.

PR #917, opened 2024-04-22
HYRAX-1404, add HDF-EOS2 swath and point data dmr and dmrpp tests to …

kyang2014:
…the hdf4 handler,dmrpp and fileout netcdf.


kyang2014:
No need to review. Only tests are added and updated.

PR #922, opened 2024-05-17
Fixed what I think is an error in FileCache.

jgallagher59701:
The private method always zeroed the cache size information in the cache info file (which doubles as a cache lock file). I modified the code to look at the size of the cache info file and if it's zero length, initialize it to zero. Otherwise, do not set it to zero.

### DIFF
--- a/dispatch/FileCache.h
+++ b/dispatch/FileCache.h
@@ -212,16 +212,18 @@ class FileCache {
         return sb.st_size;
     }
 
-    /// Open the cache info file and write a zero to it.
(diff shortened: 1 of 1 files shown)


dh-opendap (approved):
The fix explanation seems logical to me.

jgallagher59701:
@ndp-opendap @dh-opendap Please review

ndp-opendap (commented):
:shipit:

ndp-opendap (approved):

PR #923, opened 2024-05-17
Added a macro to streamline BESStopWatch use.

jgallagher59701:
Added a macro to start a timer if any of a set of BESDebug flags it tests are set.
The advantage of this macro is that it drops to zero code when NDEBUG is defined.
Our code often uses a macro that leaves the BESStopWatch object definition in the 
code when NDEBUG is defined. This is not a problem, but it does make the code a bit
slower.


jgallagher59701:
@ndp-opendap @dh-opendap Please review

ndp-opendap (approved):
:shipit:

PR #925, opened 2024-05-22
Hyrax 1358 new

kyang2014:
Make the HDF4 handler read the DAP4 data.  
Four cases are covered:
1) SDS
2) Vdata
3) HDFEOS2 lat/lon
4) trmm 7 lat/lon


modules/hdf4_handler/HDFDMRArray_EOS2LL.cc
```diff
@@ -0,0 +1,283 @@
...
+    string err_msg;
+
+    // Obtain offset,step and count from the client expression constraint
+    int nelms = format_constraint(offset.data(),step.data(),count.data());
+
+    int32 gridfd = GDopen(const_cast < char *>(filename.c_str()), DFACC_READ);
+    if (gridfd <0) {
+        ostringstream eherr;
+        err_msg = "HDF-EOS: GDopen failed";
+        throw InternalErr (__FILE__, __LINE__, err_msg);
```
jgallagher59701, line 50: These would be better as BESInternalError() exceptions. Lets not make that change with this PR, but make a new ticket for that change since I think there will be many of these call to update. Also, it's not really anything wrong with this PR but instead it is an improvement to the handler.

modules/hdf4_handler/HDFDMRArray_EOS2LL.cc
```diff
@@ -0,0 +1,283 @@
...
+    }
+
+
+    r = GDij2ll (projcode, zone, params, sphere, xdim, ydim, upleft, lowright,
+                 xdim * ydim, rows.data(), cols.data(), lon.data(), lat.data(), pixreg, origin);
+
+    if (r != 0) {
+        GDdetach(gridid);
+        GDclose(gridfd);
+        err_msg = "cannot calculate grid latitude and longitude for grid name: " + gridname;
```
jgallagher59701, line 146: I think this err_msg value is lost. Should this be a place where an exception is thrown?

modules/hdf4_handler/HDFStructure.cc
```diff
@@ -79,6 +81,59 @@ void HDFStructure::set_read_p(bool state) {
...
+
+#if 0
+    Constructor::Vars_iter vi = this->var_begin();
+    Constructor::Vars_iter ve = this->var_end();
+
+    for (; vi != ve; vi++) {
+#endif
+
+    for (auto &bt:this->variables()) {	     
+        //BaseType *bt = *vi;
```
jgallagher59701, line 94: Sonar scan will probably complain about this commented out code. Also, it would be good to remove the #if 0...#endif blocks if we're unlikely to need them again.

modules/hdf4_handler/hdfdesc.cc
```diff
@@ -5899,18 +5911,51 @@ bool add_sp_hdf4_trmm_info(D4Group *d4_grp, const string& filename, const D4Attr
...
+        float lon_start = 0;
+        float lat_res = 0;
+        float lon_res = 0;
+        
+        vector<char> grid_header_value_vc(grid_header_value.begin(),grid_header_value.end());
+        HDFCFUtil::parser_trmm_v7_gridheader(grid_header_value_vc,lat_size_gh,lon_size_gh,
+                                         lat_start,lon_start,lat_res,lon_res,false);
+
+        if(lat_size != lat_size_gh || lon_size != lon_size_gh) {
+            err_msg = "Either the latitude size is not the same as the size retrieved from the gridheader";
```
jgallagher59701, line 5928: I cannot find where err_msg is defined - is it a global?

jgallagher59701 (approved):
Some small requests - mostly about err_msg. Don't bother making the changes to the InternalErr() objects in this PR since that is better as a separate PR. There may be more important things for you to work on in the short term.

Thanks.

modules/hdf4_handler/HDFDMRArray_EOS2LL.cc
kyang2014: You are right. I add the exception.

modules/hdf4_handler/HDFStructure.cc
kyang2014: Yes, correct it.

modules/hdf4_handler/hdfdesc.cc
kyang2014: err_msg is passed as the last parameter in this function.
 `add_sp_hdf4_trmm_info(D4Group *d4_grp, const string& filename, const D4Attribute *d4_attr, string &err_msg)`

The error will be caught upstream. This is an area that I tried to do some improvement with the error handling in this PR.

modules/hdf4_handler/HDFDMRArray_EOS2LL.cc
kyang2014: Will do.

PR #927, opened 2024-05-22
Removed the 'fix' that actually broke the FileCache class

jgallagher59701:
Added an explanation.
Fixed the FitRequestHandler.


jgallagher59701:
@ndp-opendap @dh-opendap It would be great if we could review this and get it on master. I put in a comment about why my earlier fix was not needed. That 'fix' had the side effect of breaking on OSX, at least in the unit tests.

ndp-opendap (approved):
:shipit:

PR #928, opened 2024-05-22
Cmr lpdaac patch

jgallagher59701:
Patch for the CMR UMM-G record that has a RelatedURL Type GET DATA record that will break Hyrax because it references some XML file and not a data file.


modules/ngap_module/unit-tests/cmr_json_responses/README
```diff
@@ -0,0 +1,6 @@
+How to build these CMR granule_ur responses
+
+burl -o <granule_name>.json \
+'https://cmr.earthdata.nasa.gov/search/granules.umm_json_v1_4?pretty=true&collection_concept_id=<ccid>>&granule_ur=<gid>'
```
ndp-opendap, line 4: Using the term `<gid>` is very confusing and should be changed. There is a granule "id" (aka "granule concept id", or as used in granule record, `concept-id`) using the granule concept id in place of `<gid>` will cause this request to fail. I suggest replacing `<gid>` with `<GranuleUR>` which is how the expected value is actually labeled in the CMR granule documents

ndp-opendap (approved):
THis is good, but I do have  concern about the CMR granule query example in the new `cmr_json_responses/README` file which I would like you to address.

:shipit:

modules/ngap_module/unit-tests/cmr_json_responses/README
jgallagher59701: OK

PR #935, opened 2024-05-29
Moved all 'unused' code to top level retired directory

ndp-opendap:
Our BES code base is loaded with unused (aka retired) code that makes using CLion hard because search is always identifying this dead code. By placing these unused things in a top level directory, **_retired_**,  we can easily exclude that directory from CLion searches and still keep the things around for later... Or we could just dump them all...


dh-opendap (approved):
A great idea.  All tests look to be passing.  Approved.

jgallagher59701 (approved):
Long overdue! Thanks. Merge.

PR #938, opened 2024-05-31
Changed FileCache so errors in put() are easier to diagnose.

jgallagher59701:
I'm still not certain why I coded it to resturn false when put() finds that the file already is in the cache. I thiink maybe it should be returning true.


kneumiller (commented):
Is this mainly just removing deprecated code?

jgallagher59701:
In terms of LOC changed, yes. However, I added some error messages to improve tracking problems I see in the logs and that is the more important goal. There is no change in behaviour beyond the messages.

kneumiller (approved):

PR #941, opened 2024-06-10
First try at a rocky8 RPM build.

jgallagher59701:
First try at a rocky8 RPM build.

Uses a hack to get the Centos stream 8 hyrax-deps to be used.


kneumiller (approved):
Looks like the rocky8 built, approved on my end.

PR #942, opened 2024-06-10
Hyrax 1295 -draft

kyang2014:
Make dmrpp to support the access of Structure that contains strings(variable or fixed length) as the base type.


modules/dmrpp_module/DMZ.cc
```diff
@@ -1517,6 +1518,229 @@ DMZ::process_missing_data(BaseType *btp, const xml_node &missing_data)
  
 }
 
+bool 
+DMZ::supported_special_structure_type_internal(Constructor *var_ctor) {
+
+    bool ret_value = true;
+    Constructor::Vars_iter vi = var_ctor->var_begin();
+    Constructor::Vars_iter ve = var_ctor->var_end();
+    for (; vi != ve; vi++) {
```
jgallagher59701, line 1527: If you use the `const vector<BaseType*> &variables() const { return d_vars; }` method, these loops can be re-written using the range-based for loop:

for (auto bt: var_ctor->variables()) {
...
}

modules/dmrpp_module/DMZ.cc
```diff
@@ -1517,6 +1518,229 @@ DMZ::process_missing_data(BaseType *btp, const xml_node &missing_data)
...
+    btp->set_read_p(true);
+ 
+}
+
+void DMZ::process_special_structure_data_internal(DmrppStructure * dmrpp_s,  std::vector<u_int8_t> &values , size_t total_value_size, size_t & values_offset){
+
+    Constructor::Vars_iter vi = dmrpp_s->var_begin();
+    Constructor::Vars_iter ve = dmrpp_s->var_end();
+
+    for (; vi != ve; vi++) {
```
jgallagher59701, line 1635: Another loop here.

jgallagher59701 (approved):
This looks really good. I have two suggestions but they are just that.

PR #944, opened 2024-06-13
HYRAX-1434, implement he subset of the miss…

kyang2014:
…ing data


modules/dmrpp_module/DMZ.cc
```diff
@@ -1506,15 +1506,36 @@ DMZ::process_missing_data(BaseType *btp, const xml_node &missing_data)
...
         throw BESInternalError("The dmrpp::missing_data element must be the child of an array variable", __FILE__, __LINE__);
 
+    auto *da = dynamic_cast<DmrppArray *>(btp);
+
     vector<Bytef> result_bytes;
-    auto result_size = (uLongf)(btp->width_ll());
+
+    // We need to obtain the total buffer size to retrieve the whole array. 
+    // We cannot use width_ll() since it will return the number of selected elements.
+    auto result_size = (uLongf)(da->get_size(false) *da->prototype()->width());
```
jgallagher59701, line 1515: What is uLongf? if it is the same as uint64_t, let's use that typename.

jgallagher59701 (approved):
Merge this, but see my question about the type name uLongf. Thanks, James

modules/dmrpp_module/DMZ.cc
kyang2014: uLongf comes from zlib library. Searching on-line, it says it is the same as unsigned long.
Down a few lines, you can see `uncompress(result_bytes.data(), &result_size, decoded.data(), decoded.size());`
The result_size needs to be passed as uLongf pointer. So I do this way.

kyang2014: Will add comments to explain the uLongf in the next pull request.

PR #947, opened 2024-06-21
Hyrax 1297 dev

kyang2014:
Make fileout netcdf handle the DAP structure that contains string.


modules/fileout_netcdf/FONcArrayStructureField.cc
```diff
@@ -299,6 +400,121 @@ void FONcArrayStructureField::obtain_scalar_data(char *data_buf_ptr, BaseType *b
...
+
+        // the start for each of the dimensions will start at 0. We will
+        // bump this up in the while loop below
+        var_start[dim] = 0;
+    }
+
+    // Obtain the compound_buf; this is for gathering the data for individual fields.
+    vector<BaseType*> compound_buf = d_a->get_compound_buf();
+
+    for (unsigned i= 0; i<d_a->length_ll();i++) {
```
jgallagher59701, line 428: This could be replaced with 'for (auto &cb : compound_buf) { ...', as above.

modules/fileout_netcdf/FONcArrayStructureField.cc
```diff
@@ -67,25 +68,113 @@ FONcArrayStructureField::FONcArrayStructureField( BaseType *b, Array* a, bool is
...
+
+}
+
+size_t
+FONcArrayStructureField::obtain_maximum_string_length( ){
+
+    size_t asf_max_str_len = 0;
+
+    // Obtain the compound_buf; this is for gathering the data for individual fields.
+    vector<BaseType*> compound_buf = d_a->get_compound_buf();
```
jgallagher59701, line 130: This could be replaced with 'for (auto &cb : compound_buf) { ...'. Although it's just a suggestion.

jgallagher59701 (approved):
Looks good. One suggestion - but it's just that so if I missed something and that change is not trivial, ignore it. The range-based for loops are nice because they eliminate off-by-one errors in the code.

modules/fileout_netcdf/FONcArrayStructureField.cc
kyang2014: This cannot be replaced by range-get since the index i is used in this for loop.

modules/fileout_netcdf/FONcArrayStructureField.cc
kyang2014: This can. It is too trivial to trigger another travis test. Will do when I clean up the fileout netCDF module.

PR #949, opened 2024-06-26
Hyrax 1228

kyang2014:
draft


modules/dmrpp_module/DmrppRequestHandler.cc
```diff
@@ -217,8 +220,11 @@ DmrppRequestHandler::DmrppRequestHandler(const string &name) :
     msg << prolog << "Contiguous Concurrency Threshold: " << d_contiguous_concurrent_threshold << " bytes." << endl;
     INFO_LOG(msg.str());
 
-    // Is this response a netCDF-4 classic from fileout netCDF
-    // We will check if FONc.ClassicModel is set to true.
+    // Whether the default direct IO feature is disabled. Read the key in.
+    read_key_value(DMRPP_DISABLE_DIRECT_IO,disable_direct_io);
```
jgallagher59701, line 224: The only suggestion I would make, and it's more for your future knowledge, is that there are some static methods in TheBESKeys.cc/h that enable access to BES Keys of various types. For example, you can say:

auto disable_direct_io = TheBESKeys::read_bool_key(DMRPP_DISABLE_DIRECT_IO, false);

That will read the key value if it's set and if it's not set, return 'false.'

But this is similar to the other changes, like the loop, it does not affect correctness, it just cuts down on duplication. If using this will take too much time, or if I have missed some detail, don't bother with this change.

kyang2014: Thanks, James. I am wondering if there is helper function like this.

PR #951, opened 2024-06-27
Hyrax 1443

kyang2014:
Add fileout netCDF tests of handling structures that contain strings via the HDF5 handler


jgallagher59701 (approved):
Thanks. Ready to merge

PR #953, opened 2024-07-02
Hyrax 1450

kyang2014:
Add DAP4 dimension support in the HDF4 handler when directly mapping HDF4 to DMR.


jgallagher59701 (approved):
Thanks for the tests!

Ready to merge.

PR #955, opened 2024-07-09
Hyrax 1454

kyang2014:
Fix the issue caused by an chunked and compressed empty string array. Test the 3 SMAP Level 4 files that encounter the issue. The fileout netCDF can successfully generate the expected netCDF-4 files.


jgallagher59701 (approved):
Approved. Merge! Thanks.

PR #956, opened 2024-07-12
HYRAX-1466, change the default option of get_dmrpp_h4 to make it easi…

kyang2014:
…er for users. The python program will generate a temporary BES configuration file instead of relying on the system's BES configuration. Users can still use -s option to choose to use system's BES configuration.


jgallagher59701 (approved):
Approved. Thanks and ready for a merge.

PR #958, opened 2024-07-18
HYRAX-1229, add the support to disable the direct IO feature based on…

kyang2014:
… the dio="off" attribute in the dmrpp file. Also add an option to generate such a dmrpp file by using the build_dmrpp utility.


jgallagher59701 (approved):
Ready to merge. Thanks!

PR #960, opened 2024-07-25
Add creation date to build_dmrpp metadata

ndp-opendap:
This PR adds the creation time  (as an ISO-8601 formatted string)  in a DAP metadata attribute added the existing build_dmrpp metadata. This will enable us to know exactly when a particular dmr++ was created.

```
        <Attribute name="created" type="String">
            <Value>2024-07-07T12:13:21Z</Value>
        </Attribute>
```
This should also be added to the hdf4 version of build_dmrpp

### DIFF
--- a/modules/dmrpp_module/build_dmrpp_h4/build_dmrpp_util_h4.cc
+++ b/modules/dmrpp_module/build_dmrpp_h4/build_dmrpp_util_h4.cc
@@ -1436,7 +1436,10 @@ void build_dmrpp_from_dmr_file(const string &dmrpp_href_value, const string &dmr
 
 #if 0
     if (add_production_metadata) {
-        inject_version_and_configuration(argc, argv, bes_conf_file_used_to_create_dmr, &dmrpp);
+        // I updated this function call to reflect the changes I made to the build_dmrpp_utl.cc
+        // I see that it is not currently in service but it's clear that something like this
+        // will be needed to establish history/provenance of the dmr++ file. - ndp 07/26/24
+        inject_build_dmrpp_metadata(argc, argv, bes_conf_file_used_to_create_dmr, &dmrpp);
     }
 #endif
 
--- a/modules/dmrpp_module/build_dmrpp_h4/build_dmrpp_util_h4.cc
+++ b/modules/dmrpp_module/build_dmrpp_h4/build_dmrpp_util_h4.cc
@@ -1436,7 +1436,7 @@ void build_dmrpp_from_dmr_file(const string &dmrpp_href_value, const string &dmr
 
 #if 0
     if (add_production_metadata) {
-        // I updated this function call to reflect the changes I made to the build_dmrpp_utl.cc
+        // I updated this function call to reflect the changes I made to the build_dmrpp_util.cc
         // I see that it is not currently in service but it's clear that something like this
         // will be needed to establish history/provenance of the dmr++ file. - ndp 07/26/24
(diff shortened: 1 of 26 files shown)


modules/dmrpp_module/build_dmrpp_util.cc
```diff
@@ -28,6 +28,8 @@
 #include <memory>
 #include <iterator>
 #include <unordered_set>
+#include <iomanip>      // std::put_time()
+#include <time.h>      // std::gmtime_r()
```
jgallagher59701, line 32: I think the time.h include should be 'ctime.' Using ctime guarantees the names are in the std namespace while with time.h, it's only 'likely' they are.

modules/dmrpp_module/build_dmrpp_h4/build_dmrpp_util_h4.cc
```diff
@@ -1436,7 +1436,10 @@ void build_dmrpp_from_dmr_file(const string &dmrpp_href_value, const string &dmr
 
 #if 0
     if (add_production_metadata) {
-        inject_version_and_configuration(argc, argv, bes_conf_file_used_to_create_dmr, &dmrpp);
+        // I updated this function call to reflect the changes I made to the build_dmrpp_utl.cc
```
jgallagher59701, line 1439: build_dmrpp_utl --> build_dmrpp_util

modules/dmrpp_module/build_dmrpp_util.cc
```diff
@@ -1867,22 +1869,48 @@ static string recreate_cmdln_from_args(int argc, char *argv[])
...
+
+    // Convert to system time (time_t)
+    auto time_t_now = std::chrono::system_clock::to_time_t(now);
+
+    // Convert to tm structure (GMT time)
+    struct tm tbuf{};
+    const std::tm* gmt_time = gmtime_r(&time_t_now, &tbuf);
+
+    // Format the time using a stringstream
+    std::stringstream ss;
```
jgallagher59701, line 1889: Oy, stringstream... In C++17 you can write directly to a std::string. Unfortunately, we're at 14 and stringstream is slower by more than order of magnitude than writing the chars to a vector and returning that as a string. I don't have a great solution for that without looking it up.

jgallagher59701 (approved):
Minor edits; no need for more reviewing.

modules/dmrpp_module/build_dmrpp_util.cc
ndp-opendap: I lieu of string stream we could brute force it:
```
    string iso_8601;
    iso_8601.append(to_string(gmt_time->tm_year)).append("-");
    iso_8601.append(to_string(gmt_time->tm_mon)).append("-");
    iso_8601.append(to_string(gmt_time->tm_mday)).append("T");
    iso_8601.append(to_string(gmt_time->tm_hour)).append(":");
    iso_8601.append(to_string(gmt_time->tm_min)).append(":");
    iso_8601.append(to_string(gmt_time->tm_sec)).append("Z");
```

ndp-opendap: Actually, I looked around in the BES and elsewhere we do it using `strftime`:
```
    auto iso8601_fmt = "%Y-%m-%dT%H:%M:%SZ";
    char time_buf[strlen(iso8601_fmt)];
    std::strftime(time_buf, sizeof buf, iso8601_fmt, gmt_time);
```

PR #961, opened 2024-08-05
Own dmrpp

jgallagher59701:
This version of the branch is close  to being ready to merge.


modules/ngap_module/NgapRequestHandler.cc
```diff
@@ -60,8 +60,8 @@ unsigned int NgapRequestHandler::d_dmrpp_mem_cache_purge_items = 20;
 bool NgapRequestHandler::d_use_dmrpp_cache = false;
 MemoryCache<std::string> NgapRequestHandler::d_dmrpp_mem_cache;
 
-unsigned long long NgapRequestHandler::d_dmrpp_file_cache_size_mb = 10'000;    // 10,000 MB ~= 10GB, roughly
-unsigned long long NgapRequestHandler::d_dmrpp_file_cache_purge_size_mb = 2'000;  // 2,000 MB ~= 20MB
+long long NgapRequestHandler::d_dmrpp_file_cache_size_mb = 10'000;    // 10,000 MB ~= 10GB, roughly
+long long NgapRequestHandler::d_dmrpp_file_cache_purge_size_mb = 2'000;  // 2,000 MB ~= 2GB
 string NgapRequestHandler::d_dmrpp_file_cache_dir = "/tmp";     // Please change this!
```
ndp-opendap, line 65: I think that we should continue to partition the way we utilize `/tmp`. In this case maybe `/tmp/hyrax_dmrpp_cache` would be an appropriate name for the cache dir

jgallagher59701: Yes.

PR #963, opened 2024-08-21
Dmrpp bucket option

jgallagher59701:
Tests that toggle OPeNDAP- and DAAC-owned buckets.

These tests don't currently run in Travis, because that would mean 
including NASA credentials in our CI/CD system.


ndp-opendap (approved):
Looks Good.

:shipit:

PR #966, opened 2024-09-09
Hyrax 1467 - update get_dmrpp_h4 python: provide an option for user to generate a BES configuration file

kyang2014:
get_dmrpp_h4 python script is updated to have an option for users to generate a BES configuration file


jgallagher59701 (approved):
Ready for merge. Thanks.

PR #969, opened 2024-09-09
Omno2 direct io

kyang2014:
Fixed the wrong unlimited dimension size in an HDF-EOS5 DMR file. This will make the fileout netCDF module successfully generate a netCDF-4 file for an OMNO2 HDF-EOS5 file via the dmrpp module.


jgallagher59701 (approved):
Ready to merge. 

I think we might need a ticket about adding support for the unlimited dimension. Or, is that a bigger issue than a single ticket?

Thanks.

PR #970, opened 2024-09-10
Xaerdt very large

kyang2014:
Fixed an issue related to the netCDF-4 unlimited dimensions in a XAERT MODIS Aqua aerosol file. This file contains a 2-D variable with both dimensions extendable. The HDF5 handler didn't handle this case.


jgallagher59701 (approved):
Ready to merge. But, see my comment about the excluded code. Thanks.

modules/hdf5_handler/h5dmr.cc
```diff
@@ -1847,6 +1847,56 @@ hsize_t obtain_unlim_pure_dim_size_internal_value(hid_t dset_id, hid_t attr_id,
...
+            }
+        }
+        if (num_unlimited_dims >1)
+            throw InternalErr(__FILE__,__LINE__,"This variable has more than 1 unlimited dimensions. This is not supported.");
+        
+        ret_value = cur_unlimited_dim_size;
+
+        // The following code assumes that the variable has only one dimension, which may not be right. Leave the code here
+        // in case we need to handle multiple unlimited dimensions in a variable in the future.
+#if 0
```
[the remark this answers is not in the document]
kyang2014: remove the code.

modules/hdf5_handler/h5dmr.cc
```diff
@@ -1854,18 +1854,52 @@ hsize_t obtain_unlim_pure_dim_size_internal_value(hid_t dset_id, hid_t attr_id,
...
+        int did_space_num_dims = H5Sget_simple_extent_ndims(did_space);
+        if (did_space_num_dims <0) {
+            H5Aclose(attr_id);
+            H5Tclose(atype_id);
+            H5Dclose(dset_id);
+            H5Dclose(did_ref);
+            string msg = "The number of dimensions must be > 0 for the variable  " + dname;
+            throw InternalErr(__FILE__, __LINE__, msg);
+        }
+        vector<hsize_t> did_dims;
```
jgallagher59701, line 1876: You can shorten the line count by initializing the vectors like 'vector<hsize_t> did_dims(did_space_num_dims);' which might be a bit faster, too.

modules/hdf5_handler/h5dmr.cc
```diff
@@ -1854,18 +1854,52 @@ hsize_t obtain_unlim_pure_dim_size_internal_value(hid_t dset_id, hid_t attr_id,
...
             H5Dclose(did_ref);
-            H5Sclose(did_space);
-            string msg = "Cannot obtain the number of elements for space of the attribute  " + reference_name;
+            string msg = "The dataspace must be a simple HDF5 dataspace for the variable  " + dname;
             throw InternalErr(__FILE__, __LINE__, msg);
         }
 
-        ret_value =(hsize_t)num_ele_dim;
+        int did_space_num_dims = H5Sget_simple_extent_ndims(did_space);
+        if (did_space_num_dims <0) {
```
jgallagher59701, line 1868: Since this same block is repeated several times, I wonder if making a function that closes the various hdf5 ids and throws InternalErr() would be a good idea. Something like:

void close_h5_internal_error(const string &file, const int, const string &msg) {
        H5Aclose(attr_id);
        H5Tclose(atype_id);
        H5Dclose(dset_id);
        H5Dclose(did_ref);
        throw InternalErr(file, line, msg);
}

jgallagher59701 (approved):
This can be merged, but let me know what you think about my suggestions. They might not make sense given the rest of the code - I'm looking at the diff only.

Thanks

modules/hdf5_handler/h5dmr.cc
kyang2014: Good suggestion. Will do.

modules/hdf5_handler/h5dmr.cc
kyang2014: Good suggestion. Will do.

kyang2014: I may not do this way since I need to pass all the IDs to this function. I do find that I need to close the data space in the block. That's another reason to keep the current way,

PR #972, opened 2024-09-10
If this passes, lets merge and test in SIT.


jgallagher59701:


jgallagher59701:
@kneumiller or @sawer297 Can you look at this quick edit and make sure I'm not crazy. I'm looking at ways to test this code, but it's quite hard to Mock functions in C++ (I'm spoiled by what I've recently learned about Python) so there are no unit tests for the new handling of multiple exceptions.

sawer297 (approved):

PR #974, opened 2024-09-19
HYRAX-1503, add the support of Aquarius level 3 version 5.0 for the C…

kyang2014:
…F option. Mainly make sure the lat/lon are added as the previous versions.


modules/hdf5_handler/HDF5GCFProduct.cc
```diff
@@ -595,8 +595,48 @@ bool check_aquarius(hid_t s_root_id,int & s_level) {
...
+                    string msg = "Fail to determine if the HDF5 attribute  ";
+                    msg += string(Aquarius_ATTR2_NAME2);
+                    msg +=" exists ";
+                    H5Gclose(s_root_id);
+                    throw InternalErr(__FILE__, __LINE__, msg);
+                }
+            }
+    
+        }
+        else if (0 == has_aquarius_attr1_2)
```
jgallagher59701, line 630: Could this be moved up and instead of being a noop, use break? If that was done, would the earlier test on line 618 still be needed?

jgallagher59701 (approved):
I have one question, but I may not understand the logic. If you think that idea won't work, then merge this. Otherwise, if you do make that change and the code passes tests, please merge. No second PR is needed.

Thansk!

PR #977, opened 2024-09-25
Hyrax 1518 trim

kyang2014:
Update the check_dmrpp program to include the different cases when the data of a variable can be retrieved through the current data file. These cases include the structure variables, the variables stored in linked blocks, the variables that have storage fillValue provided and the variables of which the data is stored inside the dmrpp file. A testsuite is also implemented.


kneumiller (approved):
Lots of added code, looks good though!

PR #979, opened 2024-09-30
HYRAX-1536, add the support of reading the artifical coordinate value…

kyang2014:
…s in the dmrpp module. Tests are also added.


kneumiller (approved):
If the tests pass I think this is good

PR #980, opened 2024-10-01
Hyrax 1524

kyang2014:
Implement the merge_dmrpp testsuite and also fix the issue documented in HYRAX-1537.


kneumiller (approved):
Looks good. And if the missing variable is being added in the grid_1_2d.h5 without issue then it looks ready to merge.

kyang2014:
> Looks good. And if the missing variable is being added in the grid_1_2d.h5 without issue then it looks ready to merge.

The merged dmrpp file is compared with the baseline file  grid_1_2d.h5.dmrpp.mrg.baseline. If it fails, the test will not get passed. The  baseline file has the offset and length of the missing variables.

PR #983, opened 2024-10-07
Hyrax 1539

kyang2014:
draft


kneumiller (approved):
Looks good. So does this now allow for build_dmrpp to obtain 1-element VLSA strings?

kyang2014:
> Looks good. So does this now allow for build_dmrpp to obtain 1-element VLSA strings?

Yes, we will create another ticket to investigate why the previous way doesn't return the number of elements.

PR #985, opened 2024-10-10
HYRAX-1542, build_dmrpp adds the fake coordinate values to the dmrpp …

kyang2014:
…file. The fake coordinate values are based on the netCDF-4-like pure dimension generated by the EnableCF option.


kneumiller (commented):
Are the fake coordinate values a placeholder? Or do they substitute the values for the dmrpp?

kyang2014:
> Are the fake coordinate values a placeholder? Or do they substitute the values for the dmrpp?

fake coordinate values are saved in the dmrpp file.

kneumiller (approved):
Got it, thank you for your response. I think the commits look good then.

PR #986, opened 2024-10-10
Hyrax 1543

kyang2014:
HDF5 handler EnableCF option: 
Fix the wrong dimension size for the case when netCDF-4's pure dimension is unlimited.


kneumiller (approved):
Good fix!

PR #987, opened 2024-10-18
Zero length dmrpp bug

jgallagher59701:
Zero length dmrpp bug

Sorry for the comments in TempFile.cc/h - they leaked into this PR.


ndp-opendap (approved):
Looks like the thing to do - although you might what to first check that `$OUTPUT_FILE` is not the empty string. That's an allowed state in which case the program output is written to `stdout`.

In the get_dmrpp.in the OUTPUT_FILE env is set to the empty string on line 104:
```
OUTPUT_FILE=
```
And is only utilized if it's set with the `-o` option command line option.

PR #988, opened 2024-10-22
Hyrax 1548

kyang2014:
Add two python scripts to generate dmrpp and the side car files if finding the missing data info,


jgallagher59701 (approved):
This code can be merged, but it might be better for the Python programs to be modified in two ways. The file can man the main part a function called main() and then put 

if __name__ == "__main__":
    main()

at the bottom. That lets us test the functions in the file more easily and use the functions in other files via import statements.

The other thing is to have CLion or PyCharm format the files.  That makes the files more uniform without tedious typing.

jgallagher59701 (approved):
Thanks!

PR #989, opened 2024-10-22
Clion discovery fix

jgallagher59701:
Improved the clion-source-discovery target in the top-level Makefile.am

The target is now robust in the face of test failures. To use, define $prefix and
$PATH in the 'pre-configuration' commands and then use clion-source-discovery 
along with -k -j20 for the 'make arguments and targets' in the Makefile Build.
Set clean to 'clean.'


jgallagher59701:
There's little here for Travis to test since it does not run this target. Use this to enable a build that make CLion's source discovery work for the BES.

ndp-opendap (approved):
:shipit:

PR #990, opened 2024-10-22
The library paths no longer appear in the BES.modules. 

kneumiller:
Also updated the baseline tests to reflect this.


jgallagher59701 (approved):
This can be merged, but two things. Since we have 'using namespace std' at the top of the source, you don't need to include 'std::' in this code. The standard namespace is so ubiquitous that I think it's OK to have those 'using' statements in the code. Also, SonarScan complains when you define two variables on one line (string line, bes_conf;) because it says that's 'less maintainable.' I don't know that I agree, but I figured I'd pass that one. You can see the results of the scan in the GitHub PR. 

Your call about these changes, but I thought I mention them. The sonar scan can be really useful (even though in this instance it's pretty marginal) and not using std:: everywhere can save space when lines get long.

Anyway, please merge this, with our without those edits and close out the ticket. Thanks!

PR #991, opened 2024-10-25
Hyrax 1553

kyang2014:
draft


modules/hdf4_handler/hdfdesc.cc
```diff
@@ -5867,11 +5909,200 @@ bool obtain_eos2_gd_ll_info(const string & fname, const string & grid_name, int3
...
     GDdetach(gridid);
     GDclose(gfid);
 
     return true;
 }
+void add_dummy_grid_cv(D4Group *d4_grp, const eos2_grid_info_t &eg_info) {
+
+    string dummy_proj_cf_name = "eos5_cf_projection";
+    auto dummy_proj_cf_unique = make_unique<HDFEOS2GeoCFProj>(dummy_proj_cf_name, dummy_proj_cf_name);
+    HDFEOS2GeoCFProj *dummy_proj_cf = dummy_proj_cf_unique.release();
```
jgallagher59701, line 5931: Instead of calling release here, you could call get(). That gets the pointer, so you can easily pass it to the various functions but keeps dummy_proj_cf_unique in control of the pointer in case an exception is thrown. 

Alternatively, you could each of the calls like:
         add_var_dap4_attr(dummy_proj_cf_unique.get(), "longitude_of_central_meridian", attr_float64_c, "0.0");

Then down below...

modules/hdf4_handler/hdfdesc.cc
```diff
@@ -5867,11 +5909,200 @@ bool obtain_eos2_gd_ll_info(const string & fname, const string & grid_name, int3
...
+
+        add_var_dap4_attr(dummy_proj_cf, "grid_mapping_name", attr_str_c, "sinusoidal");
+        add_var_dap4_attr(dummy_proj_cf, "longitude_of_central_meridian", attr_float64_c, "0.0");
+        add_var_dap4_attr(dummy_proj_cf, "earth_radius", attr_float64_c, "6371007.181");
+        add_var_dap4_attr(dummy_proj_cf, "_CoordinateAxisTypes", attr_str_c, "GeoX GeoY");
+    } else if (eg_info.projcode == GCTP_PS)
+        add_ps_cf_grid_mapping_attrs(dummy_proj_cf, eg_info);
+    else if (eg_info.projcode == GCTP_LAMAZ)
+        add_lamaz_cf_grid_mapping_attrs(dummy_proj_cf, eg_info);
+    d4_grp->add_var_nocopy(dummy_proj_cf);
```
jgallagher59701, line 5944: ... you can use release() like this: 
        d4_grp->add_var_nocopy(dummy_proj_cf_unique.release());

The benefit is that if anything throws an exception, the unique_ptr cleans up the memory without the calling code doing anything.

jgallagher59701 (approved):
This can be merged, but I have a suggestion about using unique_ptr.

modules/hdf4_handler/hdfdesc.cc
kyang2014: I will merge the code for now. We can discuss this later.

modules/hdf4_handler/hdfdesc.cc
kyang2014: Oh, I see. Got it.

PR #992, opened 2024-10-31
Hyrax 1558

kyang2014:
This pull request consists of two parts:
1) Update/add code in the hdf4_handler, build_dmrpp_h4 utility and dmrpp module so that the dmrpp module can be used to access the HDF-EOS2 Grid with a side car file that contains the grid mapping CF variables and latitude/longitude.
2) Also update/add tests to the hyrax testsuites and NASA sample testsuites.


jgallagher59701 (approved):
Looks good. Ready to merge. Thanks.

PR #994, opened 2024-11-07
DmrppArray now uses uint_64

kneumiller:


kyang2014 (commented):
Kodi, I think you should use int64_t instead of uint64_t. James can explain to you.

modules/dmrpp_module/DmrppArray.cc
```diff
@@ -3088,7 +3088,7 @@ unsigned int DmrppArray::buf2val(void **val){
         memcpy(*val, str_buf.data(), buf_size);
         return buf_size;
     }
-    return Vector::buf2val(val);
+    return Vector::buf2val_ll(val);
```
jgallagher59701, line 3091: Let's change this just a bit (the fix is correct, but to make it easier to read, put the `return Vector::buf2val_ll(val);` in an else block and make the case to unsigned int explicit:

else {
    return (unsigned int)Vector::buf2val_ll(val);
}

Make sure I'm not off my rocker with that and merge if it passes tests on your machine.

Thanks!

jgallagher59701 (approved):

modules/dmrpp_module/DmrppArray.cc
kyang2014: Second  James' suggestion.

kyang2014 (approved):

PR #995, opened 2024-11-08
Hyrax 1530

kyang2014:
draft

### DIFF
--- a/modules/dmrpp_module/build_dmrpp_util.cc
+++ b/modules/dmrpp_module/build_dmrpp_util.cc
@@ -269,7 +269,7 @@ static void set_filter_information(hid_t dataset_id, DmrppCommon *dc, bool disab
  * @param dataset_id
  * @return
  */
-bool
+short
 is_hdf5_fill_value_defined(hid_t dataset_id)
 {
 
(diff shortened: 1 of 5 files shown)


modules/dmrpp_module/build_dmrpp_util.cc
```diff
@@ -392,20 +397,267 @@ get_value_as_string(hid_t h5_type_id, vector<char> &value)
             throw UnsupportedTypeException(msg);
         }
         case H5T_COMPOUND: {
-            string msg(prolog + "UnsupportedTypeException: Your data granule contains an H5T_COMPOUND as user-defined fillValue type. "
-                   "This is not yet supported by the dmr++ creation machinery. ");
             string str_fv(value.begin(),value.end());
-            throw UnsupportedTypeException(msg);
+            break;
```
jgallagher59701, line 401: Here I think the value of str_fv should be returned. I think no value is returned for this case.

jgallagher59701 (approved):
Please check the code in build_dmrpp_util.cc. It might make things clearer if the #if 0 ... #endif blocks are removed.

modules/dmrpp_module/build_dmrpp_util.cc
kyang2014: The compound datatype fill value is calculated elsewhere.  Here we need to use break only.

kyang2014:
> Please check the code in build_dmrpp_util.cc. It might make things clearer if the #if 0 ... #endif blocks are removed.

I removed all the #if 0 #endif blocks  I introduced in the last commit.  The other #if 0 #endif blocks may be added by other people. We can do that when we have a ticket to simple review the code.

PR #996, opened 2024-11-14
Remove CR and LF from error log lines

ndp-opendap:
This change is targeted at the general case error handling in `BESInterface.cc` in which I have added a filter function, `remove_crlf()`, and I apply it in the `log_error()` function in the same file.


jgallagher59701 (approved):
Thanks! :shipit:

PR #997, opened 2024-11-14
Hyrax 1569

kyang2014:
Make dmrpp module access simple compressed compound datatype variables. Also make the handling of the fill value of compound datatype code clearer.


jgallagher59701 (approved):
Looks good. If you have the time, the loop in this PR could be refactored to be a range-based for-loop.

PR #1000, opened 2024-11-15
Remove (#if...) the CurlHandlePool token hack

jgallagher59701:
This was added for the DMR++ POC code. It uses the credential manager to load in an EDL token for certain URLs. It generates a huge number of info messages.


jgallagher59701:
Since this only removes a bit of test code that really should not be in production, I'm merging this without a review.

PR #1001, opened 2024-11-19
Adding the double quote  character to the list of things to remove from the error…

ndp-opendap:
… log messages.


ndp-opendap:
I think this isn't really needed. remove the newline and carriage return characters keeps the log record on a single line. 

Downstream, the beslog2json.py will hoover up that stuff and escape it correctly in the resulting json.

PR #1003, opened 2024-11-20
HYRAX-1453, add the support to reduce the generated dim…

kyang2014:
…ension names by the fileout netCDF module.
Note: This support also handles the case when the same dimension size is shared in one array. In this case, a different dimension name is assigned. For example, int foo[2][2] becomes int foo[dim0=2][dim1=2]. This can make panoply plot such variables.


jgallagher59701 (approved):
Ready to merge. Thanks.

PR #1005, opened 2024-11-22
Fix the info message issue and write tests

jgallagher59701:
There is an annoying and generally useless info message that is coming out of CurlHandlePool - remove it.

As it turns out, this is true, but also there was some copy-pasta code that duplicated code un CurlUtils. The 
CurlHandlePool code was refactored to use the library function and some other clean up was performed. 

Sonar lint has found 61 smells in this and related code, so there will be some more work on this PR.

### DIFF
--- a/modules/dmrpp_module/DmrppArray.cc
+++ b/modules/dmrpp_module/DmrppArray.cc
@@ -1218,7 +1218,7 @@ void DmrppArray::read_chunks_unconstrained()
     stringstream sc_id;
     sc_id << name() << "-" << sc_count++;
     queue<shared_ptr<SuperChunk>> super_chunks;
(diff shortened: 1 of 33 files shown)


modules/dmrpp_module/DmrppArray.cc
```diff
@@ -1413,7 +1413,9 @@ void DmrppArray::read_linked_blocks(){
     
             char **destp = nullptr;
             char *dest_deflate = nullptr;
+#if 0
```
ndp-opendap, line 1416: Might drop this once you know it's not needed.

ndp-opendap (approved):
Looks good so far. Smells ahead...

PR #1006, opened 2024-11-25
Stop watch hacking

jgallagher59701:
I added calls to trigger timing on a small set of methods that will provide 
information about request processing times. This is true for all requests and
for those handled by NgapOwnedContainer. This is specifically to understand 
how we spend time in NGAP (both when working DAAC and 'owned' DMR++
documents.


dispatch/BESInterface.cc
```diff
@@ -411,9 +405,13 @@ int BESInterface::execute_request(const string &from)
 
     // TODO If this is only used for logging, it is not needed since the log has a copy
     //  of the BES PID. jhrg 11/13/17
+#if 0
```
ndp-opendap, line 408: Should it stay or should it go...  (The Clash)

xmlcommand/BESXMLInterface.cc
```diff
@@ -95,24 +95,21 @@ void BESXMLInterface::build_data_request_plan()
         xmlSetGenericErrorFunc((void *) &parseerrors, BESXMLUtils::XMLErrorFunc);
 
         // XML_PARSE_NONET
-        doc = xmlReadMemory(d_xml_document.c_str(), (int)d_xml_document.size(), "" /* base URL */,
+        doc = xmlReadMemory(d_xml_document.c_str(), (int) d_xml_document.size(), "" /* base URL */,
                             nullptr /* encoding */, XML_PARSE_NONET /* xmlParserOption */);
 
         if (doc == nullptr) {
-            string err = "Problem parsing the request xml document:\n";
+            std::string err = "Problem parsing the request xml document:\n";
```
ndp-opendap, line 102: Why add the std:: namesake here? it's omitted elsewhere in this file...

xmlcommand/BESXMLInterface.cc
```diff
@@ -193,7 +191,8 @@ void BESXMLInterface::build_data_request_plan()
                     // will be deleted by the BESXMLInterface destructor when that iterates through
                     // d_xml_cmd_list. jhrg 5/11/22
                     delete current_cmd;
-                } else {
+                }
```
ndp-opendap, line 194: I never know which you want:
```
{
else {
```
or 
```
} else {
```
I feel like this gets flip-flopped a lot...

ndp-opendap (approved):
:shipit:

With questions

dispatch/BESInterface.cc
jgallagher59701: It should go, but the things in data[] are fragile, so I replaced the ostringstream version with something that is faster.

xmlcommand/BESXMLInterface.cc
jgallagher59701: Good question. Should be removed

xmlcommand/BESXMLInterface.cc
jgallagher59701: Yeah, I used automatic formatting and it likes '} else {' but I hate it. I'm learning to live with it because gettign what I want with 'else' messes up other stuff I care about more.

PR #1010, opened 2024-12-03
HYRAX-1568, first version of code to store structure offset and struc…

kyang2014:
…ture size so that the dmrpp module can retrieve the structure data value successfully. Only 1-layer integer or float basetype are supported.


jgallagher59701 (approved):
Thanks!

PR #1012, opened 2024-12-05
Send no debug output to BESLog

ndp-opendap:
Fixing the way the BESDebug  was utilized in BESStopWtach so that 
debugging lines are not inadvertently written to the BESLog


jgallagher59701 (approved):
Approved.

PR #1015, opened 2024-12-10
Hyrax 1529

kyang2014:
Add the handling of a structure array for the chunking storage in the dmrpp module. The following has been added:
1) The array of structure for the chunking storage is supported by two steps:
(1) Obtain the data buffer from all chunks.
(2) using read_array_of_structure() to correctly assigning the individual field values of the data buffer in the structure.
2) The structure alignment is handled by adding a new member bytes_per_element in the DmrppArray class. For a structure, the bytes_per_element is retrieved from the dmrpp file that comes from the original data. For other cases, the bytes_per_element is simply the corresponding libdap datatype size.
3) Many tests regarding the check of structure alignment are added intentionally. 
4) Also include the code to reduce the search time of the linked block attribute  in DMZ.cc.


modules/dmrpp_module/DMZ.cc
```diff
@@ -2020,38 +2025,54 @@ bool DMZ::process_chunks(BaseType *btp, const xml_node &var_node) const
...
+                break;
+        }
+        if (block_count == 1)
+            throw BESInternalError(" The number of linked block is 1, but it should be > 1.", __FILE__, __LINE__);
+        if (block_count >1) {
+            // set using linked block
+            dc(btp)->set_using_linked_block();
+            // reset the count to 0 to process the blocks.
+            block_count = 0;
+            for (auto chunk = chunks.child("dmrpp:block"); chunk; chunk = chunk.next_sibling()) {
```
jgallagher59701, line 2067: Should/could 'chunk' in the above be a reference?

jgallagher59701 (approved):
This is ready to merge. Thanks!

modules/dmrpp_module/DMZ.cc
kyang2014: I did this by following the existing code. So I think it must have a reason not to use the reference. I will try to remember to check the use of reference in my next pull request.

PR #1016, opened 2024-12-13
Hyrax 1528

kyang2014:
Add the support of multi-linked blocks in a chunk to the dmrpp module.  The block index is stored in the dmrpp file and is retrieved correctly.  A test is added. The subset feature will be handled in another ticket.


modules/dmrpp_module/Chunk.h
```diff
@@ -84,8 +84,21 @@ class Chunk {
...
+    // The following two members mean this chunk is actually a linked block.
+    // We don't need to remember chunk position but to remember the linked block index.
     bool linked_block{false};
     unsigned int linked_block_index {0};
+
+    // The following two memebers mean this chunk contains multiple linked blocks.
+    // This is still a chunk but inside this chunk it contains multiple linked blocks.
+    // We will record this chunk multiple times with different multi_linked_block_index_in_dmrpp_file.
+    // This is for generating the dmrpp file. 
+    bool multi_linked_blocks{false};
```
jgallagher59701, line 97: Let's keep the "{false}" initializer for now, but I wanted to point out that it's a an awkward way to initialize scalar values. We could be using "multi_linked_blocks= false". I think the reason the code from about the time Chunk was first written uses the "{}" syntax is because using "=" with objects means using operator=() which could be specially defined for the class. But for scalar values like '0' and 'false', that's not the case.

jgallagher59701 (approved):
Some comments, but it looks ready to go.

modules/dmrpp_module/Chunk.h
```diff
@@ -386,6 +464,7 @@ class Chunk {
         d_data_url = std::move(data_url);
     }
 
+    virtual bool is_read_buffer_is_mine() { return d_read_buffer_is_mine; }
```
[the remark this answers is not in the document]
kyang2014: change it according to your suggestion.

modules/dmrpp_module/Chunk.h
kyang2014: I change  a bunch of linked block members to the normal syntax.

PR #1017, opened 2024-12-19
Remove BESLog  i/o operator implementations

ndp-opendap:
This change is something James and I have been discussing. A lot of files are affected but the substantive changes are in the `BESLog` class. removing the I/o operators from BESLog caused wide spread changes in the way the BESLog (and its macros) are used.

What this change means is that the BESLog macros will no longer support the use of i/o operators:
```
ERROR_LOG("Ouch, when I tried to read from " << filename << " I got nothing.");
```

Now this will need to be expressed as:
```
ERROR_LOG("Ouch, when I tried to read from " + filename + " I got nothing.");
```

And notably this means that numeric values will have to be wrapped with calls to `std::to_string()`:
```
ERROR_LOG("Ouch, when I tried to parse " + filename + " I encountered an error on line: " + std::to_string(line_num));
```

**NOTE**: The updated `BESLog` class now performs a check for a newline at the end of the log message and if one is not present it will add one. 

So no more of this:
```
ERROR_LOG(error_msg + endl);
INFO_LOG("some thing interesting...\n");
INFO_LOG("some thing interesting..." + "\n");
```

 @kyang2014 - There were several places in the hdf5_handler where the BESLog was being used in a way that caused the resulting log records (aka lines) to be incorrectly formatted/constructed. I replaced these with calls to the BESLog macros. Please have a look at these changes!


kyang2014:
I reviewed the parts in the HDF4/5 handlers. They look fine.

dispatch/BESLog.h
```diff
@@ -108,12 +115,10 @@ class BESLog: public BESObj {
 private:
     static BESLog * d_instance;
 
-    int d_flushed;
     std::ofstream * d_file_buffer;
```
jgallagher59701, line 118: Initialize this here to nullptr (std::ofstream *d_file_buffer = nullptr;) and remove from the ctor if needed.

dispatch/BESLog.h
```diff
@@ -108,12 +115,10 @@ class BESLog: public BESObj {
...
     std::string d_file_name;
-
-    // Flag to indicate the object is not routing data to its associated stream
-    int d_suspended;
+    std::string d_instance_id;
+    std::string d_pid;
 
     // Flag to indicate whether to log verbose messages
     bool d_verbose;
```
jgallagher59701, line 124: ...same here ... = false; or whatever.

dispatch/BESLog.h
```diff
@@ -122,56 +127,40 @@ class BESLog: public BESObj {
 
     bool d_use_unix_time; // Use the UNIX time value as the log time.
```
jgallagher59701, line 128: ... and here.

dispatch/BESLog.cc
```diff
@@ -76,10 +75,36 @@ const string BESLog::mark = string("|&|");
  * @see BESKeys
  */
 BESLog::BESLog() :
-    d_flushed(1), d_file_buffer(0), d_suspended(0), d_verbose(false), d_use_local_time(false), d_use_unix_time(false)
+    d_file_buffer(nullptr),
```
jgallagher59701, line 78: I put this as a comment in the header because I looked at it first, but these should all be initialized in the class header.

dispatch/BESLog.cc
```diff
@@ -76,10 +75,36 @@ const string BESLog::mark = string("|&|");
...
+    d_instance_id("-"),
+    d_pid("-"),
+    d_verbose(false),
+    d_use_local_time(false),
+    d_use_unix_time(false)
 {
-    d_suspended = 0;
+    // The process ID doesn't change (does it??)
+    d_pid = to_string(getpid());
```
jgallagher59701, line 86: Yay!

dispatch/BESLog.cc
```diff
@@ -154,27 +180,29 @@ BESLog::~BESLog()
 {
     d_file_buffer->close();
     delete d_file_buffer;
-    d_file_buffer = 0;
+    d_file_buffer = nullptr;
```
jgallagher59701, line 183: You can leave this in, but in a dtor, you don't need to set these to null since there is no way (unless there's a glaring compiler error) for them to be used again.

jgallagher59701 (approved):

PR #1019, opened 2024-12-23
test(bes,hdf5_handler): update NASA sample download URLs

hyoklee:
The HDF Group FTP/HTTP server for sample files will be retired soon.
This PR will update the URLs that point to the HDF Group FTP/HTTP server.
Some of them point to NASA DAACs directly, which will require NASA Earthdata Login.


kyang2014:
Joe, Could you create a branch instead of a pull request for just updating the path in the shell scripts? I can merge that branch with the similar work I am doing.

PR #1020, opened 2024-12-23
HYRAX-1593, add code to handle the access of subset data for the mult…

kyang2014:
…i-linked block chunk case.


modules/dmrpp_module/DmrppArray.cc
```diff
@@ -1575,13 +1575,95 @@ void DmrppArray::read_chunks_with_linked_blocks() {
...
     }
+    set_read_p(true);
 
 }
 
 void DmrppArray::read_chunks_with_linked_blocks_constrained() {
 
-    // Will handle this in a separate ticket.
+    reserve_value_capacity_ll(get_size(true));
+    char *dest_buf = this->get_buf();
```
jgallagher59701, line 1585: I was going to say that you don't need to use 'this->' (which I know you know) but I think it's good here because this is complicated and helps make it clear that the buffer and shape are for the chunk.

jgallagher59701 (approved):
Ready for Merge! Thanks very much!

PR #1021, opened 2024-12-24
Hyrax 1597

kyang2014:
migrate  the NASA testing files to the opendap server machine.


kyang2014:
Just add a few shell scripts to obtain files. No need to review.

PR #1022, opened 2024-12-24
HYRAX-1564 fixed the bug that causes the segmentation fault for the H…

kyang2014:
…DF4 handler to access an MOD021KM file. The Vdata field name needs to apply the CF naming rule.


modules/hdf4_handler/HDFDMRArray_VD.cc
```diff
@@ -431,20 +431,24 @@ HDFDMRArray_VD::read_multi_fields_vdata(int32 vdata_id,const vector<int>&offset,
     vector<string> field_names;
     char sep =',';
     HDFCFUtil::Split(field_name_list,sep,field_names);
+    for (auto &fn:field_names)
```
jgallagher59701, line 434: Could 'fn' be const?

jgallagher59701 (approved):
Approved. I have one tiny question, but just take a look and merge as you see fit.
Thanks!

modules/hdf4_handler/HDFDMRArray_VD.cc
kyang2014: No, fn is changed.

PR #1024, opened 2024-12-26
Added log_ingo and reqID to timing in the framework

jgallagher59701:
Morfed the BES_COMMAND_TIMING() macro into two: one for the framework and a new macro for modules.


ndp-opendap (approved):
If this makes you happy then :shipit:

PR #1025, opened 2024-12-30
Making logging reflect the pid of the child processes

ndp-opendap:


jgallagher59701 (approved):
Looks good. Merge if it passes tests. 

Which begs the question, are there tests?

ndp-opendap:
No, BESLog does not have any unit tests

PR #1027, opened 2024-12-31
Hyrax 1497

kyang2014:
For the HDF5 handler, use the UTF-8 encoding for DMR when the escapeUTF8Attr key is set to false. This check-in also provides a way for other handlers to use UTF-8 encoding in the DMR representation.


jgallagher59701 (approved):
Approved.

What do you think about making "UTF-8" the value of a constant (#define) ? I think that would probably be best in the XMLWriter.h file back in libdap - we might just keep that in mind for the future.

kyang2014:
> What do you think about making "UTF-8" the value of a constant (#define) ? I think that would probably be best in the XMLWriter.h file back in libdap - we might just keep that in mind for the future.

I think this is the right way. I put this suggestion in a future ticket.

PR #1028, opened 2024-12-31
Test the proposed fix for FileCache::put(const string&, PutItem&)

jgallagher59701:
I think an error in this method was causing the cache_info file corruption that then lead to the cache purge operations failing.


jgallagher59701:
This fixes the problem that I can replicate in a unit test - flock(2) does not work when two processes hold the same open descriptor _and_ the PutItem dtor was failing to lock the cache when it updated the cache_info file. With the fixes in place, the tests for this (the put code) work. I used 1,000 put operations in two processes and without the fix, the test failed for over 100 consecutive trials.

dispatch/FileCache.h
```diff
@@ -60,6 +61,12 @@
 // an LRU cache. jhrg 11/03/23
 #define FORCE_ACCESS_TIME_UPDATE 1
 
+// I tried using fcntl(2) because it's supposedly better than flock(2) when using
+// NFS, but I think that is really for NFS 3. For NFS 4, flock(2) is supposed to be
+// fully supported. It's much easier and it supports multi threading. The fcntl(2)
+// code here is not yet passing all the tests!
+#define USE_FCNTL 0
```
ndp-opendap, line 68: So if this works with flock() will this be something that goes away??

ndp-opendap (approved):
Questions, but no changes...

:shipit:

modules/ngap_module/NgapOwnedContainer.cc
```diff
@@ -241,6 +242,16 @@ bool NgapOwnedContainer::put_item_in_dmrpp_cache(const std::string &dmrpp_string
         return false;
     }
 
+#if 0
```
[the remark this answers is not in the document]
jgallagher59701: Well, that's a good question. If not, I'll drop it, but I want to see how things shake out in/on SIT before I drop it.

dispatch/FileCache.h
jgallagher59701: Yes. I thought I got that. I just made another push that tightens up the unit tests and I think I got this then as well.

jgallagher59701: Arrrgh, I did not. but I got it now.

PR #1029, opened 2025-01-02
Added a test for the FileCache PutItem bug

jgallagher59701:
I marked this as expected to fail so that this branch can be merged to master.


ndp-opendap (commented):
:shipit:
Although I am curious: Does this trigger the issue on a single host?

ndp-opendap (approved):

jgallagher59701:
> :shipit: Although I am curious: Does this trigger the issue on a single host?

Yes, it does. It actually can show both bugs - the cache_info file can be larger than 8 bytes (16, 24, ...) and can contain the wrong size information. It happens every time I run the test, even though it's a stochastic process.

PR #1030, opened 2025-01-02
HYRAX-1265, make the dmrpp read the default fill value of a compound …

kyang2014:
…datatype.
Note: In the process of working on this test, I fixed a bug that generates the wrong values when the structure is not well-aligned and the chunk is filled with fill values.


modules/dmrpp_module/DmrppCommon.cc
```diff
@@ -443,7 +443,18 @@ unsigned long DmrppCommon::add_chunk(
...
-
+unsigned long DmrppCommon::add_chunk(
+        const string &byte_order,
+        const string &fill_value,
+        libdap::Type fv_type,
+        unsigned long long chunk_size,
+        const vector<unsigned long long> &position_in_array,
+        unsigned int s_size)
+{
+    shared_ptr<Chunk> chunk(new Chunk(byte_order, fill_value, fv_type, chunk_size, position_in_array,s_size));
```
jgallagher59701, line 454: For this is better to use make_shared() - the resulting shared pointers are less costly to use - the reason is that make_shared() builds and object that can be used with one less indirection step.

jgallagher59701 (approved):
Ready to merge, but see my comment about improving the efficiency of shared pointers.

modules/dmrpp_module/DmrppCommon.cc
kyang2014: I am following the coding style inside existing code inside DmrppCommon.cc. I think we may need to create another ticket to just handle the shared pointers used in the dmrpp module.

PR #1037, opened 2025-01-15
Hyrax 1622 test deploy

kyang2014:
Add a python script get_dmrpp_h5 that is used by gen_dmrpp_sidecar. Without this python script, the gen_dmrpp_sidecar won't work with HDF5 files on docker.


kyang2014:
Since no really C++ source code is involved and a new release is coming, I just merge the pull request after BES passes travis.

PR #1039, opened 2025-01-17
Hyrax 1634

kyang2014:
The LPDAAC MCDGF product doesn't follow HDF-EOS2 conventions for the grid dimensions. Therefore we need to handle this product in a totally different way. This check-in also provides a template to handle the similar products in the future.


jgallagher59701 (approved):
Ready to merge. Thanks.

PR #1041, opened 2025-01-24
Fixes to error reporting

jgallagher59701:
Some of our HTTP Error messages are repeated twice, making the simple counts of errors hard to match to other counts in other places. It also makes it harder to read the logs (which are plenty hard to read without this). 

Also fixed are some things reported as errors that should be info records.


ndp-opendap (approved):
:shipit:

PR #1042, opened 2025-01-28
HYRAX-1579, first version to optimize non-contiguous chunk reading

kyang2014:
This check-in enables a superchunk to contain non-contiguous chunks by using a buffer to grab data that has non-contiguous chunks. Previously each non-contiguous chunk will be one superchunk, thus makes many potential small reads. 
This check-in only handle the whole array transfer.


jgallagher59701 (approved):
Ready to merge. Thanks.

PR #1044, opened 2025-01-30
No multiline

ndp-opendap:
There are places in the `ncml_handler/DDSLoader` that were creating multiline log messages. 
This PR fixes that and
- Adds the method `BESError::error_name()` for use in messages etc.
- Changes `BESInterface::log_error()` to use the `BESError::error_name()` method


dh-opendap (approved):
The changes appear correct to me.

PR #1047, opened 2025-02-03
No children is OK...

ndp-opendap:
Or not. We will see...


modules/cmr_module/CmrApi.cc
```diff
@@ -116,27 +116,27 @@ const nlohmann::json &CmrApi::get_children(const nlohmann::json &jobj) const
     BESDEBUG(MODULE, prolog << json.probe_json(jobj) << endl);
     bool result = jobj.is_null();
     if(result){
-        stringstream msg;
-        msg <<  "ERROR: Json document is NULL: " << endl << jobj.dump(2) << endl;
-        BESDEBUG(MODULE, prolog <<  msg.str() << endl);
-        throw CmrInternalError(msg.str(), __FILE__, __LINE__);
+        string msg;
+        msg + "ERROR: Json document is NULL. json: " + jobj.dump();
```
jgallagher59701, line 120: In these error messages, you're using the '+' operator as an assignment operator. Does that work? It's hard to look up and I cannot find a reference for it. You could rewrite the above as:

msg += "ERROR: ..." + jobj.dump();

Or, as:

string msg{"ERROR: ..."};
msg += jobj.dump;

jgallagher59701 (commented):
Let me know about using the + instead of +=

What do the tests show?

modules/cmr_module/CmrApi.cc
ndp-opendap: WTF - I had all those changes already...

jgallagher59701 (approved):
OK, ready to merge. Thanks.

jgallagher59701 (approved):
:shipit:

PR #1048, opened 2025-02-03
Hyrax 1488

kyang2014:
Map HDF5 Enum to DAP4 Enum. It supports both scalar and array. Tests are also added.


jgallagher59701 (approved):
Ready to merge. Thansk!

PR #1052, opened 2025-02-12
HYRAX-1357, Use fill value to fill in the dataset if an SDS is empty.…

kyang2014:
… We also add the _FillValue attribute to the empty SDS if the empty SDS doesn't have _FillValue attribute.


modules/hdf4_handler/hdfdesc.cc
```diff
@@ -5816,11 +5824,21 @@ void map_sds_var_dap4_attrs(HDFDMRArray_SDS *ar, int32 sds_id, int32 obj_ref, in
             throw InternalErr(__FILE__,__LINE__, "SDreadattr failed ");
             
         string dap4_attrname (attr_name);
+        if (emptySDS == 1) {
```
jgallagher59701, line 5827: I see in sonarscan that it thinks these three if statements can be combined into one. Is that true? If so, it might make this easier to understand. A minor issue, so change this at your discretion.

jgallagher59701 (approved):
One comment, but this can be merged. Thanks.

modules/hdf4_handler/hdfdesc.cc
kyang2014: I noticed this. Actually I deliberately do this way to make the logic clear since the "nested ifs" follow the logic in this case.

PR #1053, opened 2025-02-13
Refactoring in advance of work on ownership

jgallagher59701:
Refactoring in advance of work on ownership

Next step: credential injection

### DIFF
--- a/dispatch/BESStreamResponseHandler.h
+++ b/dispatch/BESStreamResponseHandler.h
@@ -22,7 +22,7 @@
 //
 // You can contact University Corporation for Atmospheric Research at
 // 3080 Center Green Drive, Boulder, CO 80301
- 
+
 // (c) COPYRIGHT University Corporation for Atmospheric Research 2004-2005
 // Please read the full copyright statement in the file COPYRIGHT_UCAR.
 //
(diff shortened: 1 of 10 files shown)


dispatch/BESStreamResponseHandler.h
```diff
@@ -22,7 +22,7 @@
 //
 // You can contact University Corporation for Atmospheric Research at
 // 3080 Center Green Drive, Boulder, CO 80301
- 
+
```
ndp-opendap, line 25: Why is this blank line not a comment like the rest of the blank lines in the copyright preamble?

modules/dmrpp_module/vlsa_util.cc
```diff
@@ -340,7 +330,8 @@ void read(const pugi::xml_node &vlsa_element, std::vector<std::string> &entries)
         }
         BESDEBUG(VLSA, prolog << "value: '" << value << "' count: " << count << "\n");
         for(uint64_t i=0; i<count ;i++){
-            entries.emplace_back(value);
+            // Using emplace_back(value) here causes an error (why?) with ASAN on. jhrg 2/11/25
```
ndp-opendap, line 333: the semantics of emplace_back are confusing and I think there are issues with using it that we should better understand!!

ndp-opendap (approved):
Looks fine. Glad to see the streamsize usage.

A couple of minor comments
:shipit:

modules/dmrpp_module/vlsa_util.cc
jgallagher59701: True. I'm not completely sure why ASAN flags this particular line. The issue is a (detected) write beyond the end of the container, but then ASAN says, (paraphrasing) "if you want to ignore these errors, define this ENV variable." So is this important or not? In this case, I switched to push_back() because the other uses of emplace_back() work fine - this was the only one that produced this warning/error from ASAN.

PR #1056, opened 2025-02-17
HYRAX-1666, make gen_dmrpp_sidecar handles suffix other than .h5 etc.

kyang2014:


modules/dmrpp_module/get_dmrpp/get_dmrpp_h5
```diff
@@ -160,7 +160,8 @@ BES.module.cmd=
 BES.module.h5=
 BES.Catalog.catalog.RootDirectory=
 BES.Data.RootDirectory=/dev/null
-BES.Catalog.catalog.TypeMatch=h5:.*\.(h5|he5|HDF5)(\.bz2|\.gz|\.Z)?$;
+BES.Catalog.catalog.TypeMatch=dmrpp:.*\.(dmrpp)$;
+BES.Catalog.catalog.TypeMatch+=h5:.*(\.bz2|\.gz|\.Z)?$;
```
jgallagher59701, line 164: This regex will assign any file that does not end in '.dmrpp' to the hdf5 handler. I can see this is a special case since this configuration is going to be used  by the get_dmrpp_h5 script. If that's the case, we can tidy up the regexes some, to make them simpler:
For `BES.Catalog.catalog.TypeMatch=dmrpp:.*\.(dmrpp)$; ` the regex could be `.*\.dmrpp$` since the `()`'s are not needed.
For `BES.Catalog.catalog.TypeMatch+=h5:.*(\.bz2|\.gz|\.Z)?$;` we can drop the optional compression suffices so the regex can be `.*$`

Here's how it can go in the script (but test this, since these regexes can be tricky and I might have made a mistake ;-)

```
BES.Catalog.catalog.TypeMatch = dmrpp:.*\.dmrpp$;
BES.Catalog.catalog.TypeMatch += h5:.*$;
```

jgallagher59701 (approved):
OK, but I think we can simplify the regexes. Also, I think we might want to add a comment that these regexes are assigning anything that does not end in `.dmrpp` to the HDF5 handler because this is the get_dmrpp_h5 program. 

But, of course, all of this is optional.

PR #1058, opened 2025-02-19
Hyrax 1655

kyang2014:


modules/dmrpp_module/build_dmrpp_util.cc
```diff
@@ -1514,6 +1529,171 @@ void process_string_in_structure(hid_t dataset, hid_t type_id, BaseType *btp) {
...
 }
 
+//Note: the error handling part may be improved in the future.
+bool handle_vlen_float_int_internal(hid_t dset_id, BaseType *btp) {
+
+    hid_t vlen_type = H5Dget_type(dset_id);
+    hid_t vlen_basetype = H5Tget_super(vlen_type);
+    if (H5Tget_class(vlen_basetype) != H5T_INTEGER && H5Tget_class(vlen_basetype) != H5T_FLOAT) {
+        H5Dclose(dset_id);
+        throw InternalErr(__FILE__, __LINE__,"Only support float or intger variable-length datatype.");
```
jgallagher59701, line 1539: It would be better to use BESInternalError here instead of the libdap InternalErr.

jgallagher59701 (approved):
Can be merged. I made one suggestion about InternalErr, but it's not a big deal and maybe what your comment referred to regarding fixing the error handling.

modules/dmrpp_module/build_dmrpp_util.cc
kyang2014: Got it. Will check all InternalErr and change them to BESInternalError in a separate ticket.

PR #1062, opened 2025-02-21
Refactor; WIP

jgallagher59701:
Refactor HTTP code, removing shared_ptr<http::url> use, etc.

This refactor also switches the AllowedHosts class to the new,
and reliable, singleton pattern (aka the Meyers singleton).

### DIFF
--- a/dispatch/BESRequestHandlerList.cc
+++ b/dispatch/BESRequestHandlerList.cc
@@ -149,14 +149,6 @@ string BESRequestHandlerList::get_handler_names()
 
     string ret;
     bool first_name = true;
-#if 0
-    BESRequestHandlerList::Handler_citer i = _handler_list.begin();
-    for (; i != _handler_list.end(); i++) {
-        if (!first_name) ret += ", ";
-        ret += (*i).first;
-        first_name = false;
-    }
-#endif
     for (auto const &handler_pair: _handler_list) {
         if (!first_name) ret += ", ";
         ret += handler_pair.first;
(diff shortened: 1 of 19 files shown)


jgallagher59701:
@ndp-opendap Can you review this? I think it's ready and if it grows more it will be unwieldy. This is a pure refactor of the code that I'm working with for the DMR++ 'ownership' objective in NASA-land.

dispatch/BESRequestHandlerList.cc
```diff
@@ -107,12 +103,11 @@ BESRequestHandlerList::find_handler(const string &handler_name)
 {
     std::lock_guard<std::recursive_mutex> lock_me(d_cache_lock_mutex);
 
-    BESRequestHandlerList::Handler_citer i;
-    i = _handler_list.find(handler_name);
+    auto i = _handler_list.find(handler_name);
     if (i != _handler_list.end()) {
         return (*i).second;
```
ndp-opendap, line 108: Based on other changes I think this could be:
```
return i->second;
```

http/AllowedHosts.cc
```diff
@@ -180,52 +143,40 @@ bool AllowedHosts::is_allowed(shared_ptr<http::url> candidate_url, std::string &
...
     else {
-        stringstream ss;
-        ss << "The candidate_url utilizes an unsupported protocol '" << candidate_url->protocol() << "'" ;
-        BESDEBUG(MODULE, prolog << ss.str() << endl);
-        throw BESInternalError(ss.str(),__FILE__,__LINE__);
+        why_not = "The URL utilizes an unsupported protocol:" + candidate_url.protocol();
+        isAllowed = false;
     }
-    BESDEBUG(MODULE, prolog << "END Access Allowed: " << (isAllowed ? "true " : "false ") << endl);
+
```
ndp-opendap, line 161: I gotta say, I'm not keen on removing all these BESDEBUG calls from AllowedHosts (They cost nothing in the production version). I added them in here because this was a beast to debug...

ndp-opendap (approved):
This got pretty big.

As I said in the comments  - I'm not so keen on removing all these BESDEBUG statements from AllowedHosts. Other wise it seems fine, basically  80% c++ modernization

dispatch/BESRequestHandlerList.cc
jgallagher59701: Yup. Changed

http/AllowedHosts.cc
jgallagher59701: Added most of them back in.

PR #1063, opened 2025-02-24
Added request_id to all BES log records.

ndp-opendap:
This PR adds the request_id (as transmitted by the OLFS) to every BES log record. If the id is not available a "-" character will be used.

There is a corresponding PR in hyrax-docker that updates the beslog2json.py for this change.

Do we need to notify NASA metrics? I did and we are good to go.

This is PR got a wee bit big. For brevity - I changed a REQUEST_ID macro to the more accurate REQUEST_ID_KEY in BesNames.h and that had a pretty big ripple.

The meat of this PR are changes to:
* **_BESStopwatch_** - Altered so that we get request_id correctly into the timing logs.  (ripples)
* **_BESLog_** - Will now add the request_id to all log records
* **_beslog2json.py_** -  Modified so that the request_id is found correctly in the log records and the JSON encoding continues to be viable.


ndp-opendap:
This PR is ready for review.

dispatch/BESStopWatch.h
```diff
@@ -68,7 +72,7 @@ if (BESISDEBUG((module)) || BESISDEBUG(TIMING_LOG_KEY) || BESLog::TheLog()->is_v
 #define BES_MODULE_TIMING(message) BESStopWatch commandTimer; \
     commandTimer.start(string("Module timing: ") + (message))
 #define BES_COMMAND_TIMING(message, DHI) BESStopWatch commandTimer; \
-    commandTimer.start(string("Command timing: ") + (message) + (d_dhi_ptr->data[REQUEST_ID] + " " + d_dhi_ptr->data[LOG_INFO]))
+    commandTimer.start(string("Command timing: ") + (message) + (d_dhi_ptr->data[LOG_INFO]), d_dhi_ptr)
```
jgallagher59701, line 75: Not sure if these changes have an impact on my earlier comments about using these macros...

modules/dmrpp_module/DmrppArray.cc
```diff
@@ -366,7 +366,7 @@ bool start_super_chunk_unconstrained_transfer_thread_dio(list<std::future<bool>>
 void read_super_chunks_unconstrained_concurrent(queue<shared_ptr<SuperChunk>> &super_chunks, DmrppArray *array)
 {
     BESStopWatch sw;
-    if (BESDebug::IsSet(TIMING_LOG_KEY)) sw.start(prolog + " name: "+array->name(), "");
+    if (BESDebug::IsSet(TIMING_LOG_KEY)) sw.start(prolog + " name: "+array->name());
```
jgallagher59701, line 369: Maybe use the macros here, too?

jgallagher59701 (approved):
This can be merged but if we can work in use of the timing macros, that would be good.

PR #1066, opened 2025-03-10
Http head

jgallagher59701:
Adding use of the AWS_SDK code into NgapOwnedContainer.


jgallagher59701:
@ndp-opendap Take a look at this when you have some time. It's a change to the NGAP module that I'm just starting on. Before I settle on an API for the AWS_SDK module, I want to see how the code will play out in actual use. Since this will (at least initially) use the credential manager and since that is all driven by HTTP/S urls and since the AWS SDK uses bucket names with objects and regions, some work needs to be done in the CredentialsManager and AccessCredentials classes.

PR #1067, opened 2025-03-12
Rocky8 fix

jgallagher59701:
Switched from Centos8 to Rocky8 for the 'el8' RPM packages.

This change was made in the .travis.yml file. As a second round, we could
make the actual string used a parameter set in a single place.


ndp-opendap (approved):
:shipit:

PR #1071, opened 2025-03-17
Using request id and uuid attributes from the xml request. 

ndp-opendap:
The set up is that the  OLFS is passing a UUID for every request so that if clients/users submit multiple requests with identical request ids we can disambiguate that in our application logs while still passing the unaltered request id string to down stream services.

I also added code to support the event where the client does not supply a UUID, in which case the BES will make one and use it.

**_So Many Files You Say?_**

Since each UUID is unique and the various BES test commands in the project did not specify a UUID in their request elements, many tests failed because of the code creating new UUIDs for every test request.  I edited many (not all) of the BES command files and by added a fixed UUID values to the request elements and to the affected baselinesI was able to get the tests to pass.


jgallagher59701:
Instead of these BAD_CAST casts, you could probably cast to (const xmlChar *) using a C-style cast. That would at least preserve constness. BAD_CAST is "#define BAD_CAST (xmlChar *)" which is bad. If you want to use a C++-style cast, you can say static_cast<const xmlChar *>(_thing_).

https://gnome.pages.gitlab.gnome.org/libxml2/devhelp/libxml2-xmlwriter.html#xmlTextWriterWriteAttribute

jgallagher59701 (approved):
There are many changed files. I think you should remove the BAD_CAST stuff if you can. I put suggestions in a comment.

PR #1072, opened 2025-03-18
Fixes typo in beslog2json.py

ndp-opendap:
Fixes typo in `beslog2json.py` that was assigning the user id to the `olfs-start-time` record


jgallagher59701 (approved):
:shipit:!

PR #1073, opened 2025-03-25
HYRAX-1695, add tests for the variable-length float/int hdf5/dmrpp th…

kyang2014:
…at has dimension names, also add a test that have the same compound data type array under different groups. This test reveals a bug inside the netCDF-C client that handles the DAP4 response.


jgallagher59701 (approved):
Thanks

PR #1074, opened 2025-03-26
Hyrax 1667

kyang2014:


modules/hdf5_handler/HDF5Array.h
```diff
@@ -127,6 +122,14 @@ class HDF5Array:public libdap::Array {
     void m_array_of_region_reference_hyperslab_selection(hid_t space_id, int ndim, const std::string &varname,
                                                              std::vector<std::string> &v_str,int64_t i) const;
     friend class HDF5Structure;
+
+  protected:
+    
+    // Parse constraint expression and make HDF5 coordinate point location.
+    // return number of elements to read. 
+   int64_t format_constraint(int64_t *cor, int64_t *step, int64_t *edg);
```
jgallagher59701, line 130: If this is going to be overloaded by a child class, then it should be virtual.

modules/hdf5_handler/HDF5VlenAtomicArray.cc
```diff
@@ -0,0 +1,262 @@
...
+            default:
+                throw InternalErr(__FILE__, __LINE__, "Vector::val2buf: bad type");
+        }
+    }
+    if(H5Dvlen_reclaim(vlen_memtype, memspace, H5P_DEFAULT, (void*)(vlen_data.data())) <0) {
+        H5Dclose(dset_id);
+        H5Fclose(file_id);
+        throw InternalErr(__FILE__, __LINE__, "H5Dvlen_reclaim failed.");
+    }
+    H5Sclose(vlen_space);
```
jgallagher59701, line 251: It's hard to tell from the diff, but are all these taken care of when exceptions are thrown?

modules/hdf5_handler/HDF5VlenAtomicArray.cc
```diff
@@ -0,0 +1,262 @@
...
+bool HDF5VlenAtomicArray::read()
+{
+    BESDEBUG("h5","Coming to HDF5VlenAtomicArray read "<<endl);
+
+    read_vlen_internal(is_vlen_index);
+    
+    return true;
+}
+
+void HDF5VlenAtomicArray::read_vlen_internal(bool vlen_index) {
```
jgallagher59701, line 63: If it's complicated to track which of the HDF5 objects (e.g., vlen_memtype) need to be closed when exceptions are thrown, maybe breaking this method up into parts would help?

jgallagher59701 (approved):
I think this might need some work in the HDF5VlenAtomicArray::read_vlen_internal(bool vlen_index) method if there are HDF5 objects that need to be closed when exceptions are thrown. Also, I'm not sure I understand why the int64_t format_constraint(int64_t *cor, int64_t *step, int64_t *edg) in HDF5Array was declared protected.

modules/hdf5_handler/HDF5VlenAtomicArray.cc
kyang2014: Actually no. In theory, they should. I agree they should be handled. Will do.

modules/hdf5_handler/HDF5VlenAtomicArray.cc
kyang2014: Will see. But mostly I will just add those H5?close() in the exception blocks.

modules/hdf5_handler/HDF5Array.h
kyang2014: This method is not going to be overloaded.  It is declared as private in the base class. So the child class cannot use it. So I use "protected." I don't want to declare this method as a public method.  Had a conversation with James and James agrees with what I did.

modules/hdf5_handler/HDF5VlenAtomicArray.cc
kyang2014: Just checked in the update.

PR #1075, opened 2025-03-31
HYRAX-1489, 1. Make build_dmrpp work for enum. 2. Add a dmrpp generat…

kyang2014:
Before this pull request, some existing work has been done  to handle enum in the current dmrpp module. However, they are not tested and unsurprisingly not working. This pull request makes the dmrpp module handle enum and many tests are added. 
The detailed for build_dmrpp: 
1) Make build_dmrpp bypass the fill value case for enum. Having fill value for enum should be really rare. We can handle them later.
2) Make the print_dap4 method in the DmrppCommon include the generation of enum def information.

The detailed for accessing the data via dmrpp:
1) Fix a bug in the DmrppType factory that creates an enum variable. DMZ should be passed as a parameter when creating the variable.
2) Use val2buf to replace set_value when reading an enum scalar to ensure the data value can be correctly retrieved.
3) Add a method to process enum def object.
4) Make group process the enum def objects.

Various tests are added in both dmrpp generation and dmrpp data handling for enum.


jgallagher59701 (approved):
Ready to merge.

PR #1077, opened 2025-04-15
Rocky8 asan fixes

jgallagher59701:
Rocky8 asan fixes


kyang2014 (commented):
These lines look more like better programming practices rather than asan fixes. If this is not true, what does asan complain about?

dispatch/BESContainerStorageList.cc
```diff
@@ -49,371 +51,205 @@ using std::endl;
...
-                    done = true;
-                    ret = true;
-                }
-            }
-            else {
-                done = true;
-                ret = false;
-            }
-        }
+
```
kyang2014, line 78: I am not familiar with this part of code. It looks like you are doing the right thing. So does this also cause ASAN complain?

dispatch/BESContainerStorageList.h
```diff
@@ -29,83 +29,167 @@
...
+                            });
+    }
+
+    // Const version of the helper
+    auto find_entry_iterator(const std::string& persist_name) const {
+        return std::find_if(d_storage_entries.begin(), d_storage_entries.end(),
+                            [&persist_name](const StorageEntry& entry) {
+                                return entry.storage_obj && entry.get_name() == persist_name;
+                            });
+    }
```
kyang2014, line 134: Again just skim the code above. Not familiar with this part of code. Not sure if asan detects issues in this part of code or you just try to modernize the code.

dispatch/unit-tests/plistT.cc
```diff
@@ -227,4 +368,4 @@ int main(int argc, char*argv[])
 
     return wasSuccessful ? 0 : 1;
 }
-
+#endif
```
kyang2014, line 371: Again just skim the code, looks all right. Does the use of unique_ptr fix the memory leak?

xmlcommand/BESXMLDefineCommand.h
```diff
@@ -38,50 +38,42 @@
...
+    void prep_request() override;
 
-    virtual void dump(std::ostream &strm) const;
+    void dump(std::ostream &strm) const override;
 
-    static BESXMLCommand * CommandBuilder(const BESDataHandlerInterface &base_dhi);
+    static BESXMLCommand *CommandBuilder(const BESDataHandlerInterface &base_dhi);
 };
 
 #endif // A_BESXMLDefineCommand_h
```
kyang2014, line 79: The above code change looks like reducing code smells. Does the change of code fix asan issues?

jgallagher59701:
> These lines look more like better programming practices rather than asan fixes. If this is not true, what does asan complain about?

This PR fixes three unit tests for the code in bes/dispatch: pfileT, plistT, and containerT.  For all of these, the reported error on Linux (rocky8) is one or more memory leaks. Some of the leaks are just in the tests, but for containerT, the leaks was in BESContainerStorage, so I fixed that a moved to a more modern approach to implement the code. I also replaced the use of a broken version of the singleton pattern with a correct implementation that does not leak memory. More above.

dispatch/BESContainerStorageList.cc
jgallagher59701: Yes, there was a memory leak in the broken implementation of the singleton pattern.

dispatch/BESContainerStorageList.h
jgallagher59701: For this I was modernizing the code.

jgallagher59701:
@kyang2014 Can this be merged?

kyang2014 (approved):
Yes, thanks for all the explanation.

PR #1078, opened 2025-04-16
Hyrax 1490

kyang2014:
Fileout netCDF4: make the module support DAP4 Enum scalar and array. Tests are added. NASA files are also tested.


modules/fileout_netcdf/FONcD4Enum.h
```diff
@@ -0,0 +1,70 @@
...
+/** @brief A DAP4 Enum with file out netcdf information included
+ *
+ * This class represents a DAP4 Enum with additional information
+ * needed to write it out to a netcdf file. Includes a reference to the
+ * actual DAP4 Enum being converted
+ */
+class FONcD4Enum : public FONcBaseType
+{
+private:
+    libdap::D4Enum *			_f = nullptr;
```
jgallagher59701, line 53: We have used the convention that private fields in a class start with 'd_' and not a leading underscore since sometimes compilers use leading underscores for their own purposes. I'm not sure that is _really_ true anymore, but if you want to avoid the 'd_' then a common practice is to use a trailing underscore as a convention. 

Not a change you need to make, but I thought I'd mention it.

modules/fileout_netcdf/FONcD4Enum.h
```diff
@@ -0,0 +1,70 @@
...
+    nc_type                             basetype = NC_NAT;
```
jgallagher59701, line 54: But, 'basetype' and nc_enum_type_id should follow one of those conventions.

modules/fileout_netcdf/FONcTransform.h
```diff
@@ -59,19 +59,20 @@ class BESDataHandlerInterface;
  */
 class FONcTransform: public BESObj {
 private:
-	int _ncid = {0};
-	libdap::DDS *_dds = {nullptr};
+    int _ncid = {0};
```
jgallagher59701, line 62: Another FYI, when fields like ints and pointers are initialized, there is no need to use {}. The curly braces tell the compiler to use the object's constructor, but for those types, there is no constructor so it will just assign the value. Using the {} notation can make a big difference with fields that are objects.

jgallagher59701 (approved):
I had some comments, but they are mostly just things I saw that might save you a tiny bit of time in the future.

You might change one or two files or not, as you see fit. Then merge.

Thanks.

modules/fileout_netcdf/FONcD4Enum.h
kyang2014: I changed the code I added to follow the hyrax's naming conventions for private members.

modules/fileout_netcdf/FONcD4Enum.h
kyang2014: Make the new code follow the d_  name conventions.

modules/fileout_netcdf/FONcTransform.h
kyang2014: Yes, actually this is from the old code. I correct them.  However, I find quite a few private members not follow the _d naming conventions. I decide not to handle them. Instead I add a comment on ticket hyrax-1422 once I find time to clean up the fileout netCDF code.

PR #1079, opened 2025-04-16
Consolidating retired/unused/old code into single top-level retired dir

ndp-opendap:
This PR consolidates all of our retired/unused/old code into a single top level directory "retired" which contains a mirror of our code tree with the retired items in place. This will allow us to exclude the _bes/retired_ directory from examination by our IDE tools like **CLion** which will make searching the code base less likely to result in us inadvertently updating/changing unused code during a refactoring effort.

Going forward, when we retire a class or a package, the source code (and possibly attendant directories) should be moved into the _bes/retired_ directory into the subdirectory that matches its origin.

I included all of you as reviewers in order to ensure you're aware of this new practice.


dh-opendap (approved):
Sonuds reasonable.

jgallagher59701 (approved):
Thanks! Ready to merge.

PR #1080, opened 2025-04-22
Hyrax 1697 fix

kyang2014:
Improve the performance of large string writing in the fileout netCDF significantly as the follows:
1) Increase the variable chunk cache if the variable chunk size is big.
2) Re-implement the writing of large string data with one netCDF-4 API call rather than with many small subset calls.
3) Retire the equal-length string writing routine since the new string writing routine is a superset of that call without spending time  on checking the string size of each element.


modules/fileout_netcdf/FONcArray.cc
```diff
@@ -731,6 +734,15 @@ void FONcArray::define(int ncid) {
                         FONcUtils::handle_error(stax, err, __FILE__, __LINE__);
                     }
                 }
+                        size_t total_chunksizes = 1;
+        for (const auto& chunk_size:d_chunksizes)
+            total_chunksizes *= chunk_size;
+        // If the chunk size is greater than 4M, we increase the chunk cache size to be 64M.
+        if (total_chunksizes  >4194304) {
```
jgallagher59701, line 741: These might be better as constants so it's easier to see, at the top of the file, what the values are if we need to tune the parameters.

modules/fileout_netcdf/FONcArray.cc
```diff
@@ -1055,6 +1075,39 @@ void FONcArray::write_for_nc3_types(int ncid) {
     }
 }
 
+void FONcArray::write_string_array(int ncid) {
+    vector<size_t> var_count(d_ndims);
+    vector<size_t> var_start(d_ndims);
+
+    vector<char> text_data;
+    size_t last_dim_size = d_dim_sizes[d_ndims-1];
```
jgallagher59701, line 1083: Maybe make this a vector<size_t> ?

PR #1081, opened 2025-04-28
Hyrax 1499

kyang2014:
Add unlimited dimension support in the HDF5 handler. We store the information in the DMR as an attribute. The testsuite and the NASA testsuite are updated.


modules/hdf5_handler/h5dmr.cc
```diff
@@ -628,11 +635,70 @@ read_objects_base_type(D4Group * d4_grp, hid_t pid, const string & varname, cons
 
         dimnames_size = (int) (dt_inst.dimnames.size());
 
+        // Here we need to add the unlimited dimension info if dimension scale dimension names are present.
```
jgallagher59701, line 638: This might be hard to do, but could this part of the function be extracted to its own function? That would simplify the code. If it's too hard to do that, this can stay as it is. Your call on that.

jgallagher59701 (approved):
One suggestion. Thanks.

PR #1082, opened 2025-04-29
HYRAX-1741, add byteOrder to variables that don't have data.

kyang2014:
Dmrpp: 
1) Add byteOrder to variables that don't have any data. Add a test and also update two existing tests.
2) Also return empty data when the array size is 0. Before it throws an error. A test is added.


modules/dmrpp_module/DmrppCommon.cc
```diff
@@ -546,6 +546,14 @@ DmrppCommon::print_chunks_element(XMLWriter &xml, const string &name_space)
             throw BESInternalError("Could not write fillValue attribute.", __FILE__, __LINE__);
     }
 
+    if(d_chunks.empty()){
+        if(d_byte_order.empty()==false) {
```
jgallagher59701, line 550: The d_chunks.empty() and d_byte_order.empty() == false tests could be combined. It's tempting to combine all three, but since the third 'if' is run for side effect, I think that would make it pretty hard to read.

jgallagher59701 (approved):
I made one suggestion, but it's really up to you if that's an improvement or not. 

Ready to merge. Thanks.

PR #1083, opened 2025-04-29
Correcting embedded BES conf string issues:

ndp-opendap:
* Switched the bes configuration strings to python raw strings so that the various usages of the back slash character don't break python and cause runtime warnings. 

When I tried to use the software I encountered this error:
```
/Users/ndp/OPeNDAP/hyrax/build/bin/get_dmrpp_h4:156: SyntaxWarning: invalid escape sequence '\.'
  bes_tmp_conf_str='''BES.LogName=./bes.log
```
Which I fixed in this PR.

Also:
* Added new line to the ends of the conf strings
* Removed extraneous end of line semicolons


jgallagher59701 (approved):
Thanks!

PR #1085, opened 2025-05-02
HYRAX-1743, add the unlimited dimension support. 1) Make sure the unl…

kyang2014:
…imited dimension info is passed to netCDF. 2) Use the netCDF write APIs that can handle both unlimited dimension and normal cases to replace the current ones that can only handle the normal case. 3) Update and add tests.


modules/fileout_netcdf/FONcArray.cc
```diff
@@ -581,7 +592,16 @@ void FONcArray::define(int ncid) {
             int dimnum = 0;
             for (; i != e; i++) {
```
jgallagher59701, line 593: This loop could be a range-based for-loop.

jgallagher59701 (approved):
One comment - but, really, this is ready to merge.

I did see there's a bit of older code inside #if 0 ... #endif blocks and we could remove that.

Thanks.

modules/fileout_netcdf/FONcArray.cc
kyang2014: Somehow the iterartor doesn't like the range-based for-loop. So just keep the current code.

kyang2014: I have another ticket to clean up fileout netCDF. I will remove #if 0 #endif and other things when I work on that ticket

PR #1090, opened 2025-05-08
ReturnManager --> Meyers singleton, other refactors

jgallagher59701:
ReturnManager --> Meyers singleton, other refactors


sawer297 (approved):
looks good to me

PR #1091, opened 2025-05-09
HYRAX-1751, adding the gen_dmrpp_sidecar tests to th…

kyang2014:
…e integration test. Now the two python tests that generate dmrpp are integrated to the the BES testsuite. The most challenging part is to get "make distcheck" working.

### DIFF
--- a/modules/dmrpp_module/get_dmrpp/unit-tests/Makefile.am
+++ b/modules/dmrpp_module/get_dmrpp/unit-tests/Makefile.am
@@ -27,7 +27,9 @@ cp_build_script:
 	cp -f  ../$(srcdir)/get_dmrpp_h5 $(builddir); \
 	fi;
 
-check-local: cp_build_script check-local-gen-dmrpp 
+check-local: check-local-gen-dmrpp 
+
+check-local-gen-dmrpp: cp_build_script
 
 check-local-gen-dmrpp:
 	if test $(srcdir) != .; then \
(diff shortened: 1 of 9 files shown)


modules/dmrpp_module/get_dmrpp/unit-tests/Makefile.am
```diff
@@ -0,0 +1,55 @@
...
+	cp -f $(abs_top_builddir)/modules/fileout_netcdf/.libs/libfonc_module.so $(builddir); \
+	cp -f $(abs_top_builddir)/xmlcommand/.libs/libdap_xml_module.so $(builddir); \
+	cp -f $(abs_top_builddir)/dap/.libs/libdap_module.so $(builddir); \
+	else \
+	cp -f  ../$(srcdir)/gen_dmrpp_side_car $(builddir); \
+	cp -f  ../$(srcdir)/get_hdf_side_car $(builddir); \
+	cp -f  ../$(srcdir)/get_dmrpp_h5 $(builddir); \
+	fi;
+
+check-local: cp_build_script check-local-gen-dmrpp
```
jgallagher59701, line 30: I was thinking about this and I _think_ it's OK, but it might be that a parallel make might build these two dependencies at the same time. That might be OK most of the time, but not always. A way to safeguard against that would be to make "cp_build_script" a dependency of the "check-local-gen-dmrpp" target. That would look like this:

```
check-local: check-local-gen-dmrpp 

check-local-gen-dmrpp: cp_build_script
    if test $(srcdir) != .; then \
    ...

```

So check-local depends on check-local-gen-dmrpp and check-local-gen-dmrpp depends on cp_build_script.

jgallagher59701 (approved):
I think the suggestion to the Makefile.am might help avoid 'random' behavior, but I'm not 100% sure it's needed.

modules/dmrpp_module/get_dmrpp/unit-tests/Makefile.am
kyang2014: Will do and check.

PR #1095, opened 2025-05-20
meyer applied

sawer297:
deleted the deconstructor, rest of the singleton looks to be already in meyer pattern


ndp-opendap (approved):
The rub with changing things here is that the only real regression test we have for this is to deploy it in NGAP and see if something breaks. Or maybe the unit-tests for EUC are better than I remember...

:shipit:

jgallagher59701 (approved):
I would make the dump public and have it do something useful or just drop it. The method is defined in BESObj and that version shadows BESObj::dump().

Also, we can leave the destructor in as defined - it'll be the default impl, but we might as well include it just to keep it in mind that the destructor is the default one.

jgallagher59701 (approved):

PR #1097, opened 2025-05-20
Hyrax 1753

kyang2014:
For the disk cache handling in the HDF5 handler CF option, we need to use a loop to write or read the cache data when the cache size is bigger than 2GB due to the limitation of the underneath system IO calls.  This pull request resolves this issue. The code is tested successfully with a real NASA file.


jgallagher59701 (approved):
This looks good, but I have a question. Why did you settle on 2^30 and not 2^31 (or 2^31 - 1) as the value for max_read_buf_size and HDF5DiskCache::CACHE_BUF_SIZE? I'm asking because I'm about to write code very similar to this!

Did you know that in C++14 you can write numbers like this: 10'737'341'842 (or you could use 0x3F'FF'FF'FF).

Also, since that's a number and it's used in more than one place, maybe it should be a constant in a header?

kyang2014:
> This looks good, but I have a question. Why did you settle on 2^30 and not 2^31 (or 2^31 - 1) as the value for max_read_buf_size and HDF5DiskCache::CACHE_BUF_SIZE? I'm asking because I'm about to write code very similar to this!
> 
> Did you know that in C++14 you can write numbers like this: 10'737'341'842 (or you could use 0x3F'FF'FF'FF).
> 
> Also, since that's a number and it's used in more than one place, maybe it should be a constant in a header?
When searching on-line, https://stackoverflow.com/questions/70368651/why-cant-linux-write-more-than-2147479552-bytes#70370002
it says the limitation for Linux IO call is a little bit less than 2^31(2^31-one page size), the exact number is 2147479552, so I choose 2^30 rather than 2^31.

PR #1098, opened 2025-05-20
meyers applied

sawer297:
applied meyers pattern


jgallagher59701 (changes requested):
Instead of 'virtual ~SingletonList()' this should be '~SingletonList() override'. A minor point, but now's a good time to make that change.

jgallagher59701 (approved):

PR #1099, opened 2025-05-22
HYRAX-1688,HYRAX-1756, remove the core dump and uncaught c++ exceptio…

kyang2014:
…n when just typing build_dmrpp_h4. Also make the help description clear and accurate.

### DIFF
--- a/modules/dmrpp_module/build_dmrpp_h4/build_dmrpp_h4.cc
+++ b/modules/dmrpp_module/build_dmrpp_h4/build_dmrpp_h4.cc
@@ -51,18 +51,22 @@ using namespace build_dmrpp_util_h4;
 
 void usage() {
     const char *help = R"(
-    build_dmrpp -h: Show this help
+    build_dmrpp_h4 -h: Show this help
 
-    build_dmrpp -V: Show build versions for components that make up the program
+    build_dmrpp_h4 -V: Show build versions for components that make up the program
 
-    build_dmrpp -f <data file> -r <dmr file> [-u <href url>]: As above, but uses the DMR
-       read from the given file.
+    build_dmrpp_h4 -f <data file> -r <dmr file> [-u <href url>] [-M] [-D] [-v] [-V] [-d] 
 
-    Other options:
-        -v: Verbose
-        -d: Turn on BES software debugging output
-        -M: Add information about the build_dmrpp software, incl versions, to the built DMR++
-        -D: Disable the generation of HDF-EOS2/HDF4 missing latitude/longitude)";
+    options:
+        -f: HDF4/HDF-EOS2 file to build DMR++ from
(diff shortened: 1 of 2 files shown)


modules/dmrpp_module/build_dmrpp_h4/build_dmrpp_h4.cc
```diff
@@ -51,18 +51,22 @@ using namespace build_dmrpp_util_h4;
...
+    options:
+        -f: HDF4/HDF-EOS2 file to build DMR++ from
+        -r: DMR file to build DMR++ from
+        -u: The href value to use in the DMR++ for the data file
+        -M: Add information about this software, incl versions, to the built DMR++
+        -D: Disable the generation of HDF-EOS2/HDF4 missing latitude/longitude
+        -h: Show this help
+        -v: Verbose information helpful for debugging and understanding the program working flow
+        -V: Show build versions for components that make up the program
+        -d: Turn on BES software debugging output)";
```
jgallagher59701, line 69: There is an extra ')' at the end of the text on this line.

modules/dmrpp_module/build_dmrpp_h4/build_dmrpp_h4.cc
```diff
@@ -51,18 +51,22 @@ using namespace build_dmrpp_util_h4;
...
     const char *help = R"(
-    build_dmrpp -h: Show this help
+    build_dmrpp_h4 -h: Show this help
 
-    build_dmrpp -V: Show build versions for components that make up the program
+    build_dmrpp_h4 -V: Show build versions for components that make up the program
 
-    build_dmrpp -f <data file> -r <dmr file> [-u <href url>]: As above, but uses the DMR
-       read from the given file.
+    build_dmrpp_h4 -f <data file> -r <dmr file> [-u <href url>] [-M] [-D] [-v] [-V] [-d]
```
jgallagher59701, line 58: I think there should be a vertical bar between "-f <data file> -r <dmr file>" like "-f <data file> | -r <dmr file>" although I'm not 100% certain.

jgallagher59701 (approved):
I think these two changes are needed, but it's easy to not see some things in a diff.

Thanks for fixing this info up.

modules/dmrpp_module/build_dmrpp_h4/build_dmrpp_h4.cc
kyang2014: That ) has to be there for the raw string literal to work.

modules/dmrpp_module/build_dmrpp_h4/build_dmrpp_h4.cc
kyang2014: No, there should be no vertical bar. Just -f -r.

PR #1102, opened 2025-05-28
Use Python `unittest`'s `addCleanup` to handle asset deletion

hannahilea:
Previously, if the `test_gen_dmrpp_side_car` or `test_gen_dmrpp_side_car2` tests failed, the files they created persisted, which was confusing in cases where later runs seemed to (but did not actually) generate those same files. Now, using the `addCleanup` functionality, the cleanup happens at the end of the test execution, regardless of its outcome.

### DIFF
--- a/modules/dmrpp_module/get_dmrpp/unit-tests/gen_dmrpp_side_car_test.py
+++ b/modules/dmrpp_module/get_dmrpp/unit-tests/gen_dmrpp_side_car_test.py
@@ -3,6 +3,9 @@
 Test gen_dmrpp_side_car script.
 python3 -m unittest gen_dmrpp_side_car_test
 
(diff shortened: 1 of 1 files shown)


kyang2014 (approved):

hannahilea:
> Otherwise it can be difficult to trace problems if the intermediate results have been removed. 

I added the flag, and I agree. 

In some ways the original behavior here (no intermediate results removed if test fails) are good---but also, they can be misleading, as in the case of my/our original debugging, where it took some sleuthing to realize that those intermediate results were stale and not being regenerated due to an upstream failure (the same one made more verbose in #1101). 

...which is to say, it would be fair to decide not to merge this change at all; as always, it is a tradeoff.

PR #1104, opened 2025-06-02
Don't use direct IO when this variable has an unlimited dimension.

kyang2014:


modules/fileout_netcdf/FONcTransform.cc
```diff
@@ -1619,6 +1634,25 @@ bool FONcTransform::obtain_unlimited_dimension_info(libdap::D4Group *d4_grp, vec
     return ret_value; 
 }
 
+bool FONcTransform::check_var_unlimited_dimension(libdap::Array *t_a, const vector<string> &unlimited_dimnames) {
+
+    bool ret_value = false;
+    for (const auto &und:unlimited_dimnames) {
+        Array::Dim_iter dim_i = t_a->dim_begin();
```
jgallagher59701, line 1641: Check and see if libdap::Array has a method like dimensions() that can be used with a range-based for loop. If so, use that unless it will take too much time to make the change.

jgallagher59701 (approved):
One suggestion, but it's just that. Your judgement if it's worth the time.

Thanks, ready to merge.

PR #1106, opened 2025-06-04
meyers pattern applied

sawer297:
meyers singleton pattern applied

### DIFF
--- a/dispatch/BESCatalogList.cc
+++ b/dispatch/BESCatalogList.cc
@@ -72,28 +72,6 @@ BESCatalogList::TheCatalogList()
     return &catalog;
 }
 
-///**
-// * private static that only get's called once by using pthread_once and
-// * pthread_once_t mutex.
-// */
-//void BESCatalogList::initialize_instance()
-//{
-//    d_instance = new BESCatalogList;
-//#ifdef HAVE_ATEXIT
-//    atexit(delete_instance);
-//#endif
-//
-//}
-
-///**
-// * Private static function can only be called by friends and pThreads code.
-// */
-//void BESCatalogList::delete_instance()
-//{
-//    delete d_instance;
-//    d_instance = 0;
-//}
-
 /** @brief construct a catalog list
  *
  * @see BESCatalog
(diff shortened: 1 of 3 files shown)


dispatch/BESCatalogList.h
```diff
@@ -98,13 +98,17 @@ class BESCatalogList: public BESObj {
...
     typedef std::map<std::string, BESCatalog *>::iterator catalog_iter;
     typedef std::map<std::string, BESCatalog *>::const_iterator catalog_citer;
 
+    BESCatalogList(const BESCatalogList&) = delete;
+    BESCatalogList& operator=(const BESCatalogList&) = delete;
+
     static BESCatalogList * TheCatalogList();
 
-    BESCatalogList();
     virtual ~BESCatalogList();
```
jgallagher59701, line 112: Since BESCatalog is a child of BESObj, the destructor should be _~BESCatalogList() override;_ and not _virtual ..._

dispatch/BESCatalogList.h
```diff
@@ -98,13 +98,17 @@ class BESCatalogList: public BESObj {
 
     friend class BESCatalogListTest;
 
+    BESCatalogList();
+
 public:
     typedef std::map<std::string, BESCatalog *>::iterator catalog_iter;
```
jgallagher59701, line 104: This can be rewritten with the C++14 _using_ syntax lie _using catalog_iter = td::map<std::string, BESCatalog *>::iterator;_

This is a general rule - we can replace the old C (c. 1970) with a _using_ declaration. It's not required, but I've been doing this as I come across typedefs.

dispatch/BESCatalogList.h
```diff
@@ -98,13 +98,17 @@ class BESCatalogList: public BESObj {
...
     typedef std::map<std::string, BESCatalog *>::const_iterator catalog_citer;
```
jgallagher59701, line 105: Same deal here.

jgallagher59701 (approved):
This can be merged, but I have made some suggestions

jgallagher59701 (commented):
Wait... There's only a change to the header?

dispatch/BESCatalogList.cc
```diff
@@ -68,32 +68,31 @@ BESCatalogList *BESCatalogList::d_instance = 0;
...
- * private static that only get's called once by using pthread_once and
- * pthread_once_t mutex.
- */
-void BESCatalogList::initialize_instance()
-{
-    d_instance = new BESCatalogList;
-#ifdef HAVE_ATEXIT
-    atexit(delete_instance);
-#endif
+///**
```
jgallagher59701, line 75: Just remove this. However, if you think old code sections should be retained for a while, wrap them in 
#if 0
...
#endif

Same for the sections below.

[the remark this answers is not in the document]
sawer297: 10-4
was waiting on confirm that the code was good to go
then i would make one last change to remove all comments before merging

sawer297: kk
on it

PR #1107, opened 2025-06-05
HYRAX-1771, turn off the generation of Unlimited_Dimensions attribute…

kyang2014:
… by default, Users can turn it on with BES key.


jgallagher59701 (approved):
Ready to merge. Thanks.

PR #1108, opened 2025-06-06
meyers applied

sawer297:
meyers singleton pattern applied


jgallagher59701 (approved):
Please remove the commented out code and then squash and merge.
Thanks.

PR #1109, opened 2025-06-06
Check for existing ".dmrpp" extension in url

kneumiller:


modules/ngap_module/NgapApi.cc
```diff
@@ -434,6 +434,11 @@ string NgapApi::convert_ngap_resty_path_to_data_access_url(const std::string &re
     cmr_response.Parse(cmr_json_string.c_str());
     data_access_url = find_get_data_url_in_granules_umm_json_v1_4(restified_path, cmr_response);
 
+    // Check for existing .dmrpp and remove it if found. - kln 6/6/25
+    size_t pos = data_access_url.find(".dmrpp");
```
jgallagher59701, line 438: This will avoid replacing the .dmrpp in the middle of the string. I think the code should not do that. Admittedly, that seems like an obscure case, but it might happen.

const std::string suffix = ".dmrpp";

if (data_access_url.size() >= suffix.size() &&
    str.compare(data_access_url.size() - suffix.size(), suffix.size(), suffix) == 0) {
    str.erase(data_access_url.size() - suffix.size());
}

modules/ngap_module/unit-tests/NgapApiTest.cc
```diff
@@ -337,6 +337,52 @@ class NgapApiTest: public CppUnit::TestFixture {
...
+        }
+        CPPUNIT_ASSERT_EQUAL_MESSAGE("URL without .dmrpp should not be changed.",
+                                     original_url, url_without_suffix);
+
+        // Case 3: URL with '.dmrpp' in the middle. The current 'find' logic will
+        // remove the first occurrence.
+        string url_with_middle_dmrpp = "https://my-domain.com/path/file.dmrpp.nc";
+        string expected_middle_removal_url = "https://my-domain.com/path/file.nc";
+
+        size_t pos3 = url_with_middle_dmrpp.find(".dmrpp");
```
jgallagher59701, line 376: Should the code remove the '.dmrpp' in this case? I think this test should fail.

jgallagher59701 (changes requested):

PR #1110, opened 2025-06-09
meyers applied to BESStoreDapResultCache

sawer297:
modified the BESStoreDapResultCache.h and *.cc file to follow the meyers pattern
removed a few lines in the unit-tests/ResponseBuilderTest.cc

### DIFF
--- a/dap/BESStoredDapResultCache.cc
+++ b/dap/BESStoredDapResultCache.cc
@@ -229,9 +229,6 @@ BESStoredDapResultCache::get_instance(const string &data_root_dir, const string 
                     "Cache is DISABLED"<< endl);
            }
             else {
-#ifdef HAVE_ATEXIT
-                atexit(delete_instance);
-#endif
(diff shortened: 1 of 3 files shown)


dap/BESStoredDapResultCache.cc
```diff
@@ -247,24 +235,19 @@ BESStoredDapResultCache *
...
+        static BESStoredDapResultCache instance;
         d_enabled = d_instance->cache_enabled();
         if(!d_enabled){
-            delete d_instance;
-            d_instance = NULL;
             BESDEBUG("cache", "BESStoredDapResultCache::"<<__func__ << "() - " <<
                 "Cache is DISABLED"<< endl);
+            return nullptr;
        }
         else {
```
jgallagher59701, line 245: You can remove this else {} and practically you can drop the BESDebug lines

sawer297: 10-4

PR #1113, opened 2025-06-12
Repair AWS query string filtering

ndp-opendap:
The previous incarnation of this dropped everything after the first kvp. Now only keys beginning with "X-Amz-" will be removed and the rest preserved.


jgallagher59701 (approved):
Suggestions, but thanks. This is ready to merge.

PR #1115, opened 2025-06-16
HYRAX-1781, make gen_dmrpp_side_car and get_dmrpp_h4 add the build_dm…

kyang2014:
…rpp metadata information to the dmrpp file by default. Add update the unit test script and the baseline file.


jgallagher59701 (approved):
Ready to merge. Thanks!

PR #1116, opened 2025-06-18
HYRAX-1362,improve the error handling. Make sure every exception is c…

kyang2014:
…aught and the program exits gracefully.
Tested manually to make sure the exception message is correct when the exception is caught successfully.

### DIFF
--- a/modules/dmrpp_module/build_dmrpp_h4/build_dmrpp_util_h4.cc
+++ b/modules/dmrpp_module/build_dmrpp_h4/build_dmrpp_util_h4.cc
@@ -55,7 +55,7 @@
 
 #define COMP_INFO 512 /*!< Max buffer size for compression information.  */
 
-#define ERROR(x) do { cerr << "ERROR: " << x << " " << __FILE__ << ":" << __LINE__ << endl; } while(false)
+#define ERROR(x) do { cerr << "Internal Error: " << x << " at " << __FILE__ << ":" << __LINE__ << endl; } while(false)
 
 /*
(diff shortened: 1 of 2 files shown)


modules/dmrpp_module/build_dmrpp_h4/build_dmrpp_util_h4.cc
```diff
@@ -233,16 +234,18 @@ int read_chunk(int sdsid, SD_mapping_info_t *map_info, int *origin)
...
 
     intn emptySDS = 0;
+    string ret_value;
     
     if (SDcheckempty(sdsid,&emptySDS) == FAIL) {
         SDendaccess(sdsid);
-        throw BESInternalError("SDcheckempty fails. ",__FILE__,__LINE__);
+        dmrpp_h4_error = true;
+        err_msg = "SDcheckempty fails at "+string(__FILE__) +":" + to_string(__LINE__);
+        return ret_value;
```
jgallagher59701, line 246: Is the intent here to return "" on error?

jgallagher59701 (approved):
I have a question, but if the answer is 'yes' then this can be merged. Thanks.

modules/dmrpp_module/build_dmrpp_h4/build_dmrpp_util_h4.cc
kyang2014: Yes, since dmrpp_h4_error is set to true, the caller will handle this as an error. The ret_value doesn't matter since it will not be used.

PR #1118, opened 2025-06-23
meyers applied to BESContextManager.h and *.cc

sawer297:
Unnecessary ticket reference: Hyrax-1774
applied the Meyers singleton pattern to BESContextManager.h and *.cc

### DIFF
--- a/dispatch/BESContextManager.cc
+++ b/dispatch/BESContextManager.cc
@@ -46,7 +46,6 @@ using std::string;
 using std::ostream;
 
 BESContextManager *BESContextManager::d_instance = nullptr;
-static std::once_flag d_euc_init_once;
 
 #define MODULE "context"
 #define prolog std::string("BESContextManager::").append(__func__).append("() - ")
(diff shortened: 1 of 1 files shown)


dispatch/BESContextManager.cc
```diff
@@ -213,8 +211,9 @@ void BESContextManager::dump(ostream &strm) const
 BESContextManager *
 BESContextManager::TheManager()
 {
-    std::call_once(d_euc_init_once,BESContextManager::initialize_instance);
-    return d_instance;
+    static BESContextManager manager;
+    // std::call_once(d_euc_init_once,BESContextManager::initialize_instance);
```
jgallagher59701, line 215: Please remove the commented out code and also delete the d_euc_init_once variable.

jgallagher59701 (approved):
Two simple edits, make those and merge. No need for second review.
Thanks.

dispatch/BESContextManager.cc
sawer297: 10-4

PR #1120, opened 2025-06-25
Hyrax-1774: meyers applied to BESInfoList

sawer297:
Meyers singleton pattern applied to BESInfoList.h and *.cc


jgallagher59701 (approved):
One change, then merge.

PR #1128, opened 2025-07-02
Hyrax-1803: meyers applied to CredentialsManager

sawer297:
applied the Meyers singleton pattern to ./http/CredentialsManager.h and *.cc


jgallagher59701 (approved):
These changes look good. If it passes CICD, merge. Thanks.

PR #1129, opened 2025-07-03
Hyrax-1803: meyers applied to BESReturnManager

sawer297:
applied the meyers singleton pattern to ./dispatch/BESReturnManager.h and *.cc


jgallagher59701 (approved):
Ready to merge. Thanks.

PR #1130, opened 2025-07-08
Optional checksums

ndp-opendap:
This updates the bes tests so that checksums are both requested and handled during the testing process.

In general this means modifying the bes command files to include the setContext to trigger the production of checksums:
```xml
<setContext name=\"dap4_checksums\">true</setContext>
```
And then modifying the handler test macros M4 files by adding a `-C` to the `getdap4` invocations.

In some cases existing baselines contained duplicate `DAP4_Checksum_CRC32` Attributes, this was also corrected.

**NOTE: This PR will not pass until [the companion libdap4 PR for optional checksums is merged](https://github.com/OPENDAP/libdap4/pull/289).**


dap/BESDapResponseBuilder.cc
```diff
@@ -120,6 +120,7 @@
 #include "BESStopWatch.h"
 #include "DapFunctionUtils.h"
 #include "RequestServiceTimer.h"
+#include "../dispatch/BESContextManager.h"
```
jgallagher59701, line 123: It's better to not use a relative path like this to include the file. You could add `-I$abs_topsrcdir)` to AM_CPPFLAGS in the Makefile.am and then use "dispatch/BESContextManager.h"

jgallagher59701 (changes requested):
See if you can make that change to the include. Let me know if there are hassles.

dap/BESDapResponseBuilder.cc
```diff
@@ -1453,6 +1454,19 @@ void BESDapResponseBuilder::send_dap4_data(ostream &out, DMR &dmr, bool with_mim
     }
 }
 
+const auto DAP4_CHECKSUMS_KEY="dap4_checksums";
+
+bool use_dap4_checksums() {
+    bool found;
```
kyang2014, line 1460: Should provide an initial value for found.

kyang2014 (changes requested):

dap/BESDapResponseBuilder.cc
ndp-opendap: Thanks! Fixed.

kyang2014 (approved):

PR #1131, opened 2025-07-08
Hyrax 1800

kyang2014:


jgallagher59701 (approved):
This looks fine. Ready to merge. Thanks.

One thing that sonar scan shows is that there is a fairly high level of duplication in this code - blocks that appear more than once that are identical. It can be hard to remove those, but we should think about how we might tackle that in the future.

PR #1138, opened 2025-07-18
Hyrax-1846: meyers applied to BESStoredDapResultCache

sawer297:
applied the meyer singleton pattern to ./dap/BESStoredDapResultCache.h and *.cc


jgallagher59701 (approved):
Remove the old code inside #if 0 ... #endidf sections and it's ready to merge.

PR #1139, opened 2025-07-21
Removed the trailing null byte

kneumiller:
Since C++11, std::string is guaranteed to store its data contiguously in memory and to be null-terminated. This means that buf.data() and buf.c_str() will always return a pointer to a null-terminated C-style string. Specifically, buf[buf.size()] is guaranteed to be '\0'.


jgallagher59701 (approved):
Ready to merge. Thanks.

PR #1141, opened 2025-08-01
HYRAX-1835: Add logging checkpoints to support BES portion of end-to-end hyrax profiling

hannahilea:
## Description

From list in ticket, add logging checkpoints for:
- [ ] DMRpp
- [ ] data response

## Tasks

- [ ] ~Tests added/updated~ N/A
- [ ] ~Dead code removed~ N/A
- [ ] No TODOs added


http/CurlUtils.cc
```diff
@@ -1242,6 +1242,7 @@ static size_t string_write_data(void *buffer, size_t size, size_t nmemb, void *d
  */
 void http_get(const string &target_url, string &buf) {
     BESDEBUG(MODULE, prolog << "BEGIN\n");
+    BES_PROFILE_TIMING(string("Make HTTP GET request - ") + target_url);
```
hannahilea, line 1245: Supports item 5 (see description)

dispatch/BESInterface.cc
```diff
@@ -411,7 +411,10 @@ int BESInterface::execute_request(const string &from)
         // HK-474. The exception caused by the errant config file in the ticket is
         // thrown from inside SaxParserWrapper::rethrowException(). It will be caught
         // below. jhrg 11/12//19
-        execute_data_request_plan();
+        {
+            BES_PROFILE_TIMING("Execute data request plan");
```
hannahilea, line 415: Supports item 6.2 (see description)

http/CurlUtils.cc
```diff
@@ -1503,6 +1504,7 @@ curl_slist *add_edl_auth_headers(curl_slist *request_headers) {
  */
 curl_slist *
 sign_s3_url(const string &target_url, AccessCredentials *ac, curl_slist *req_headers) {
+    BES_PROFILE_TIMING(string("Sign URL - ") + target_url);
```
hannahilea, line 1507: Supports item 5 (see description)

modules/dmrpp_module/Chunk.cc
```diff
@@ -1228,6 +1229,10 @@ void Chunk::read_chunk() {
     if (d_is_read)
         return;
 
+    if (get_data_url() != nullptr) {
+        BES_PROFILE_TIMING(string("Read chunk - ") + get_data_url()->str() + string(" - size: ") + ::to_string(get_size()) + string(" - offset: ") + ::to_string(get_offset()));
```
hannahilea, line 1233: Supports item 6.1 (see description)

modules/dmrpp_module/CurlHandlePool.cc
```diff
@@ -227,6 +228,8 @@ dmrpp_easy_handle::~dmrpp_easy_handle() {
  * all cleanup.
  */
 void dmrpp_easy_handle::read_data() {
+    BES_PROFILE_TIMING(string("DMRPP read data request - ") + d_url->str());
```
hannahilea, line 231: Supports item 6 (see description)

modules/dmrpp_module/ngap_container/NgapOwnedContainer.cc
```diff
@@ -213,11 +213,11 @@ bool NgapOwnedContainer::file_to_string(int fd, string &content) {
  */
 string NgapOwnedContainer::build_data_url_to_daac_bucket(const string &rest_path) {
     BES_MODULE_TIMING(prolog + rest_path);
+    BES_PROFILE_TIMING(string("Build granule URL - ") + rest_path);
```
hannahilea, line 216: Supports item 3 (see description)

modules/dmrpp_module/Chunk.cc
ndp-opendap: My understanding was that the mission was to profile SuperChunk performance.  Placing a timer here will mean that it gets triggered thousands of times, even tens of thousands of time per request. I believe that this  will flood the logs and have noticeable performance consequences.

ndp-opendap: Also - I am pretty sure that the way the code is written is to put all the chunks into a SuperChunk. (Some SuperChunk represent only a single Chunk) And then we read the SuperChunk(s). Once the SuperChunk has read the constituent Chunk data the constituent Chunks are all marked as read, and the individual Chunk::read() methods are not actually called.  I think...

hannahilea: > Placing a timer here will mean that it gets triggered thousands of times, even tens of thousands of times per request depending on the granule. I believe that this will flood the logs and have noticeable performance consequences.

Yikes! Great point, removing. 

Why did I add this in the first place? In one of my test cases, where I have a single constraint expression, I never hit the SuperChunk data retrieval calls (the checkpoints added to SuperChunk.cc). I'll try to hunt down a less-noisy checkpoint site for that case.

modules/dmrpp_module/Chunk.cc
```diff
@@ -1269,11 +1274,14 @@ void Chunk::read_chunk() {
 
 // direct IO method that reads chunks.
 void Chunk::read_chunk_dio() {
-
     // Read chunk for dio - use read_chunk() as a reference.
     if (d_is_read)
         return;
 
+    if (get_data_url() != nullptr) {
+        BES_PROFILE_TIMING(string("Read chunk dio - ") + get_data_url()->str() + string(" - size: ") + ::to_string(get_size()) + string(" - offset: ") + ::to_string(get_offset()));
```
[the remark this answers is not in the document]
hannahilea: Removed.

modules/dmrpp_module/Chunk.cc
hannahilea: (Also, fwiw, you are correct about 

> And it seems like I have contradicted myself: If the Chunk::read() methods are never called then why not instrument them? Indeed. Maybe you should leave and see what shakes out in testing.

They are called, and called a lot! It was one of the things I'd flagged to ask you about in person, even before you'd made this comment on the PR. :) )

dispatch/BESInterface.cc
hannahilea: Removed

modules/dmrpp_module/SuperChunk.h
```diff
@@ -81,6 +81,7 @@ class SuperChunk {
     std::shared_ptr<http::url> get_data_url() { return d_data_url; }
     virtual unsigned long long get_size() const { return d_size; }
     virtual unsigned long long get_offset() const { return d_offset; }
+    virtual size_t get_chunks_size() const { return d_chunks.size(); }
```
hannahilea, line 84: I'd prefer to name this something like `get_chunk_count()` or `get_num_chunks()`, but `get_chunks_size` is used elsewhere for this same concept, and having function name parity seemed useful.

http/CurlUtils.cc
```diff
@@ -1850,6 +1850,7 @@ std::shared_ptr<http::EffectiveUrl> get_redirect_url(const std::shared_ptr<http:
         throw BESSyntaxUserError(err, __FILE__, __LINE__);
     }
 
+    BES_PROFILE_TIMING("Request redirect url - " + origin_url->get_url_no_query());
```
hannahilea, line 1853: Supports item 5 (see description)

modules/dmrpp_module/SuperChunk.cc
```diff
@@ -714,6 +715,9 @@ void SuperChunk::retrieve_data() {
         BESDEBUG(SUPER_CHUNK_MODULE, prolog << "SuperChunk (" << (void **) this << ") has already been read! Returning." << endl);
         return;
     }
+
+    BES_PROFILE_TIMING(string("Request SuperChunk data - ") + (get_data_url() ? get_data_url()->get_url_no_query() : "") + string(" - ") + ::to_string(get_size()) + string(" byte(s) - ") + ::to_string(get_chunks_size()) + " chunk(s)");
```
hannahilea, line 719: Supports item 6.1 (see description)

modules/dmrpp_module/SuperChunk.cc
```diff
@@ -904,6 +914,8 @@ void SuperChunk::read_unconstrained_dio() {
     //Retrieve data for the direct IO case.
     retrieve_data_dio();
 
+    BES_PROFILE_TIMING(string("Handle SuperChunk data dio - ") + (get_data_url() ? get_data_url()->get_url_no_query() : "") + string(" - ") + ::to_string(get_size()) + string(" byte(s) - ") + ::to_string(get_chunks_size()) + " chunk(s) - Using multithreading: " + (DmrppRequestHandler::d_use_compute_threads ? "true" : "false"));
```
hannahilea, line 917: Supports item 6.2 (see description)

modules/dmrpp_module/Chunk.cc
hannahilea: Okay, actually, not adding any non-superchunk checkpoints unless we intentionally decide otherwise.

server/beslog2json.py
```diff
@@ -385,7 +387,9 @@ def timing_log_to_json(log_fields, json_log_record):
     """
     debug("Processing TIMING_MESSAGE_TYPE")
     send_it = False
-    if TRANSMIT_TIMING_LOG:
+
+    transmit_log = TRANSMIT_TIMING_LOG or (TRANSMIT_PROFILING_LOG and log_fields[12].startswith("Profile timing:"))
```
hannahilea, line 391: I was able to manually test the `square_bracket_timing_record()` path, but not this `timing_log_to_json` path; not sure if there are tests somewhere that I don't see to test this behavior? Or is a switch for outputting this alternate log type from the bes, so I can at least test manually?

server/beslog2json.py
```diff
@@ -72,6 +72,7 @@
 TRANSMIT_ERROR_LOG   = True
 TRANSMIT_VERBOSE_LOG = True
 TRANSMIT_TIMING_LOG  = False
+TRANSMIT_PROFILING_LOG  = True
```
hannahilea, line 75: ‼️ **Setting this to default true!!!**

Alternatively, we could set this to default false and set the flag directly in hyrax-docker/ngap/entrypoint.sh. Let's make sure we're on the same page about this choice before merging this branch!

ndp-opendap: This is a nice minimally invasive way to go.

server/beslog2json.py
ndp-opendap: You can test it manually by taking grabbing a timing/profiling  line from the `/var/log/bes/bes.log` where the records are separated by `|&|` . The square bracket business is for log lines (early in the startup) where each value is enclosed in square brackets: `[time stuff][pid][message stuff]`

server/beslog2json.py
```diff
@@ -704,11 +711,18 @@ def usage():
             output. Passing 'value' that begins with an 'f' 
             or 'F' will evaluate to False. All else evaluates to True.
             (set to: {TRANSMIT_TIMING_LOG})
+
+        -P value, --profiling value
```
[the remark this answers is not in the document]
ndp-opendap: 👍

ndp-opendap (approved):
Nice work Hannah!!

:shipit:

modules/dmrpp_module/SuperChunk.cc
ndp-opendap: 👍

modules/dmrpp_module/SuperChunk.cc
ndp-opendap: 👍

jgallagher59701 (approved):
+1 for putting the BESStopWatch object inside the macro!

Ready to merge.

PR #1144, opened 2025-08-22
Rename `get_chunks_size` to `get_chunk_count`

hannahilea:
Rename for clarity, prompted by addition of similar function in #1141.


ndp-opendap (approved):
Nice. I feel like this improves the semantics of the method name. The old method name, `get_chunks_size()`, seems like it would return total size of all of the chunks. 

:shipit:

Issue #1145, opened 2025-09-24
Reduce the CICD detritus in our S3 buckets, without doing too much work.


jgallagher59701:
We have an S3 bucket that our CICD scripts use for interim build products. I would like to limit the number of those files in the bucket, deleting older files as newer ones are added. What automatic options for that exist? This does not have to be perfect, it just needs to reduce the clutter that builds up from processes that are automated. For example, deleting any file older than three months or deleting all but the newest 5 copies of a file (object) matching a given prefix. Please suggest some ways we could do this. 

The response:

Great problem to automate away. Here are a few solid, low-maintenance ways to keep your “temp builds” bucket tidy—ranging from zero code to a tiny Lambda.

# 1) S3 Lifecycle rules (zero code)

**Best for “delete anything older than N days/months.”**
Create a lifecycle rule that targets your CI artifacts by **prefix** (e.g., `builds/`) and/or **tag** (e.g., `purpose=ci-temp`):

* **Expire current objects after N days** (e.g., 90 days).
* **Abort incomplete multipart uploads** after a short time (e.g., 1–3 days) to avoid orphaned parts.
* If you use **versioning**, you can also:

  * **Expire noncurrent versions after N days**.
  * **Limit the number of noncurrent versions** (keep only the newest K noncurrent versions).

**Example lifecycle JSON (age-based delete, no versioning):**

```json
{
  "Rules": [
    {
      "ID": "expire-ci-builds-90d",
      "Status": "Enabled",
      "Filter": { "Prefix": "builds/" },
      "Expiration": { "Days": 90 },
      "AbortIncompleteMultipartUpload": { "DaysAfterInitiation": 3 }
    }
  ]
}
```

**Why this is nice:** dead simple, free, and completely automatic.
**Caveat:** S3 lifecycle can’t “keep newest 5 *objects* by prefix” (it’s time-based). Version limits only help if you’re uploading new **versions of the same key**, not new keys like `build-123.tgz`, `build-124.tgz`, etc.

---

# 2) Event-driven cleanup Lambda (count-based retention per prefix)

**Best for “keep only the newest K per prefix/branch.”**
When a new object is added, trigger a Lambda to prune older siblings:

* **Trigger:** S3 Event Notification on `ObjectCreated:*` for `builds/` (or any prefix).
* **Lambda logic (boto3):**

  1. Identify the object’s **logical group** (e.g., prefix `builds/main/` or `builds/feature-x/`).
[...]

PR #1149, opened 2025-09-26
bes-1145: Added simple cleaning of old test data

jgallagher59701:
## Description

**Reference ticket:** bes issue 1145 (https://github.com/OPENDAP/bes/issues/1145)

We have an S3 bucket that our CICD scripts use for interim build products (opendap.travis.tests). Add a S3 lifecycle rule that will delete the tar balls holding autotest results from CICD builds when they become 30 days old. 

The idea here is to get some practice with this because it might matter a bit more with the bucket that holds our builds, but also, this ...tests bucket is full of junk and there's no reason for that.

## Tasks

- [ x] Ticket exists and is linked in title
- [n/a ] Tests added/updated
- [n/a ] Dead code removed
- [ x] No TODOs added

### DIFF
--- a/travis/README.md
+++ b/travis/README.md
@@ -12,7 +12,7 @@ where for us YOUR-BUCKET is ```opendap.travis.tests``` (see upload-test-results.
 
 ## Backstory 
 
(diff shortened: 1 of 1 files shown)


hannahilea (approved):
🎉 

Nonblocking---see note to consider optionally removing part of the README.md.

PR #1151, opened 2025-09-30
[HYRAX-1883] DMR++ threading bug

jgallagher59701:
## Description

Aaaaaaand in the DmrppRequestHandler.cc i see that the defaults for both schemes is true:

bool DmrppRequestHandler::d_use_transfer_threads = true;
unsigned int DmrppRequestHandler::d_max_transfer_threads = 8;

bool DmrppRequestHandler::d_use_compute_threads = true;
unsigned int DmrppRequestHandler::d_max_compute_threads = 8;

Fix this - change DmrppRequestHandler::d_use_transfer_threads so it is false by default.

TODO

## Tasks

- [x] Tests added/updated
- [x] Dead code removed
- [x] No TODOs added


hannahilea (approved):
Does what the task says to do! 

A few follow-up questions:
- Is it worth throwing in a warning here, where if it a user has enabled multithreading here (the non-default behavior) we show a warning about what might happen, and link to the ticket about the buggy behavior?
- ...do we already have that follow-up ticket for the buggy behavior that needs fixing (and which prompted this default change)? 
- Does this change in default behavior need to be communicated to any users (ngap/otherwise)? 
- Is there any documentation of threading behavior that needs updating in line with this change?

jgallagher59701:
> Does what the task says to do!
> 
> A few follow-up questions:
> 
> * Is it worth throwing in a warning here, where if it a user has enabled multithreading here (the non-default behavior) we show a warning about what might happen, and link to the ticket about the buggy behavior?

Hmmm. Where would we put that warning? I suppose it could go in a log message.

> * ...do we already have that follow-up ticket for the buggy behavior that needs fixing (and which prompted this default change)?

Yes, we have an outstanding ticket to fix this mess. It's a big job

> * Does this change in default behavior need to be communicated to any users (ngap/otherwise)?

Not the default behavior for NGAP, because we override the default with a site.conf file. But what if that site.conf file was edited ... then the default behavior would creep back in. So this is more of an insurance policy.

> * Is there any documentation of threading behavior that needs updating in line with this change?

No.

PR #1161, opened 2025-10-08
[HYRAX-1891] Test AWS C++ SDK pre-signed url behavior from bes

hannahilea:
## Description

Relying on pre-installed AWS C++ SDK, test that presigned urls can be generated and function as expected.

Requires presence of AWS SDK, i.e. as installed via through hyrax-dependencies. Because of this, blocked on HYRAX-1894 (https://github.com/OPENDAP/hyrax-dependencies/pull/73).


hannahilea:
Tested locally as well, aws tests pass! 🎉

PR #1171, opened 2025-10-14
[HYRAX-1900] Various AWS/S3 fixs

jgallagher59701:
## Description

**Reference ticket:** HYRAX-1900

Various AWS/S3 tests in the DMR++ module fail when the CMAC_* env vars are not set
exactly correct. I added a new env variable 'CMAC_ON' that has to be set and has to be
the value "yes". If those conditions are met, the various tests are run. If they are met but
the old tests, which look for a value for CMAC_ID are not met, the tests still (mostly) 
don't run (as before). But CMAC_ON switches all of the tests.

Unlike other PRs that fix something or add a new feature to the server, this alters the 
build/test process. Is it an improvement? 

Back story, I found a number of tests were failing on my system, Turning off CMAC_ID
didn't fix all the failures because some tests rely in a script that probes the aws cli and
still others need to check the 'bucket url.' I tried to make this a comprehensive change,
But let's think it over before merging (and also see how CICD reacts).

## Tasks

- [x] Tests added/updated
- [x] Dead code removed
- [x] No TODOs added

### DIFF
--- a/modules/dmrpp_module/ngap_container/unit-tests/NgapOwnedContainerTest.cc
+++ b/modules/dmrpp_module/ngap_container/unit-tests/NgapOwnedContainerTest.cc
@@ -333,7 +333,6 @@ public:
 
         string dmrpp; // = container.access();
         CPPUNIT_ASSERT_NO_THROW_MESSAGE("This should not throw an exception", dmrpp = container.access());
-        //string dmrpp = container.access();
         DBG2(cerr << "DMR++: " << dmrpp << '\n');
         CPPUNIT_ASSERT_MESSAGE("The response should not be empty", !dmrpp.empty());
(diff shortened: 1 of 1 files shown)


modules/dmrpp_module/ngap_container/unit-tests/NgapOwnedContainerTest.cc
```diff
@@ -329,7 +331,9 @@ class NgapOwnedContainerTest: public CppUnit::TestFixture {
         // Set the location of the data as a file:// URL for this test.
         container.set_data_source_location(DMRPP_TEST_BUCKET_OPENDAP_AWS);
 
-        string dmrpp = container.access();
+        string dmrpp; // = container.access();
+        CPPUNIT_ASSERT_NO_THROW_MESSAGE("This should not throw an exception", dmrpp = container.access());
+        //string dmrpp = container.access();
```
hannahilea:
```suggestion
```

PR #1174, opened 2025-10-16
[HYRAX-1732] Extract s3 credentials urls from CMR granule

hannahilea:
Previously, we pulled only the http/s url from the CMR granule's json document. Now we pull three URLs: the original one, the s3credentials endpoint, and the s3 data url.

We don't do anything with them yet; this PR purely adds and tests functionality for accessing them.

### DIFF
--- a/modules/dmrpp_module/ngap_container/NgapApi.cc
+++ b/modules/dmrpp_module/ngap_container/NgapApi.cc
@@ -306,7 +306,7 @@ string NgapApi::build_cmr_query_url(const string &restified_path) {
     return cmr_url;
 }
 
-string get_data_access_url(rapidjson::Value &obj) {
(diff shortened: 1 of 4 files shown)


modules/dmrpp_module/ngap_container/NgapApi.cc
```diff
@@ -338,24 +338,93 @@ string get_data_access_url(rapidjson::Value &obj) {
     return {""};
 }
 
+
+string get_data_s3_url(rapidjson::Value &obj) {
```
hannahilea: there is a little code duplication here, but not enough that it felt worth the time to refactor and make shared---especially since we'll likely be ripping out one of the three of these duplicate-ish functions in the near future anyway.

PR #1176, opened 2025-10-16
[HYRAX-1904] I am trying Ubuntu jammy to get Java 17

jgallagher59701:
## Description

**Reference ticket:** HYRAX-1904

This will stop us from having to load java 17 as a separate step, which is often the cause for failed CICD builds. Also, I have set CMAC_ON=yes which is needed after a recent branch merge that introduced this feature. The CICD builds _do_ run the S3 tests, so that switch should be set.

## Tasks

- [x] Tests added/updated - N/A
- [x] Dead code removed
- [x] No TODOs added

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,6 +44,7 @@ addons:
       - gcovr
       - grep
       - libfl-dev   # needed for focal, but not xenial. jhrg 1/18/24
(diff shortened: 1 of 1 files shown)


ndp-opendap (approved):
Jammy it is then Captn'

:shipit:

PR #1183, opened 2025-10-29
[HYRAX-1732] Add SignedUrlCache and cache contents of s3credential endpoints

hannahilea:
## Description

- Set up and test SignedUrlCache, which is similar in structure to EffectiveUrlCache
- When BES is in-region (us-west-2), access chunk data via direct s3 access with a presigned url (signed with STS credentials pulled from TEA endpoint)
- STS credentials are cached, as they are valid for ~an hour; this means that we make TEA requests once per hour per DAAC endpoint, instead of once per granule path (!)

## Tasks

- [x] Tests added/updated
- [x] Dead code removed
- [x] No TODOs added (well, one! but intentionally b/c that HAS to happen next)
- [x] Linking works on all Travis builds
- [x] End-to-end tests added (and run on travis)
- [x] Tested on SIT
  - [x] More extensive testing passes on SIT! :D (hyrax500, etc)
- [x] Undo temporary Travis changes


hannahilea:
There is a fair amount of code added; probably the easiest way to understand it is to start by reading through the test page, since that is fairly comprehensive (I think).

modules/dmrpp_module/ngap_container/NgapOwnedContainer.cc
```diff
@@ -559,6 +568,49 @@ bool NgapOwnedContainer::get_dmrpp_from_cache_or_remote_source(string &dmrpp_str
...
+
+// TODO-docstring
+// Implementation borrowed from DMZ::process_dataset
+NgapApi::DataAccessUrls NgapOwnedContainer::extract_s3_data_urls_from_dmrpp(const string &dmrpp_string) {
+    // If the dmrpp is invalid, we will have hit a failure before now---so we can assume
+    // it's generally safe---so do basically no additional safety checking
+
+    // Load the xml document
+    pugi::xml_document result_xml_doc;
+    pugi::xml_parse_result result = result_xml_doc.load_string(dmrpp_string.c_str(), pugi::parse_default | pugi::parse_ws_pcdata_single);
```
ndp-opendap: The variable `result` is never checked. Does that matter? Do we care??

ndp-opendap (approved):
Looks great!
I had one question, but I think it's ready to go.
N

:shipit:

jgallagher59701 (approved):
I have two cautionary comments and Nathan asks about an unused return value,, but on balance, I think it's ready to merge.

hannahilea:
Thanks for the eyes on this mid-point! I'm going to convert it back to a draft and merge my aws-specific branch into it, and then do my SIT testing here. Will re-open for re-review after that testing is complete!

modules/dmrpp_module/Chunk.cc
```diff
@@ -1359,14 +1362,20 @@ std::shared_ptr<http::url> Chunk::get_data_url() const {
...
+
+    std::shared_ptr<http::EffectiveUrl> url = SignedUrlCache::TheCache()->get_signed_url(d_data_url);
+
+    if (url == nullptr) {
+        INFO_LOG(prolog + "Failed to locally sign url; constructing effective_url via redirects.");
+        url = EffectiveUrlCache::TheCache()->get_effective_url(d_data_url);
+    }
+    BESDEBUG(MODULE, prolog << "Using data_url: " << url->str() << endl);
 
 #if ENABLE_TRACKING_QUERY_PARAMETER
```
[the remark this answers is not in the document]
hannahilea: filed as #1199

modules/dmrpp_module/Chunk.cc
```diff
@@ -1359,14 +1362,20 @@ std::shared_ptr<http::url> Chunk::get_data_url() const {
     // The d_data_url may be nullptr(fillvalue case). 
     if (d_data_url == nullptr) 
         return d_data_url;
-    std::shared_ptr<http::EffectiveUrl> effective_url = EffectiveUrlCache::TheCache()->get_effective_url(d_data_url);
-    BESDEBUG(MODULE, prolog << "Using data_url: " << effective_url->str() << endl);
+
+    std::shared_ptr<http::EffectiveUrl> url = SignedUrlCache::TheCache()->get_signed_url(d_data_url);
```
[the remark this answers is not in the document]
hannahilea: They are, yeah! That's why I'm returning EffectiveUrls here, rather than making a new SignedUrl class. (Returning EffectiveUrls also makes it clean to (1) attempt to get an EffectiveUrl via the SignedUrlCache and then (2) fall back to using the EffectiveUrlCache if we fail, as a safety-net until we're REALLY confident in the new signing code in production.)

modules/dmrpp_module/ngap_container/NgapOwnedContainer.cc
hannahilea: Weird, I was *sure* that I already responded to this comment somewhere. Apologies for the duplication if I'm just missing it!

Anyway, yes: I'll add a check here, that's a smart thing to do! Also, it would be fairly unlikely to fail at this point, as this is the second time we've parsed this same dmrpp_string, and upon failure the earlier time we'd be blocked from ever reaching this point. 

- [ ] Check result here

modules/dmrpp_module/Chunk.cc
```diff
@@ -1359,14 +1362,20 @@ std::shared_ptr<http::url> Chunk::get_data_url() const {
...
+
+    if (url == nullptr) {
+        INFO_LOG(prolog + "Failed to locally sign url; constructing effective_url via redirects.");
```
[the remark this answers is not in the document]
hannahilea: (and for non-ngap, we know they'd NEVER get self-signing following this method, this is extremely ngap-specific)

Makefile.am
```diff
@@ -29,11 +29,7 @@ AM_CXXFLAGS=
 AM_LDFLAGS =
 include $(top_srcdir)/coverage.mk
 
-SUBDIRS = dispatch xmlcommand ppt server http cmdln standalone docs bin templates hello_world
-
-if USE_AWS_SDK
-SUBDIRS += aws
-endif
+SUBDIRS = dispatch xmlcommand ppt server http cmdln aws standalone docs bin templates hello_world
```
hannahilea: note: because aws is now required for standalone, i think we need to make it non-optional in the rest of the configuration. if this is *not* true, and there's a better approach, lmk---otherwise I'll make that non-optional change here.

- [ ] Get rid of USE_AWS_SDK all over, as aws is no longer optional

hannahilea:
Oooookay. @jgallagher59701 and @ndp-opendap, you both reviewed (and approved) an earlier checkpoint of this PR, but I wanted to request re-review because it is hefty and a fair amount has changed since then.

In terms of specific functionality, all of the new functions have good unit test coverage, so I'm not particularly worried about the micro-level. I do want to make sure that we're all on the same page about what's happening at the macro level! 

Note that I haven't removed any of the original logic for how data chunks are retrieved; I've simply inserted an optional bypass of the redirect-to-and-back-from-tea portion, such that if we're in-region and able to generate a presigned URL for direct access, we do so. If we fail to generate one for any reason (missing or bad STS credentials, etc), we fall back to the original TEA-redirect pathway.

ndp-opendap (approved):
Honestly this is a little overwhelming for me. That's in no way a criticism of the excellent work in this PR, but a statement about my own limitations. I'm super glad/relieved/pleased that this months long effort is finally arriving at The Show! @hannahilea - you have my appreciation and gratitude for grinding out this change! What a road it's been!

:shipit:

aws/SignedUrlCache.h
```diff
@@ -0,0 +1,124 @@
+// -*- mode: c++; c-basic-offset:4 -*-
+
+// This file is part of the BES aws package, part of the Hyrax data server.
+
+// Copyright (c) 2025 OPeNDAP, Inc.
+// Authors: Nathan Potter <ndp@opendap.org>, Hannah Robertson <hrobertson@opendap.org>
```
jgallagher59701: Unless Nathan wrote this, you should change the copyright header.

dap/BESDASResponseHandler.cc
```diff
@@ -85,7 +85,9 @@ BESDASResponseHandler::execute( BESDataHandlerInterface &dhi )
     GlobalMetadataStore::MDSReadLock lock;
 
     dhi.first_container();
-    if (mds) lock = mds->is_das_available(*(dhi.container));
+    if (mds != nullptr) {
+        lock = mds->is_das_available(*(dhi.container));
```
jgallagher59701: I think that *(dhi.container) is the reference to a potentially null pointer sonarscan is talking about.

modules/dmrpp_module/build_dmrpp_h4/Makefile.am
```diff
@@ -14,13 +14,28 @@ M_VER = 0.1.0
 
 AM_CPPFLAGS = $(HDF4_CFLAGS) $(HDFEOS2_CPPFLAGS) -I$(top_srcdir) -I$(top_srcdir)/dispatch -I$(top_srcdir)/modules/dmrpp_module \
     -I$(top_srcdir)/pugixml/src -I$(top_srcdir)/http -I$(top_srcdir)/modules/dmrpp_module/ngap_container -I$(top_srcdir)/dap \
-    $(DAP_CFLAGS)
+    $(DAP_CFLAGS) -I$(top_srcdir)/aws
 
 AM_CPPFLAGS += -DMODULE_NAME=\"$(M_NAME)\" -DMODULE_VERSION=\"$(M_VER)\"
 
 AM_CXXFLAGS=
```
jgallagher59701, line 21: This line should be `AM_CXXFLAGS = $(CXXFLAGS) or $(AM_CXXFLAGS)`. Have you merged master to this branch lately?

jgallagher59701 (approved):
I have some comments, but it can be merged. Most of my comments are not a big deal and they can be additional tickets or we can say I'm just being too pedantic ;-)

dap/BESDASResponseHandler.cc
hannahilea: oh weird, this was not something i intended to touch in this PR! reverting.

modules/dmrpp_module/build_dmrpp_h4/Makefile.am
hannahilea: ...I have, it is up to date with master. Looks like this is also the current state on master:
https://github.com/OPENDAP/bes/blob/dd93c031165bab2f277fa0347f12d5c62ee3b104/modules/dmrpp_module/build_dmrpp_h4/Makefile.am#L21

hannahilea: I can set it to $(CXXFLAGS) or $(AM_CXXFLAGS) here if you'd like, though.

jgallagher59701: Probably a wash - I intend to refactor the Makefile.am files someday soon. It's OK to leave it or edit it. Your call.

PR #1186, opened 2025-11-03
HYRAX-1916, make the AIRS CO2 product's lat/lon correct. Also reducin…

kyang2014:
…g the calculation time of HDF-EOS2's Geographic projection lat/lon by not using the HDF-EOS2 library. This also avoids the wrong values generated by the HDF-EOS2 library for some cases.

## Description

**Reference ticket:** HYRAX-1916, HYRAX-1925,HYRAX-1926,HYRAX-1915

## Tasks

- [x] Tests added/updated
- [x] No TODOs added


kyang2014:
> I see that the new code adds a lot of complexity to the build_dmrpp_util_h4.cc and HDFDMRArray_EOS2LL.cc files. Could some of that new code be factored out into other functions? If that is gong to be time consuming, given that I just approved three Critical bug tickets, maybe make a ticket to go back make that change later when there's less time pressure? But, that's an improvement. This can be merged as it is.
> 
> Thanks.

The complexity is necessary because this is a localized fix for a specific product and bug. The logic is straightforward when coding this way and it makes the developer who understands the issue easier to follow logically.

PR #1189, opened 2025-11-04
HYRAX-1913, make HDF5 handler compile with HDF5 1.14.6.

kyang2014:
## Description

**Reference ticket:** HYRAX-1913

We need to update the HDF5 handler if we want to use the HDF5 1.14.6 in the hyrax dependencies.
In this ticket, I update the HDF5 handler so that it can work with both with the current hyrax dependencies(HDF5 1.10.10) and HDF5 1.14.6. 

## Tasks

- [x] No TODOs added


jgallagher59701 (approved):
Ready to merge

PR #1198, opened 2025-11-14
Jhrg/hyrax 1898 cmr client id fix 2

jgallagher59701:
## Description

**Reference ticket:** HYRAX-1898

Remove the CMR 'client_id' from the query string and add it to the CMR Client-Id request header.

This might be a good time to revisit HYRAX-1036, since these changes are so close to the old ECHO token code.

## Tasks

- [x] Tests added/updated
- [x] Dead code removed
- [ ] No TODOs added


jgallagher59701:
Note that I added TODOs about HYRAX-1036

PR #1203, opened 2025-11-20
NO MERGE: test branch for bisecting test failures (that only happen on travis)

hannahilea:
to assist in debugging #1183


hannahilea:
success! my work here is done.

(the approach of "have it print out the failed test log on output" worked pretty well, though.)

PR #1211, opened 2025-12-04
Hyrax 1957 test

kyang2014:
## Description

**Reference ticket:** HYRAX-1957
1. Make the buffer chunk support extend to the case when the first and the second chunks are NOT adjacent.
2. Try to reduce the buffer size by calculating the optimal buffer size for each buffer.
3. Extend the use of buffer chunk to the direct IO case.
4. Add many tests.

TODO

## Tasks

- [x] Tests added/updated
- [x] Dead code removed

### DIFF
--- a/modules/dmrpp_module/DmrppArray.cc
+++ b/modules/dmrpp_module/DmrppArray.cc
@@ -684,7 +684,7 @@ unsigned long long DmrppArray::get_size(bool constrained) {
 
 unsigned long long DmrppArray::get_maximum_constrained_buffer_nelmts()
 {
-    // The stride here doesn't matter since we need to obtain the maximum contrained buffer. KY 2025-12-08
+    // The stride here doesn't matter since we need to obtain the maximum constrained buffer. KY 2025-12-08
     unsigned long long asize = 1;
     for (Dim_iter dim = dim_begin(), end = dim_end(); dim != end; dim++) {
         int64_t start = dimension_start_ll(dim,true);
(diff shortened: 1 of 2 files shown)


modules/dmrpp_module/DmrppArray.cc
```diff
@@ -2021,51 +2129,51 @@ void DmrppArray::read_buffer_chunks() {
...
-    // to loop through all the needed chunks for the constraint case.
-    bool first_needed_chunk = true;
-    unsigned long long first_needed_chunk_offset = 0;
-    unsigned long long first_needed_chunk_size = 0;
-    for (const auto &chunk : get_immutable_chunks()) {
+    // The maximum buffer size is limited to the current constrained domain.
+    unsigned long long max_buffer_size = bytes_per_element * this->get_maximum_constrained_buffer_nelmts();
+
+    // TODO: Sometimes the constraint size is too small and the constraint is across quite a few chunks. We may need
+    // to enlarge the max_buffer_size to 4K or even 1M.
```
jgallagher59701: 4K Seems really small. Even 1M doesn't seem that large for an S3 transfer. Should we talk about this?

jgallagher59701 (approved):
I have one question about a comment, but it's not a high priority issue for this PR, just for my understanding.

I think it might be useful to look at the sonarscan comments about complexity. Take its comments/suggestions with a grain of salt - sometimes it's better to keep one function that sonar scan says is 'too big' than to break it apart. But, sometimes it makes good suggestions.

modules/dmrpp_module/DmrppArray.cc
kyang2014: We can.

PR #1218, opened 2026-01-08
Jhrg/hyrax 1963 bump the dmr version

jgallagher59701:
## Description

**Reference ticket:** HYRAX-1963

Update the tests' baselines to work with the new libdap version that sets the dmrVersion to 2.0.

## Tasks

- [x] Tests added/updated
- [x] Dead code removed
- [x] No TODOs added


ndp-opendap (approved):
Are there baseline changes that reflect the changes in serialization order in addition to the changes in version??

jgallagher59701:
> Are there baseline changes that reflect the changes in serialization order in addition to the changes in version??

Yes. That is, the only changes were to the version number. The responses for our tests didn't change because we didn't/don't have tests in the BES that exercise datasets with child groups. That seems odd, but in libdap we didn't write any, only tests with _only_ child groups. In the BES, all our tests came from data files and those only recently started mixing both child groups and variables in the root group.

hannahilea (approved):
Approach makes sense/what's here looks good; if I understand correctly that we don't have any tests in bes that demonstrate the new behavior (and that would have failed without it), I think it's worth adding one---but it's up to you whether to block this PR on it. (Also, I may have misunderstood the above conversation; disregard if so!)

I do think it's a shame that we need to change all of the versions, given that technically the behavior demonstrated here matches v1.0 behavior as well, but I agree that given the way tests are currently set up, updating all baseline version numbers makes sense as the fastest way to get the new features out to users.

In the future, it might be a slightly nicer to do something like leave the existing dmrVersion in the baseline from when it was generated and then do one of our sed/awk substitutions to replace `dmrVersion="x.y"` with `dmrVersion=removed` before baseline comparison, as we do for version numbers and dates.

Issue #1223, opened 2026-01-14
Refactor REMOVE_VERSIONS m4 macros


ndp-opendap:
# Overview

This is a ticket to refactor our REMOVE_VERSIONS m4 macro and make the associated changes to the baseline files.

It currently has two behaviors we want to alter/add:

1. When we replace the `dmrpp:version` attribute and value we do so in a way that makes the resulting XML file incorrectly formed. This should be fixed.
2. The value of the `dmrVersion` attribute recently changed from 1.0 to 2.0. All of the dmr/dap baselines were changed to 2.0. Rather than doing that we should also add a replacement clause to remove the `dmrVersion` value.

**Corollary**: We might also consider making this operation a bash script saved as `bes/bin/remove_versions` and utilizing it from there rather than have it's programmatic expression be in multiple m4 file. Single Source of Truth and all that. This would also allow us to have a more understandable invocation without the m4 `[ ]` shenanigans,  and we could write unit tests as functions in the script file. Just sayin...

Current macro:
```bash
m4_define([REMOVE_VERSIONS], [dnl
  awk '{
    gsub(/<Value>[[0-9]+]\.[[0-9]+]\.[[0-9]+](-[[0-9]+])?<\/Value>/, "<Value>removed version</Value>");
    gsub(/<Value>[[a-zA-Z._]+]-[[0-9]+]\.[[0-9]+]\.[[0-9]+](-[[0-9]+])?<\/Value>/, "<Value>removed version</Value>");
    gsub(/dmrpp:version="[[0-9]+]\.[[0-9]+]\.[[0-9]+](-[[0-9]+])?"/, "removed dmrpp:version");
    print
  }' < $1 > $1.awk
  mv $1.awk $1
])
```

## Item 1
Change the current `dmrpp:version` removal to one that doesn't produce incorrectly formed XML as the result 
```bash
    gsub(/dmrpp:version="[[0-9]+]\.[[0-9]+]\.[[0-9]+](-[[0-9]+])?"/, "removed dmrpp:version");
```
Would become:
```bash
    gsub(/dmrpp:version="[[0-9]+]\.[[0-9]+]\.[[0-9]+](-[[0-9]+])?"/, "dmrpp:version= \"removed\"");
```

## Item 2

Add support for removing the `dmrVersion` attribute value:
```bash
[...]

PR #1225, opened 2026-01-14
BES-1223: Strip dmrVersion attribute value out during tests

hannahilea:
## Description

Subtask of https://github.com/OPENDAP/bes/issues/1223 - remove dmrVersion attribute value during dmrpp_module tests.

## Tasks

- [x] Tests added/updated
- [x] Dead code removed
- [x] No TODOs added


hannahilea:
When reviewing, I'd recommend turning off the .baseline changes to see the functional changes---given how many .baseline changes there are. E.g., 
image

ndp-opendap:
I have concerns about changing this:
```
xmlns:dmrpp="http://xml.opendap.org/dap/dmrpp/1.0.0#
```
to this:
```
xmlns:dmrpp="http://xml.opendap.org/dap/dmrpp/removed_version#
```
Some thoughts:
Down the road, make it difficult to understand why things are different. If we do adopt a new namespace name I think we would not discard support for the older namespace and maybe new tests for a later version of the namespace would be in order.

And I think this is different than Jame's `dmrVersion` change - that was the result a bug and we aren't ever going back to the previous `dmrVersion`

ndp-opendap (commented):
I think with the namespace name rewind that this is good to go. 

In my eternal quest for a single source of truth I would like to ask you and @jgallagher59701 to consider if we might want look creating a `remove_versions` shell script to place in bes/bin , and use it in the definitions of the REMOVE_VERSION macros.

ndp-opendap (approved):

PR #1227, opened 2026-01-16
HYRAX-1993, handle the default option dmr generation  when a variable…

kyang2014:
… has two unlimited dimensions. Previously we throw an error. Since we found such a case in a NASA file, now we just make the dmr not have the unlimited dimension when we find that a variable has two unlimited dimensions.

## Description

**Reference ticket:** HYRAX-1993

## Tasks

- [x] Tests added/updated
- [x] Dead code removed
- [x] No TODOs added


jgallagher59701 (approved):
Thanks!

PR #1228, opened 2026-01-20
BES-1181: Skip instead of fail `aws` tests when no credentials provided

### LINKED ISSUE #1181: In `AWS_SDK_Test`, skip instead of fail tests if AWS credentials are not available

hannahilea:
## Description

Fix #1181.

## Tasks
- [x] Confirm that Travis tests still run unit tests. (Well, it says that unit tests pass...and I've confirmed locally that it correctly only says that and does NOT have the warning about credentials if the credentials actually exist)


ndp-opendap (approved):
+1 for conditional tests. !woot!

PR #1234, opened 2026-01-23
BES-1236: Add gha workflow for building on mac

### LINKED ISSUE #1236: Build bes in CI/CD on MacOS15 intel and m-chip
We need CI/CD running on machines that match our dev environments, so that we (a) always know that the bes can build on those machines, (b) we don't inadvertently break dev workflows for teammates on machines different than our own, and (c) we have a known test state (ideally all passing!) for the architectures we use to develop, so that it is easy to know what test failures are due to intentional changes as we develop new features. 

We can do this through GHA, as we recently did for hyrax-dependencies: https://github.com/OPENDAP/hyrax-dependencies/pull/82

hannahilea:
## Description

Fix #1236. 

Add GHA CI job to build and run tests on mac (m4, intel), to try to maintain good repo state for local development.

~~Tests themselves will fail until #1229 is merged. EDIT: Complete~~

### DIFF
--- a/.github/workflows/.build-osx.yml
+++ b/.github/workflows/.build-osx.yml
@@ -25,8 +25,8 @@ jobs:
     runs-on: ${{ matrix.runs-on }}
     env:
       GDAL_OPTION: "--without-gdal"
-      CMAC_ID: "${ secrets.AWS_GHA_BES_ID }"
-      CMAC_ACCESS_KEY: "${ secrets.AWS_GHA_BES_SECRET_KEY }"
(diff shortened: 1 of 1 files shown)


jgallagher59701 (approved):
If this works, it's a huge win. Merge. We can make tickets for GDAL and libdap and put them in the IP sprint. Pull them forward if we need a moral boost later on (the libdap one at least, GDAL builds may never be a moral boost, unless we live in very dark times).

PR #1252, opened 2026-02-04
Fix the AT_BESCMD_BINARY_DAP4_RESPONSE_TEST test macro for OSX Intel and Apple silicon

jgallagher59701:
## Description

**Reference ticket:** HYRAX-2009

Fix the AT_BESCMD_BINARY_DAP4_RESPONSE_TEST test macro for OSX Intel and Apple silicon

## Tasks

- [x] Tests added/updated
- [x] Dead code removed
- [x] No TODOs added

### DIFF
--- a/modules/common/handler_tests_macros.m4
+++ b/modules/common/handler_tests_macros.m4
@@ -300,21 +300,10 @@ m4_define([AT_BESCMD_BINARY_DAP2_RESPONSE_TEST],  [dnl
         [
         AT_CHECK([besstandalone -c $abs_builddir/$bes_conf -i $input | getdap -Ms -], [0], [stdout])
 
-        AS_IF([[ test -f $baseline && test -f $baseline.m_proc ]],
-            [
-                # If both baselines exist, hold off on printing any diffs until knowing that neither is a match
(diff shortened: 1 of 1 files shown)


modules/common/handler_tests_macros.m4
```diff
@@ -357,21 +357,13 @@ m4_define([AT_BESCMD_BINARY_DAP4_RESPONSE_TEST],  [dnl
...
-                ])
-            ],
-            [
-                AT_CHECK([diff -b -B $baseline stdout])
-            ])
-        ])
+        # if the tests are running on Apple silicon, use the *.mproc baseline,
+        # otherwise use the plain baseline. jhrg 2/3/26
+        echo "uname -m: $(uname -m)"
+        AS_IF([ test "x$(uname -m)" = "xarm64" -a -f $baseline.m_proc],
```
hannahilea: Out of curiosity, because I've also seen it elsewhere in our codebase: what is the reason for prefixing the various comparison strings with an x? e.g., why 
```
        "x$(uname -m)" = "xarm64"
```
instead of 
```
        "$(uname -m)" = "arm64"
```
?

PR #1256, opened 2026-02-16
HYRAX-2017, make gen_dmrpp_side_car support the HDF4 file without fil…

kyang2014:
…e name extension.
Just makes any file that doesn't have the file name extension as .dmpp be handled by the HDF4 handler. Since we distinguish the handling of HDF4 files and HDF5 files with an explicit option, this will not be a problem.
## Description

**Reference ticket:** HYRAX-2017

## Tasks

- [x] Dead code removed
- [x] No TODOs added


dh-opendap (approved):
Simple enough change, approved.

PR #1267, opened 2026-03-03
HYRAX-2029, check if a dimension is  unlimited to decide if can do th…

kyang2014:
…e direct IO

## Description

**Reference ticket:** HYRAX-2029

When a dimension is unlimited and the dimension size is smaller than the chunk size, this ticket makes direct chunk IO still apply to this variable. We already have a test that covers this case.  ncdump data and header are the same.  We can see the new file keep the original chunk and compression level with the HDF5 tool. 

## Tasks

- [x] Dead code removed
- [x] No TODOs added


jgallagher59701 (approved):
Thanks!

PR #1295, opened 2026-04-01
Add GDAL back into the Travis CICD build.

jgallagher59701:
## Description

**Reference ticket:** HYRAX-1968, Also, appears to fix 

- [x] HYRAX-2087

Turn on GDAL in the CICD tasks

For this PR, I really wanted to see how much some simple changes could speed up a CICD build. The answer is, somewhat, but it's not huge. We can think about taking what I've done here and moving them into the BES builds at a later date. I added two OSX builds - both use cmake - and one fails in a way that indicates something that I think needs its own ticket. 

I suggest that we merge this PR without addressing that because it seems pretty far removed from the goal of this PR.

## Tasks

- [x] Tests added/updated
- [x] Dead code removed
- [x] No TODOs added

### DIFF
--- a/cmdln/tests/Makefile.am
+++ b/cmdln/tests/Makefile.am
@@ -23,11 +23,9 @@ clean-local:
 	test ! -f '$(TESTSUITE)' || $(SHELL) '$(TESTSUITE)' --clean
 	-rm -f $(TESTSUITE) $(srcdir)/package.m4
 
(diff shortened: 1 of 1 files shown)


hannahilea:
ok, makes sense. i'll approve, although i am a bit hesitant about the lack of synchronicity (since that's what gets us into confusion around things failing to build/test locally when they succeed in CI...). but as long as it'll be tackled promptly, lgtm!

PR #1301, opened 2026-04-03
Jhrg/hyrax 2012 re enable gdal

jgallagher59701:
## Description

**Reference ticket:** HYRAX-2012

The idea here is to get all of the production builds to use GDAL, given that the simpler initial Travis check and distcheck builds do.

## Tasks

- [x] Tests added/updated
- [x] Dead code removed
- [x] No TODOs added


modules/gdal_module/Makefile.am
```diff
@@ -42,7 +41,7 @@ GDAL_SRCS = GDALModule.cc GDALRequestHandler.cc
 GDAL_HDRS = GDALModule.h GDALRequestHandler.h
 
 libgdal_module_la_SOURCES = $(GDAL_SRCS) $(GDAL_HDRS)
-libgdal_module_la_LDFLAGS = $(AM_LDFLAGS) -avoid-version -module $(GDAL_LDFLAGS)
+libgdal_module_la_LDFLAGS = -avoid-version -module -L$(builddir)/reader -L$(builddir)/writer $(AM_LDFLAGS) $(GDAL_LDFLAGS)
```
hannahilea, line 44: huh, where does $AM_LDFLAGS come from, since it was removed from line 18?

.travis.yml
```diff
@@ -51,14 +52,13 @@ git:
...
-    # Should the BES build with GDAL? This is here to round up build errors. Set this
-    # to the 'configure' script option to use (--with-gdal) or not use (--without-gdal).
-    # jhrg 12/2/25
-    # - GDAL_OPTION=--without-gdal
-    - GDAL_OPTION=
     # NB: This value of $prefix must be shared between the hyrax-deps, libdap and bes builds.
     - prefix=$HOME/install
     - PATH=$prefix/bin:$prefix/deps/bin:$PATH
+    # Should the BES build with GDAL? This is here to round up build errors. Set this
+    # to the 'configure' script option to use (--with-gdal) or not use (--without-gdal).
```
hannahilea:
```suggestion
    # Should the BES build with the NCML handler? This is here to round up build errors. Set this
    # to the 'configure' script option (--disable-ncml) to disable it.
```

AGENTS.md
```diff
@@ -103,3 +103,5 @@ When asked to review, prioritize:
 - State assumptions and environment details explicitly, especially configure flags and dependency locations.
 - If full validation was not run, say exactly what was run and what was not.
 - If a change depends on external prerequisites such as `libdap`, hyrax-dependencies, AWS-fetched artifacts, or system packages, note that clearly.
+- Do not flatter me
```
hannahilea, line 106: lol

configure.ac
```diff
@@ -957,6 +981,17 @@ AC_CHECK_STARE(
   $ac_bes_dependencies_prefix
 )
 
+dnl should the NCML handler be built? 
+AC_ARG_ENABLE([ncml], AS_HELP_STRING([--enable-ncml],[Enable NCML (default is YES)]),
```
hannahilea, line 985: hrm. above, we pass through `--disable-ncml` instead of this positive version. Perhaps there's some AC_ARG_ENABLE macro magic that means that writing it this way also supports the alternate case? 

- [x] At any rate, before merging, confirm that we are checking for the same flag values that we set in .travis.yml and .build-osx.yml!

modules/gdal_module/Makefile.am
hannahilea: ah, in configure.ac! Disregard. :)

hannahilea (changes requested):
There's another reference to GDAL_OPTION that isn't replaced in these changes---see https://github.com/OPENDAP/bes/blob/c20afd885af7dd92097690ae5ccf2a2cddda30f0/Makefile.am#L22

Also confirm that the `--enable-ncml` / `--disable-ncml` flag setting is as desired (see other comment).

Otherwise, lgtm!

hannahilea (approved):

configure.ac
hannahilea: I see from AC_ARG_ENABLE docs that this enable/disable magic does in fact happen. consider concern addressed!

jgallagher59701: The default sets a ncml_enabled=yes, so when the option is not used we (try to) build the NCML code.

For now, we force our builds to use --disable-ncml, making that action explicit.

PR #1310, opened 2026-04-15
Hyrax 2090

kyang2014:
## Description

**Reference ticket:** HYRAX-2090

The current code is using an HDF5 API to retrieve variable's dimension names. However, This HDF5 general API to retrieve an object's name given an object ID is very expensive when there are many objects in a file. In this ticket, we try to find the dimension names in a fast way if possible.

## Tasks

- [x] Tests added/updated
- [x] Dead code removed
- [x] No TODOs added


jgallagher59701 (approved):
This should be merged. It's a huge improvement and makes the difference between supporting these new data and not being able to support them. 

Please make a ticket to look into the complexity issues that sonarscan is flagging. I don't think it's a high priority, but it's the kind of thing that will make the software more valuable for the long term.

Also, for that header, can you explain why it cannot be at the top? Does that header require a struct and/or class to be defined before the header is included?

kyang2014:
> This should be merged. It's a huge improvement and makes the difference between supporting these new data and not being able to support them.
> 
> Please make a ticket to look into the complexity issues that sonarscan is flagging. I don't think it's a high priority, but it's the kind of thing that will make the software more valuable for the long term.
> 
Done. 
`> Also, for that header, can you explain why it cannot be at the top? Does that header require a struct and/or class to be defined before the header is included?`

Yes

PR #1325, opened 2026-04-21
HYRAX-2118, makes the HDF5 handler only support the HDF5 library version in the hyrax dependencies

kyang2014:
## Description

**Reference ticket:** HYRAX-2118

The HDF5 handler contains code that uses the HDF5 APIs that are only supported in the older HDF5 library versions. However, BES only depends on the HDF5 library distributed by Hyrax dependencies. So older HDF5 library versions will never be used. This ticket will remove the older version support. It will also make the handler use the newer APIs without considering the backward compatibilities. 

## Tasks

- [x] Dead code removed
- [x] No TODOs added


jgallagher59701 (approved):
Thanks

PR #1326, opened 2026-04-21
Build on OSX w/o GDAL

jgallagher59701:
## Description

The macos-15-intel CICD step is linking with GDAL but missing a needed library.

Patch so that it does not link with GDAL (configure using --without-gdal) to unblock the pipeline. Then,...
Check that the hyrax-dependencies are built with GDAL for both macos-15 intel and arm64.
Debug both builds once that is the case.

I can see that the current arm64 hyrax-deps binary is missing GDAL while the intel version has GDAL but not the missing libdefalte that causes the build to fail.

Work debug potion of this issue on a second branch so that we can merge the patch ASAP.

## Tasks

- [ ] ~Tests added/updated~
- [ ] ~Dead code removed~
- [x] No TODOs added

### DIFF
--- a/.github/scripts/upload-test-results.sh
+++ b/.github/scripts/upload-test-results.sh
@@ -1,0 +1,33 @@
+#!/bin/sh
+#
+# Upload the results of failed GitHub Actions tests to S3.
+
+set -e
+
+RUN_ID=${GITHUB_RUN_ID:-unknown-run}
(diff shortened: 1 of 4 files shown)


hannahilea (approved):

jgallagher59701:
This went past the place where macos-15-intel failed, but it still takes nearly an hour. Trying some simple speed ups - turned on parallel autotest runs.

PR #1330, opened 2026-04-22
HYRAX-2119: Make the csv handler fail to test log uploading for GHA CICD builds.

jgallagher59701:
## Description

Make the csv handler fail to test log uploading for GHA CICD builds. Those uploads were not working because of a policy error in AWSland

## Tasks

- [x] Tests added/updated
- [x] Dead code removed
- [x] No TODOs added


jgallagher59701:
The log upload worked. We don't need anything from this branch since the only change was the forced error.

PR #1331, opened 2026-04-22
Hyrax 2120

kyang2014:
## Description

**Reference ticket:** HYRAX-2120

Further improve the performance to generate DMR for netCDF-4-like HDF5 files. Now we can support almost all the cases. Tested with NASA testsuites. 

## Tasks

- [x] Dead code removed
- [x] No TODOs added


jgallagher59701 (approved):
Have you looked at the suggestions from sonarscan? We can merge this as is, but it might be a good idea to see if some of the 'complexity' suggestions can be implemented.

kyang2014:
> Have you looked at the suggestions from sonarscan? We can merge this as is, but it might be a good idea to see if some of the 'complexity' suggestions can be implemented.

I removed some. We may leave the left as is since the code is easy to logically follow with the current form.

PR #1334, opened 2026-04-23
Jhrg/macos cicd build test fix

jgallagher59701:
## Description

This is an offshoot of another ticket - Our CICD for macos is broken and this is a fix

## Tasks

- [x] Tests added/updated
- [x] Dead code removed
- [x] No TODOs added


jgallagher59701:
This is simple and it's blocking CICD, so I'm going to merge it. It's worth noting that the fix is to use `grep -E -f` in place of `diff`.

Issue #1332, opened 2026-04-23
BES Forever and Ever bug


ndp-opendap:
```
ssh ... test.opendap.org
cd hyrax
sudo -s
source ./hyraxctl
dshell # logs into Hyrax container

# Locate the PID of the bad actor (ps, top, etc)
ps -el | grep ’bes\|UID’ # Look for the long running beslistener(s) consuming max CPU

exit # (leave the continer, now you're on the host)

grep "|$PID|" log/bes/bes.log | python3 log/beslog2json.py -a | jq '.' > beslistener.json

cat beslistener.json | jq -r '."request-id"'

# take the last request_id
export req_id="http-nio-8080-exec-2_21-37d4c4f4-953c-4870-8118-0640a19a0818"

# find that request in the bes.log file
grep "$req_id" log/bes/bes.log | python3 log/beslog2json.py -a | jq

# inspect the result:
{
  "elapsed-us": 2,
  "instance-id": "-",
  "pid": 1054,
  "request-id": "http-nio-8080-exec-2_21-37d4c4f4-953c-4870-8118-0640a19a0818",
  "start-us": 1776956431193713,
  "stop-us": 1776956431193715,
  "time": "2026-04-23T15:00:31 GMT",
  "timer-name": "Command timing: BESXMLInterface::transmit_data() - set container in catalog values catalogContainer,/GESDISC/3B-HHR.MS.MRG.3IMERG.19980101-S000000-E002959.0000.V07B.HDF5.dmrpp;",
  "type": "timing"
}
{
  "elapsed-us": 1,
  "instance-id": "-",
  "pid": 1054,
  "request-id": "http-nio-8080-exec-2_21-37d4c4f4-953c-4870-8118-0640a19a0818",
  "start-us": 1776956431193774,
  "stop-us": 1776956431193775,
  "time": "2026-04-23T15:00:31 GMT",
  "timer-name": "Command timing: BESXMLInterface::transmit_data() - define d1 in default as catalogContainer;",
  "type": "timing"
}
{
  "bes-action": "get.dap",
  "ce": "-",
  "client-ip": "186.77.204.95",
  "http-verb": "HTTP-GET",
  "instance-id": "-",
  "local-path": "/usr/share/hyrax/GESDISC/3B-HHR.MS.MRG.3IMERG.19980101-S000000-E002959.0000.V07B.HDF5.dmrpp",
  "olfs-start-time": "-",
  "pid": 1054,
  "query-string": "-",
[...]

PR #1336, opened 2026-04-25
Fix the Max compute threads config option.

jgallagher59701:
## Description

**Reference ticket:** HYRAX-2130

Also relates to (solves?) #1332 

In DmrppRequestHandler.cc, read_bool_key() is used to read the number of threads. When the feature is turned on, this leads to a race condition.

## Tasks

- [x] Tests added/updated
- [x] Dead code removed
- [x] No TODOs added


kneumiller (approved):
If it passes tests, looks good to me

PR #1337, opened 2026-04-28
HYRAX-2143, remove more code that doesn't use the HDF5 latest APIs.

kyang2014:
## Description

**Reference ticket:** HYRAX-2143

Make sure the HDF5 handler is updated to use only the HDF5 latest APIs to prepare for the upgrades of HDF5 2.x in the hyrax dependencies.

## Tasks

- [x] Dead code removed
- [x] No TODOs added


jgallagher59701 (approved):
Thanks!

PR #1340, opened 2026-05-01
HYRAX-2144, use get_parent to traverse the ancestor group to avoid th…

kyang2014:
…e infinite loop and unexpected behavior.

## Description

**Reference ticket:** HYRAX-2144

Use libdap's get_ancestor() method to traverse DAP4 groups may give you unexpected output and it causes infinite loops for one NASA testing file.  We use the get_parents() to traverse with the expected results. 

## Tasks

- [x] Dead code removed


jgallagher59701 (approved):
Thanks for this! If get_parent () doesn't return a D4Group, we might want to put a note there that the cast will return nullptr.

PR #1341, opened 2026-05-04
HYRAX-2146, add tests for multi-unlimited dimensions that are defined…

kyang2014:
… under different groups.

## Description

**Reference ticket:** HYRAX-2146

Add tests(including NASA files) for multi-unlimited dimensions that are defined under different groups

## Tasks

- [x] Tests added/updated
- [x] No TODOs added


jgallagher59701 (approved):
Thanks

PR #1342, opened 2026-05-05
Hyrax 2147

kyang2014:
## Description

**Reference ticket:** HYRAX-2147

When the chunk size is greater than the array size, the only case for direct chunk IO to work is when a dimension is unlimited. However, even for this case when the unlimited dimension doesn't have a name,  we cannot apply direct chunk IO.  We didn't consider this case before so a failure is generated. This ticket  fixes this issue. 

## Tasks

- [x] Tests added/updated
- [x] Dead code removed
- [x] No TODOs added


jgallagher59701 (approved):
Thanks

PR #1344, opened 2026-05-05
Use the two arch test compare - 'minus zero' issue in gdal

jgallagher59701:
## Description

**Reference ticket:** None

Building hyrax-dependencies on the ARM64 (aka Apple silicon) fails three tests in the gdal handler because gdal returns minus zero. Fixed

## Tasks

- [ ] ~Ticket exists and is linked in title~
- [x] Tests added/updated
- [x] Dead code removed
- [x] No TODOs added


hannahilea (approved):
looks good, although I think you may also need to check in the updated baseline.m_proc files for these to pass locally?

jgallagher59701:
> looks good, although I think you may also need to check in the updated baseline.m_proc files for these to pass locally?

It's odd, but the *.m_proc baselines are/were already there - I think this is an issue that was fixed previously and then fix was 'undone' when we coalesced the autotest macros and the gdal handler was build was 'turned off' because of the (recurrent) gdal woes.

PR #1347, opened 2026-05-06
Jhrg/warnings removal 2

jgallagher59701:
## Description

**Reference ticket:** HYRAX-2108

Remove more warnings - this time in the HDF4/5 code

I have also addressed issue on OSX that caused binary build mismatches, but that requires a new hyrax-dependencies build as well. There will be a PR for that, too, soon.

And I have committed the text files that list the warnings that were removed over successive passes. I think this is possibly and interesting stat, and those also show some remaining warnings we might want to think about. For now, I claim this is good enough.

## Tasks

- [x] Tests added/updated
- [x] Dead code removed
- [x] No TODOs added


modules/hdf4_handler/hdfclass/sds.cc
```diff
@@ -477,7 +477,7 @@ hdfistream_sds & hdfistream_sds::operator>>(hdf_sds & hs)
 
 // Functor to help look for a particular map's ce in the vector of array_ce
 // objects.
-class ce_name_match:public std::unary_function < array_ce, bool > {
```
kyang2014, line 480: This is a very old code. I don't understand why this std::unary_function < array_ce, bool >  is there in the first place. Could you explain why you remove it?

modules/functions/tests/codex-bzip2-issue.md
kyang2014, line 1: Not sure why you add this to the repo? Perhaps it is fine, but when you have a new release, I don't think including this file is a good idea. Same thing for nlohmann and all the warnings txt below.

hannahilea: I agree---I think perhaps these prompts would be good to attach/copy into the ticket in which they were addressed, but I don't think we need a record of them directly in the repository themselves.

http/unit-tests/Makefile.am
```diff
@@ -2,11 +2,6 @@
 
 AUTOMAKE_OPTIONS = foreign
 
-# FIXME Remove this hack. Set these with configure. jhrg 11/25/19
```
hannahilea, line 5: ...I assume these are successfully set via configure now?

modules/hdf4_handler/hdfclass/sds.cc
jgallagher59701: The unary_function<> template is from pre-c++11 days and is now deprecated. I believe the C++17 still has it, but beyond that it is gone. I removed it because it's easy to do - I took output from our builds (make and make check) and then worked on removing as many of the warnings as I could. This was part of making it easier to spot the build system warnings that I think play into the issues with the NcML code.

modules/functions/tests/codex-bzip2-issue.md
jgallagher59701: Oh. Hah, yes, I do not mean to include those in the repo, but I've been switching from work in my office and home and I have used github to keep the context. 

So yes, this should not be in the code.

http/unit-tests/Makefile.am
jgallagher59701: IIRC this is about openssl and yes, it is handled correctly by configure now.

modules/hdf4_handler/hdfclass/sds.cc
jgallagher59701: I believe with this and a previous PR, I have removed all uses of unary_function<> in the BES. Not sure about libdap.

modules/functions/tests/codex-bzip2-issue.md
jgallagher59701: Removed. There are some other files called warnings.*.txt that will also not be merged to master (FYI). There are issues here that require tests against changes in hyrax-deps.

kyang2014 (commented):
Not sure what's purpose to have this?

nlohmann/json.hpp
```diff
@@ -1,9 +1,9 @@
 //     __ _____ _____ _____
 //  __|  |   __|     |   | |  JSON for Modern C++
-// |  |  |__   |  |  | | | |  version 3.11.2
+// |  |  |__   |  |  | | | |  version 3.12.0
```
hannahilea: I feel like all the changes in this file should maybe be a separate PR, since (unless I'm misunderstanding) they involve bumping a version of JSON, rather than the more-superficial fixes in the rest of this PR?

3819 lines of changes is really hefty to review, so I think it might be worth motivating that... I've looked over the other changes, but not the changes in this file.

hannahilea (changes requested):
The code changes except in nlohmannjson.hpp look good to me.

I'm not sure why the warnings-on-osx are checked in here; might those be better attached to a ticket for future fixes? If not, consider adding an explanation README.md in that warnings directory

kyang2014 (commented):
Also these warning files, should they be removed?

nlohmann/json.hpp
jgallagher59701: Hmm. OK. I added these here because the code is an 'include library' and it resulted in a large number of warnings wherever it was used. I can move this to a separate PR.

I can move the warnings files to a ticket, but I'd prefer not to in this case. I'll add a README explaining why.

kyang2014:
> Not sure what's purpose to have this?

You haven't answered this question yet.

nlohmann/json-3.11.2.hpp
kyang2014, line 1: What is the purpose of this directory nlohmann?

jgallagher59701: That holds the json.hpp header library we use to parse json.

jgallagher59701: That's not new - it's been there for some time. I updated the library since it was old and the source of a number of warnings.

ndp-opendap: This is a header only c++ json library that we use in a number of places.
We used to use "rapidjson" (I think it's still around and we should refactor/replace its usages with the nlohmann api)

kyang2014: Thanks for all the replies. I may miss something here. Why do we need to parse json in BES?

ndp-opendap: Many of the services that we use like CMR, AWS, and others respond with JSON documents.

ndp-opendap: And we want to get away from using `rapidjson` because the API is very very difficult to use because of the way that the API has changed the semantics of "`=`" from copy to move.

i.e. normally:
```
x = y
```
Means that the value of `y` is copied to `x`.

In `rapidjson` that expressions means that thing `y` is moved to thing `x` and thing `y` is set to `null`/`nullptr`

ndp-opendap: https://github.com/OPENDAP/bes/issues/1351

hannahilea (approved):
Might be worth a comment about where the nlohmann/json-3.xx.x.hpp files came from / what SHAs they are copied from elsewhere, as tracking; since they are essentially an external dependency, I haven't read/reviewed them, but it would be good to have a record (in code) of where they came from.

nlohmann/json-3.11.2.hpp
kyang2014: Thanks for all the explanation. That makes sense.

PR #1355, opened 2026-05-12
Hyrax 2050

kyang2014:
## Description

**Reference ticket:** HYRAX-2050

Provide an option to store the variable-length data in the side car file.

## Tasks

- [x] Tests added/updated
- [x] Dead code removed
- [x] No TODOs added


jgallagher59701 (approved):
Approved. 

This would be for another ticket: We might want to look at some of this code and see if the larger methods/functions can be broken down into shorter pieces. Sonarscan has a complexity threshold of 25 (and I'm not exactly sure how that arrive at that score, but I'm inferring that it's something like a McCabe score). Twenty five is pretty small, and we might not be able to hit that mark, but some of these functions are now very complex.

PR #1356, opened 2026-05-12
These files include pugixml.hpp without need.

jgallagher59701:
## Description

**Reference ticket:** HYRAX-2069

I found these while crosschecking the XML schema and the code.

## Tasks

- [x] Tests added/updated
- [x] Dead code removed
- [x] No TODOs added

### DIFF
--- a/modules/dmrpp_module/DmrppCommon.cc
+++ b/modules/dmrpp_module/DmrppCommon.cc
@@ -42,12 +42,6 @@
 #include <libdap/XMLWriter.h>
 #include <libdap/util.h>
 
-#if 0
-#define PUGIXML_NO_XPATH
-#define PUGIXML_HEADER_ONLY
-#include <pugixml.hpp>
-#endif
-
 #include "url_impl.h"
 #include "BESIndent.h"
 #include "BESDebug.h"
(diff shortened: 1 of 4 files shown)


modules/dmrpp_module/DmrppCommon.cc
```diff
@@ -42,10 +42,11 @@
 #include <libdap/XMLWriter.h>
 #include <libdap/util.h>
 
-
+#if 0
```
hannahilea: ...instead of switching this off, should be be deleting the lines?

modules/dmrpp_module/DmrppRequestHandler.cc
```diff
@@ -65,10 +65,12 @@
 
 #include "DapUtils.h"
 
+#if 0
```
hannahilea: ibid - consider leaving a note for when we'd WANT to turn them on, if we're going to keep them in but turned off.

hannahilea (changes requested):

modules/dmrpp_module/DmrppCommon.cc
jgallagher59701: Yes

modules/dmrpp_module/DmrppRequestHandler.cc
jgallagher59701: Yes. I was being sloppy... I'll delete them. But, I wanted to make sure that the code will build on Linux. Sometimes OSX is more lenient than linux (clang vs gcc)

hannahilea (approved):

PR #1358, opened 2026-05-13
Hyrax 2121

jgallagher59701:
## Description

**Reference ticket:** HYRAX-2121

Modifications to the BES build so that the new HDF5 2.1 library will link correctly.

When you are reviewing this, there are really only two changes:
1. AM_LDFLAGS was badly mangled in many of the Makefile.am files. FIXED
2. @hannahilea: Many baselines in generated tests were updated by Kent because they used generated HDF5 files and the new library generates slightly different files for the same data values.
3. There are a few misc changes for GDAL that are also linker flags.

## Tasks

- [x] Tests added/updated
- [x] Dead code removed
- [x] No TODOs added


kyang2014 (approved):
Thanks for the hard work. Still one question in my mind: looks like most of these change should not be related to the upgrade of the HDF5 library. Is it just by coincidence that the issues were revealed while we are upgrading the HDF5 library.

modules/dmrpp_module/get_dmrpp/unit-tests/grid_1_2d.h5.dmrpp.baseline
```diff
@@ -82,7 +82,7 @@
             <Value>libdap-3.21.1</Value>
         </Attribute>
         <Attribute name="invocation" type="String">
-            <Value>build_dmrpp -f /Users/myang6/work/opendap/hyrax-1913-test/bes/modules/dmrpp_module/get_dmrpp/unit-tests/grid_1_2d.h5 -r grid_1_2d.h5.dmr -u OPeNDAP_DMRpp_DATA_ACCESS_URL -M</Value>
+            <Value>build_dmrpp -f /home/rocky/kent/hyrax-2121-test/bes/modules/dmrpp_module/get_dmrpp/unit-tests/grid_1_2d.h5 -r grid_1_2d.h5.dmr -u OPeNDAP_DMRpp_DATA_ACCESS_URL -M</Value>
```
hannahilea, line 85: i hadn't noticed before that personal paths were saved into these baselines; I assume this if fine, but it is interesting.

modules/dmrpp_module/get_dmrpp/unit-tests/grid_2_2d_sin.h5.dmrpp.baseline
```diff
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="ISO-8859-1"?>
-<Dataset xmlns="http://xml.opendap.org/ns/DAP/4.0#" xmlns:dmrpp="http://xml.opendap.org/dap/dmrpp/1.0.0#" dapVersion="4.0" dmrVersion="removed" name="grid_2_2d_sin.h5" dmrpp:href="OPeNDAP_DMRpp_DATA_ACCESS_URL" dmrpp:version="3.21.1">
+<Dataset xmlns="http://xml.opendap.org/ns/DAP/4.0#" xmlns:dmrpp="http://xml.opendap.org/dap/dmrpp/1.0.0#" dapVersion="4.0" dmrVersion="1.0" name="grid_2_2d_sin.h5" dmrpp:href="OPeNDAP_DMRpp_DATA_ACCESS_URL" dmrpp:version="3.21.1">
```
hannahilea, line 2: ...do we want to switch this version back to 1.0 instead of removed? It seems like we shoudl be running the baselines through the version stripping paths (or removing them manually). 

While this won't cause failures now, it does undo some of the future version flexibility we spent effort on recently; consider switching this (and other) new `dmrVersion="1.0"` to `dmrVersion="removed"`

hannahilea (approved):

modules/dmrpp_module/get_dmrpp/unit-tests/grid_1_2d.h5.dmrpp.baseline
jgallagher59701: Yeah. There's a few things about the 'metadata' saved in the DMR++ I don't like.

modules/dmrpp_module/get_dmrpp/unit-tests/grid_2_2d_sin.h5.dmrpp.baseline
jgallagher59701: We should be. I had not caught that. However, lets make that a separate ticket.

jgallagher59701:
> Thanks for the hard work. Still one question in my mind: looks like most of these change should not be related to the upgrade of the HDF5 library. Is it just by coincidence that the issues were revealed while we are upgrading the HDF5 library.

The AM_LDFLAGS changes are general fixes, but we didn't need them until this moment. I decided to fix them all because once I fixed the behavior in the configure.ac script for this new library, all of the problems with AM_LDFLAGS surfaced all over the place (AWS SDK, GDAL, ...).

The tests 1.0 <--> 'removed' is I think something that might be a problem on Kent's machine. I believe all of those baselines have small changes like an 'offset' going from 1155 to 1156 and that's because those baselines are for _generated_ hdf5 files and the new HDF5 library makes slightly different HDF5 files. 

So, these changes are all about the new HDF5 library, but we had (have, to be candid) so many issues with our build process that the change took on global scope.

PR #1361, opened 2026-05-18
Hyrax 2116

kyang2014:
## Description

**Reference ticket:** HYRAX-2116

1. Reduce code complexity by dividing a complex function into several simple functions in HDF5 to DAP4 mapping code.
2. Use the more efficient parameter to retrieve HDF5 object information.
3. For fileout netCDF, separate the tests that consume the huge memory to a separate testsuite.
4. Also make BES compile with the older version of the openssl library.

## Tasks

- [x] Tests added/updated
- [x] Dead code removed
- [x] No TODOs added


modules/hdf5_handler/h5gmcfdap.cc
```diff
@@ -729,7 +725,16 @@ void gen_gmh5_cfdmr(D4Group* d4_root,const HDF5CF::GMFile *f) {
                 if(d4_root->attributes() != nullptr) {
 
                 // TODO: The following lines cause seg. fault in libdap4, needs to investigate
+                // I still find segmentation fault even with the commented code. Something is not right.
```
jgallagher59701, line 728: Is there a ticket for this?

jgallagher59701 (approved):
Approved, but see my comment. Thanks.

modules/hdf5_handler/h5gmcfdap.cc
kyang2014: No, thanks for the reminder.  I want to check if this is the case by using the libdap4 only. Then I will create a ticket for it.

PR #1363, opened 2026-05-20
Hyrax 1577

kyang2014:
## Description

**Reference ticket:** HYRAX-1577

For the dmrpp file generated from the HDF4, add an attribute configuration in the dmrpp file to indicate which option (EnableCF=true or EnableCF=false) is used to generate the dmr file.
Also issue an error message in gen_dmrpp_side_car if people try to generate the dmrpp with the HDF4 EnableCF=true. This is currently not supported.
Updated the gen_dmrpp_side_car test baseline files and the script.

## Tasks

- [x] Tests added/updated
- [x] Dead code removed
- [x] No TODOs added


jgallagher59701 (approved):
Thanks

PR #1364, opened 2026-05-20
Hyrax 2156

kyang2014:
## Description

**Reference ticket:** HYRAX-2156

Reduce code complexity in the build_dmrpp_util.cc. Also use the unique pointer to replace new for the dmrpp metadata generation.

## Tasks

- [x] Tests added/updated
- [x] Dead code removed
- [x] No TODOs added


jgallagher59701 (approved):
Thanks!

PR #1366, opened 2026-05-26
HYRAX-2169: Improve error for failed STS credentials parse

hannahilea:
## Description

**Reference ticket:** HYRAX-2169

## Tasks

- [x] Tests added/updated
- [x] Dead code removed
- [x] No TODOs added


ndp-opendap:
@hannahilea - Did you intend to assign this to me? Or maybe ask for my review?

hannahilea:
~~monday~~ tuesday-post-holiday brain! def meant to ask for a review. 🙃

aws/SignedUrlCache.cc
```diff
@@ -46,6 +46,7 @@
 #include <aws/s3/S3Client.h>
 
 #include "rapidjson/document.h"
+#include <rapidjson/error/en.h>
```
ndp-opendap: Maybe @jgallagher59701 will have something to say about:
```
#include <rapidjson/error/en.h>
```
vs
```
#include "rapidjson/error/en.h"
```

It's a c++ thing/convention that is eluding me...

ndp-opendap (approved):
Looks good but I think there's a small point to address if you're willing. 

See my comment which fails to actually say what to do but only alludes to what would @jgallagher59701 do (wwjd - snerk)

aws/SignedUrlCache.cc
hannahilea:
```suggestion
#include "rapidjson/error/en.h"
```

jgallagher59701: If the header is part of the repo, it should use double quotes. If it's a 'system' header (which I claim includes the stuff in libdap and the hyrax-dependencies) it should use angle brackets. Really, I don't think it matters that much anymore. But it's a good convention to follow - it can remind you that the stuff a given header includes was installed as part of some other build process.

PR #1367, opened 2026-05-26
Hyrax 1956

kyang2014:
## Description

**Reference ticket:** HYRAX-1956

Make fileout netCDF not to compress large size floating-point array,

## Tasks

- [x] Tests added/updated
- [x] Dead code removed
- [x] No TODOs added


modules/dmrpp_module/DmrppNames.h
```diff
@@ -48,6 +48,12 @@
 #define DMRPP_MAX_COMPUTE_THREADS_KEY "DMRPP.MaxComputeThreads"
 
 #define DMRPP_USE_CLASSIC_IN_FILEOUT_NETCDF "FONc.ClassicModel"
+#if 0
```
jgallagher59701, line 51: We should remove these #if 0 ... #endif blocks unless they are going to be used later at some point. If you were planning to do that as a next step, that's ok with me.

modules/fileout_netcdf/FONcArray.cc
```diff
@@ -528,6 +528,22 @@ bool FONcArray::is_unlimited_dim(const string &dim_name) const {
     return ret_value;
 }
 
+bool FONcArray:: check_float_write_opt() const {
+
+    if (FONcRequestHandler::float_write_opt == false || d_a->get_dio_flag() || (d_array_type != NC_FLOAT && d_array_type != NC_DOUBLE) )
+        return false;
+    unsigned array_size_in_bytes = d_nelements*d_a->var()->width_ll();
```
jgallagher59701: Sonar scan has a warning here that "implicit conversion loses integer precision: 'size_t' (aka 'unsigned long') to 'unsigned int'" which is kind of confusing to me because I thought BawseType::width_ll() returned 'long long'. Lets try to address this warning, maybe by making array_size_in_bytes 'auto'?

jgallagher59701 (approved):
I have two comments, let's address those and the merge this. Thanks.

PR #1373, opened 2026-06-02
HYRAX-2166

dh-opendap:


modules/asciival/get_ascii.cc
```diff
@@ -156,7 +156,13 @@ basetype_to_asciitype( BaseType *bt )
 	    return new AsciiGrid( dynamic_cast<Grid *>(bt) ) ;
 
     default:
-        throw InternalErr(__FILE__, __LINE__, "Unknown type");
+        string errMsg;
+        if ( bt->is_dap4() )
+            errMsg = "THIS IS A DAP4 DATASET: Use it with the DAP4 API, ASCII output not supported for the type '" + bt->type_name() + "' which this dataset contains.";
```
ndp-opendap, line 163: What do you think about including a link to the dap4 specification section for constraint expressions? Or a ce tutorial? (@Mikejmnez Is that a thing we even have??) 
Something to help the user take the next step...

Constraints In dap4:
```
https://opendap.github.io/dap4-specification/DAP4.html#_constraints
```

ndp-opendap: After looking at the code for asciivall and checking how this happens in other places I find that BESDapResponder methods for dap2:

- `BESDapResponseBuilder::send_das()`
- `BESDapResponseBuilder::send_dds()`
- `BESDapResponseBuilder::send_dap2_data()`

All call the utility function:
```
        dap_utils::throw_for_dap4_typed_vars_or_attrs(*dds, __FILE__, __LINE__);
```
Which I believe addresses the dap4 things in a dap2 response issue.
It's notable that the asciival pathway calls `BESDapResponseBuilder::intern_dap2_data()` which does NOT call our friend `dap_utils::throw_for_dap4_typed_vars_or_attrs()` Which I think is an old mistake and I think would address the core desire behind this ticket.

So think we should skip this change for now  and try adding a call to our friend `dap_utils::throw_for_dap4_typed_vars_or_attrs()` to the `BESDapResponseBuilder::intern_dap2_data()` method on line 894.

ndp-opendap (changes requested):

dh-opendap:
1. Modified BESDapResponseBuilder::intern_dap2_data() to call throw__for_dap4_typed_vars_and_attrs(); this returns a more verbose and correct error message, but only covers requests with no CEs, or with CEs containing DAP4 variables.
2. Modified modules/asciival/get_ascii.cc - datadds_to_ascii_datadds(); only convert projected variables.

modules/asciival/get_ascii.cc
dh-opendap: Changes made to address your concern.  Tested and resolves the problem better than the initial approach.

ndp-opendap: woot!!

ndp-opendap (approved):
Nice! Thanks Dan!

PR #1375, opened 2026-06-03
HYRAX-2178: Add EDL user id to relevant cache keys in SignedURLCache

hannahilea:
## Description

**Reference ticket:** HYRAX-2178

## Tasks
- [x] Tests added/updated
- [x] Dead code removed
- [x] No TODOs added

### DIFF
--- a/aws/SignedUrlCache.cc
+++ b/aws/SignedUrlCache.cc
@@ -68,7 +68,7 @@ namespace bes {
 void SignedUrlCache::cache_presigned_s3_url(string const &unsigned_url,
                                             shared_ptr<http::EffectiveUrl> const signed_url) {
     std::string key = append_edl_username_to_key(unsigned_url);
-    INFO_LOG(prolog + "Caching presigned url with key - " + key);
+    BESDEBUG(MODULE, prolog + "Caching presigned url with key " + key);
     d_presigned_s3_urls_cache[key] = signed_url;
 }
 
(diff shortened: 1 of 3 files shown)


aws/SignedUrlCache.cc
```diff
@@ -102,20 +117,50 @@ bool SignedUrlCache::is_timestamp_after_now(std::string const &timestamp_str) {
...
+ * @note This method is not, itself, thread safe.
+ */
+std::string SignedUrlCache::append_edl_username_to_key(string const &key) {
+    bool found = false;
+    string uid = BESContextManager::TheManager()->get_context(EDL_UID_KEY, found);
+    BESDEBUG(MODULE, prolog << "EDL_UID_KEY(" << EDL_UID_KEY << "): " << uid << endl);
+    if (found && !uid.empty()) {
+        return key + ":" + uid;
+    }
+    INFO_LOG(prolog + string("SERVICE CHAIN WARNING - EDL UID missing; using raw key " + key));
```
ndp-opendap, line 132: In the event that the BESContextManager doesn't have the `uid` value would there be a benefit from assigning the `uid` to some "signal" value? Like "`missing-uid`" or "`unknown-user`"??  Either within Hyrax or possibly in the various logs??

ndp-opendap (approved):
Looks great - one question but it's good to go!

aws/SignedUrlCache.cc
hannahilea: Welp, was overeager and merged before seeing this message. 

That's a good question. For the purposes of this specific feature, I don't think it would make much difference---I think it would be extremely unlikely (so many different checkpoints to prove EDL identity before now), so having the logged warning seems sufficient to me. BUT it isn't a strong opinion, I'm happy to update it if you feel like it would provide a benefit.

PR #1376, opened 2026-06-03
HYRAX-2177: Add EDL user id to cache key in EffectiveURLCache

hannahilea:
## Description

**Reference ticket:** HYRAX-2177

## Tasks
- [x] Tests added/updated
- [x] Dead code removed
- [x] No TODOs added


http/EffectiveUrlCache.cc
```diff
@@ -56,8 +57,9 @@ namespace http {
  * @param url_key Key to a cached effective URL.
  * @note This method is not, itself, thread safe.
  */
-shared_ptr <EffectiveUrl> EffectiveUrlCache::get_cached_eurl(string const &url_key) {
+shared_ptr <EffectiveUrl> EffectiveUrlCache::get_cached_eurl(string const &key_prefix) {
     shared_ptr<EffectiveUrl> effective_url(nullptr);
+    auto url_key = append_edl_username_to_key(key_prefix);
```
ndp-opendap: Here's my pedantic comment, which you should feel free to ignore...

Renaming `url_key` to `key_prefix` in the method signature and then appending the edl user name to the `key_prefix` it to get `url_key` is confusing.

Maybe use `url_prefix` instead?
```
shared_ptr <EffectiveUrl> EffectiveUrlCache::get_cached_eurl(string const &url_prefix) {
    auto url_key = append_edl_username_to_key(url_prefix);
```

Also, though it may not matter here at all, when we did this in for the disk cache files (cmr etc) we prepended the edl username to the cache file name so that it was super easy to see who owned what things.

ndp-opendap (approved):
Not a huge side quest I hope!

http/EffectiveUrlCache.cc
hannahilea: yeah, that's a good point. I don't love `url_prefix`, but I'll switch it to `url`---that doesn't imply that it *is* the key, at least. 

And nope, luckily not a huge side quest, especially since I did #1375 first! phew.

PR #1377, opened 2026-06-04
Hyrax 2176

kyang2014:
## Description

**Reference ticket:** HYRAX-2176

This is just to test if the fix of DAP4 attribute bug works as expected in BES. 

## Tasks

- [x] Dead code removed
- [x] No TODOs added


jgallagher59701 (approved):
Thanks

PR #1392, opened 2026-06-17
Control authentication scope for  the curl::http_get_and_write_resource() function

ndp-opendap:
## Description

This PR amends the function `curl::http_get_and_write_resource()` so that it can be asked to conditionally add authentication information to the outgoing request headers. Currently it always adds these headers, even when that's not useful.

 `curl::http_get_and_write_resource()` is used in exactly one place, `RemoteResource::get_url()`. Which, in turn, is only ever called by `RemoteResource::retrieve_resource`, which is widely used.

Here's is the how the change affects the dependent code:
- `CmrContainer.cc` - _Uses EDL to authenticate with CMR_
- `GatewayContainer.cc` - _No Auth_
- `HttpdCatalogContainer.cc` - _No Auth_
- `JsonUtils.cc` - _Exclusively used by the CmrApi to interrogate CMR. Uses EDL to authenticate with CMR_
- `S3Container.cc` - _Authenticates (I think with S3 signing headers)_

This PR addresses the authentication needs and stops authentication headers from being broadcast by Gateway and HttpdCatalog.

## Tasks

- [ ] Tests added/updated
- [ ] Dead code removed
- [ ] No TODOs added

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -237,7 +237,7 @@ jobs:
       - autoreconf --force --install --verbose
       - ./configure $CONFIGURE_OPTIONS --disable-dependency-tracking --prefix=$prefix --with-dependencies=$prefix/deps --enable-developer --enable-coverage
       - build-wrapper-linux-x86-64 --out-dir bw-output make -j16
-      - sonar-scanner -Dproject.settings=sonar-bes-framework.properties -Dsonar.login=$SONAR_LOGIN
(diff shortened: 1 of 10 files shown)


ndp-opendap:
also migrated the SONAR API to sonar.token from sonar.login and put the token value in TravisCI settings and not in the .travis.yml file. One token for all the targets, on token to rule them all,

PR #1393, opened 2026-06-18
HYRAX-2191, map the 1-character string array as the equvalient char a…

kyang2014:
…rray in netCDF.

## Description

**Reference ticket:** HYRAX-2191

Make the 1-character string array as the  char array with the same number of dimensions. Currently an additional dimension is added.

## Tasks

- [x] Tests added/updated
- [x] Dead code removed
- [x] No TODOs added


modules/fileout_netcdf/FONcArray.cc
```diff
@@ -388,48 +407,37 @@ void FONcArray::convert(vector<string> embed, bool _dap4, bool is_dap4_group) {
...
+            char_max_last_dim_length++;
+
+            vector<string> empty_embed;
+            string lendim_name;
+            if (is_dap4_group == true) {
+                // Here is a quick implementation. 
+                // We just append the DimNameNum(globally defined)
+                // and then increase the number by 1.
+                ostringstream dim_suffix_strm;
+                dim_suffix_strm << "_len" << FONcDim::DimNameNum + 1;
```
jgallagher59701, line 420: It's not necessary here, but now that we have C++ 14 this could be done without ostringstream using to_string(FONcDim::DimNameNum + 1). In a loop that would make the code faster.

jgallagher59701 (approved):
Thanks for fixing this! One comment, no change needed.

modules/fileout_netcdf/FONcArray.cc
kyang2014: I copy this to the comment in a ticket to clean up the fileout netCDF.

PR #1394, opened 2026-06-18
Refactor curl::http_get_and_write_resource() to use passed in request_headers

### LINKED ISSUE #1389: Stop indiscriminately broadcasting EDL stuff with every outgoing request.
## Overview
The BES uses a centralized cURL/HTTP library (_bes/http_) for doing almost all of its remote access requests. When making HTTP GET requests the code chacks to see if the BES has been handed EDL information (Bearer Token, UID, etc) and if present these are added to the outgoing requests as request headers. This fine if the target is CMR/EDL/TEA/etc, but for other remote services these information need to be elided.

## Approach
I think we need one or more new functions in `CurlUtils.cc`:

- `curl::edl_http_get_and_write_resource()`  - Which would need to be integrated into `CmrContainer.cc` and `JsonUtils::get_as_json()` (which is used by `CmrApi.cc`) - perhaps `JsonUtils::edl_get_as_json()` 

- `curl::edl_http_get()` - which would replace the use of `curl::http_get()` in `NgapApi.cc`, `NgapOwnedContainer.cc`, and `SignedUrlCach.cc`

- The code in `curl::get_redirect_url()` - which uses `curl::gru_mk_attempt()` and is used by `EffectiveUrlCache::get_effective_url()` which in turn is **ONLY** used by `Chunk.get_data_url()` already adds these headers pedantically.  This one is a bit thorny because, in general, chunks are held in S3 and we don't want to send EDL things to S3. So work to be done there for sure.

Basically the idea here is to refactor the code so that edl information is only ever sent to the small number of endpoints (CMR, TEA, etc) that require it.

ndp-opendap:
## Description
This PR changes the curl::http_get_and_write_resource() ecosystem to support passing in HTTP request headers so that top level code which knows that the request may need to be authenticated can handle creating the associated headers and not rely on having the request authentication semantics baked into the http/curl package.
 
This is one step in resolving [the linked issue regarding broadcasting EDL authentication information](https://github.com/OPENDAP/bes/issues/1389).

**Reference ticket:** There is no HYRAX ticket.

## Tasks

- [ ] Tests added/updated
- [ ] Dead code removed
- [ ] No TODOs added


kneumiller:
I believe it looks good. Do the other Tasks like tests and TODOs need to be added/checked before this can be closed out?

http/CurlUtils.cc
```diff
@@ -1482,6 +1499,44 @@ curl_slist *add_edl_auth_headers(curl_slist *request_headers) {
...
+    if (found && !s.empty()) {
+        request_headers = append_http_header(request_headers, UID_REQUEST_HEADER_KEY, s);
+    }
+
+    s = BESContextManager::TheManager()->get_context(EDL_AUTH_TOKEN_CONTEXT_KEY, found);
+    if (found && !s.empty()) {
+        request_headers = append_http_header(request_headers, AUTHORIZATION_REQUEST_HEADER_KEY, s);
+    }
+    else {
+#ifdef DEVELOPER
```
jgallagher59701: I think we should remove #ifdef DEVELOPER block and the else-clause. This was used by me to enable testing part of the "Get DMR++ documents from an S3 bucket we 'owned'". We have abandoned that idea and keeping this in the code just leaves us with the chance to release 'DEVELOPER' code to PROD and include some sort of security hole (I don't know what that would be exactly, but I'm uncomfortable leaving this in the code). 

My vote is that we remove this. What do you think?

jgallagher59701 (approved):
This can be merged as it is, but see my comment about removing the code that is conditional on DEVELOPER being defined. 

Thanks for cleaning up the sonar login stuff.

http/CurlUtils.cc
ndp-opendap: I agree it should go. But I fear it may be in use... I have had to rewind a bunch of changes and there's a test in bes/bescmdln/tests that fails:
```
./testsuite -v 61
```
And it fails because it succeeds on my machine and it's expected to fail, but in Travis it fails but it's expected to succeed.

And I thought it was orthogonal to my changes so it's a bit crazy making. It seems to unexpectedly find credentials on my system and it doesn't find them (when it should) on Travis.

jgallagher59701: That should not be in use. (should not i the sense that I don't think it is, but in the sense that it's part of work I did to push past an obstacle to test other functionality. Then we - ESDIS - decided not to have us do the work). You can try removing it by changing 'DEVELOPER' to '0'.

jgallagher59701: Or, we can make a new ticket for it's removal - that might be best all around.

ndp-opendap: My problems seem to be unrelated to our `#ifdef DEVELOPER` discussion... More soon

PR #1397, opened 2026-06-22
Refactor the curl::http_get() ecosystem to utilize passed request header lists

### LINKED ISSUE #1389: Stop indiscriminately broadcasting EDL stuff with every outgoing request.

ndp-opendap:
## Description

TODO

## Tasks

- [ ] Tests added/updated
- [ ] Dead code removed
- [ ] No TODOs added

### DIFF
--- a/http/CurlUtils.cc
+++ b/http/CurlUtils.cc
@@ -1746,16 +1746,41 @@ bool process_get_redirect_http_code(const long http_code,
 }
 
 /**
+ *
+ * @param source_list A pointer to a curl_slist to copy
+ * @return A copy of the source_list
+ */
+curl_slist *copy_curl_slist(const curl_slist *source_list) {
+    curl_slist *new_list = nullptr;
+    const curl_slist *current = source_list;
+
+    while (current != nullptr) {
+        curl_slist *temp = curl_slist_append(new_list, current->data);
+        if (temp == NULL) {
+            // Memory allocation failed; free the newly created list to avoid leaks
+            curl_slist_free_all(new_list);
+            return nullptr;
+        }
+        new_list = temp;
+        current = current->next;
+    }
+
+    return new_list;
+}
+
+/**
  * @brief  Make a single attempt to acquire a redirect response from origin_url
  * @param origin_url The URL to access
  * @param attempt The attempt number of this effort.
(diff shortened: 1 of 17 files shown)


http/CurlUtils.cc
jgallagher59701, line 629: We should remove this and add a line the git log that has happened.

jgallagher59701 (approved):
Lets remove the unused function and merge this. No need for another review.

ndp-opendap:
@jgallagher59701 & @dh-opendap - Before we merge this I think we should do a test deployment to SIT. This touches a bunch of the EDL authentication points and I want to see it work before we get it onto the main branch.

http/CurlUtils.cc
ndp-opendap: I will deal with this in as part of [the `awsv4.cc` issue](https://github.com/OPENDAP/bes/issues/1402)

PR #1398, opened 2026-06-23
Hyrax-2190: Upload test logs for Docker

kneumiller:
## Description

**Reference ticket:** HYRAX-2190

TODO

## Tasks

- [x] Tests added/updated
- [x] Dead code removed
- [x] No TODOs added


.travis.yml
```diff
@@ -185,6 +185,12 @@ jobs:
           export SNAPSHOT_IMAGE_TAG="opendap/${DOCKER_NAME}:snapshot-${DIST}"
           export BES_REPO_DIR=${TRAVIS_BUILD_DIR}
         - ./travis/build-rhel-docker.sh
+        # Note: Take the test log tarball that was created in the Dockerfile and copied to the final mount
```
jgallagher59701, line 191: Maybe: and copied to the --> and copy it to the

.travis.yml
```diff
@@ -185,6 +185,12 @@ jobs:
...
+        #   and then run this copy command to copy it into Travis.
+        #   Use "$1" because Single quotes prevent the host shell (Travis) from expanding ${TRAVIS_JOB_NUMBER} before invoking
```
jgallagher59701: I had to ask Claude about this pattern - the -c 'cp ...' sh "/scratch..." part.

It becomes: cp /bes-test-logs/bes-test-logs.tar.gz /scratch/bes-autotest-${TRAVIS_JOB_NUMBER}-logs.tar.gz.

The reason for the obscure $1 and `sh "/scratch/...${TRAVIS_JOB_NUMBER}..." is that the TRAVIS_JOB_NUMBER env var gets expanded in the container using the pattern. I'll forget that by next week ;-) We better put in a comment like: This pattern means the environment variable is passed into the container where it is then expanded.

But, now that I've thought about it, I don't think we need to do that. Since the TRAVIS_JOB_NUMBER is set in/by Travis, it'll have the same value in both places. I think you can use the plain:

-c "cp /bes-test-logs/bes-test-logs.tar.gz /scratch/bes-autotest-${TRAVIS_JOB_NUMBER}-logs.tar.gz"

and it'll work  (and we will know what's going on and won't have to write a huge comment. ;-)

jgallagher59701 (commented):
Look at these changes - see if the simpler **docker run .... -c "..."** works and if so lets use that. Else, put a comment in the Dockerfile explaining what's going on. Thanks.

ndp-opendap (commented):
This looks good.

Have you confirmed that the test log tar ball is **_not_** included in the end product  _bes_core_ image generated by Travis and used downstream?

ndp-opendap (approved):
I'm satisfied and we can move on.

PR #1400, opened 2026-06-24
Removed sed operations on unused PEM files in bes/Dockerfile

dh-opendap:
## Description

**Reference ticket:** HYRAX-2081

Updated Dockerfile; removing sed operations on unused PEM files.

## Tasks

- [ ] Tests added/updated
- [ ] Dead code removed
- [ ] No TODOs added


ndp-opendap (approved):
Looks great: Sorry about the `-test-deploy` confusion!

jgallagher59701 (commented):
This looks good.

jgallagher59701:
I was looking for all the places where .*\.pem files are mentioned and there are 32 other files - mostly 'bes.conf.in' (for example in bes/dispatch. The lines are commented out, but we could just get rid of this stuff - probably with an find/replace command in CLion, vscode or emacs.

I think we should remove those as part of this.

PR #1401, opened 2026-06-25
Hyrax 1225

kyang2014:
## Description

**Reference ticket:** HYRAX-1225
Draft, just for testing.

Re-engineer the fixed-size string array handling to avoid the memory leaks.

## Tasks

- [x] Tests added/updated
- [x] Dead code removed


modules/dmrpp_module/DmrppArray.cc
```diff
@@ -3614,4 +3620,417 @@ int DmrppArray::obtain_subset_dims(vector<unsigned long long>& var_start,vector<
...
+            // Compute where we are going to read it from
+            (*chunk_element_address)[dim] = chunk_start;
+
+            unsigned long long target_fstr_start_index =
+                get_index(*target_element_address, constrained_array_shape);
+            unsigned long long chunk_fstr_start_index = get_index(*chunk_element_address, chunk_shape);
+
+            for (unsigned long long temp_count= 0; temp_count <count; temp_count++) { 
+                unsigned long long source_char = (chunk_fstr_start_index+temp_count) * chars_per_string;
+                get_str()[target_fstr_start_index+temp_count] = ingest_fixed_length_string(source_buffer+source_char,chars_per_string,pad_type);
```
jgallagher59701, line 3956: ingest_fixed_length_string() has a comment about potential issues with buffer overflows. Here is source_char is larger than the size of source_buffer, that overflow would triggered - at least I think so.

modules/dmrpp_module/DmrppArray.cc
```diff
@@ -3614,4 +3620,417 @@ int DmrppArray::obtain_subset_dims(vector<unsigned long long>& var_start,vector<
...
+    auto last_dim = (unsigned int)(chunk_shape.size() - 1);
+    if (dim == last_dim) {
+        const char *source_buffer = chunk->get_rbuf();
+        auto pad_type = get_fixed_length_string_pad();
+
+        // This is the last dimension's offset.
+        array_offset += chunk_origin[dim];
+        for (unsigned long long temp_count= 0; temp_count <=chunk_end; temp_count++) { 
+            unsigned long long source_char = (chunk_offset+temp_count) * chars_per_string;
+            string temp_str = ingest_fixed_length_string(source_buffer+source_char,chars_per_string,pad_type);
```
jgallagher59701, line 4014: Same issue (essentially) as above.

modules/dmrpp_module/DmrppArray.cc
```diff
@@ -3614,4 +3620,417 @@ int DmrppArray::obtain_subset_dims(vector<unsigned long long>& var_start,vector<
...
+            unsigned long long target_fstr_start_index =
+                               get_index(*target_element_address, constrained_array_shape);
+            unsigned long long array_temp_count = 0;
+
+            for (unsigned long long chunk_index = chunk_start; chunk_index <= chunk_end; chunk_index += thisDim.stride) {
+                // Compute where we are going to read it from
+                (*chunk_element_address)[dim] = chunk_index;
+
+                unsigned long long chunk_fstr_start_index = get_index(*chunk_element_address, chunk_shape);
+                unsigned long long source_char = chunk_fstr_start_index * chars_per_string;
```
jgallagher59701, line 3972: source_char could exceed the source_buffer (in the call to ingest_fixed_length_string() just below here.

modules/dmrpp_module/DmrppArray.cc
jgallagher59701, line 2285: I'm not sure how we would fix this, but there is a potential buffer overflow here when fixed_str_len is bigger than the size of the buffer 'buf'. Maybe we could pass in the buffer size and check that here. I see there are 9 calls to this function in this file.

Maybe that is complicated enough so that we make a new ticket just for that and merge this as is with the caveat that there a bit more work to do to make the software less likely to have buffer overflow problems.

What do you think?

jgallagher59701 (approved):
There is an issue in ingest_fixes_length_string(). Maybe we can fix that now, or maybe it is a future ticket. What do you think? If you make ticket for it - see my comments - then this can be merged now. Or maybe the issue is simple to fix now. Your call.

modules/dmrpp_module/DmrppArray.cc
kyang2014: Agree. I will create another ticket to handle this case.

PR #1405, opened 2026-06-30
Hyrax 2193

kyang2014:
## Description

**Reference ticket:** HYRAX-2193

When calculating the request var size of a string array, the string size is always 24. This may cause false alarm and it happens to a real file in service. This pull request will relax the calculation by always assuming the string size is 2, which is the real file case. Since string data is easy to be compressed and decompressed, the service time for a bigger string is relatively small, so we think this should be OK even for bigger string size array.

## Tasks

- [x] Tests added/updated
- [x] Dead code removed
- [x] No TODOs added

### DIFF
--- a/dap/DapUtils.cc
+++ b/dap/DapUtils.cc
@@ -307,13 +307,15 @@ uint64_t crsaibv_process_variable(
         }
         else {
             // width_ll() returns the number of bytes needed to hold the data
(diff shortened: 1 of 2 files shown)


dh-opendap (approved):
Changes all seem reasonable.

PR #1406, opened 2026-07-06
Hyrax 2197

kyang2014:
## Description

**Reference ticket:** HYRAX-2197, HYRAX-2200

1. Support the subset of the fixed-size string array for the case when the chunk size is greater than the array size.
2. Add more fixed-size string tests to the dmrpp and the fileout netCDF modules. 

## Tasks

- [x] Tests added/updated
- [x] Dead code removed
- [x] No TODOs added


jgallagher59701 (approved):
Thanks!

PR #1408, opened 2026-07-08
HYRAX-2205, add variable-string array tests for different encoding me…

kyang2014:
…thods in the dmrpp module to the dmrpp and fileout netCDF modules.

## Description

**Reference ticket:** HYRAX-2205

Add variable-length string array tests to the dmrpp and fileout netCDF modules. It covers all the variable-length string encoding methods in the dmrpp module.

## Tasks

- [x] Tests added/updated
- [x] Dead code removed
- [x] No TODOs added


dh-opendap (approved):
These all seem correct, they are all test files or test file generators.

PR #1414, opened 2026-07-24
update README with tags, version, and DOI from zenodo

Mikejmnez:
## Description
Updated to display the updated DOI badge associated with new release


jgallagher59701 (commented):
I thought we were using the 'latest' DOI for these README files?

Mikejmnez:
This is outdated (opened during ESIP). But I see the readme file already points to the right zenodo doi

PR #1417, opened 2026-08-03
HYRAX-2005, remove unused FONcByte.h and FONcByte.cc and the correspo…

kyang2014:
…nding affected files.

## Description

**Reference ticket:** HYRAX-2005
Clean up the source code files(*.cc and *.h) with the following:
1. Remove unused  FONcByte.h and FONcByte.cc.
2. BES key update: 1) Initialize the default BES key values. 2) Remove an unused BES key.  3) Change a BES key to correct name, type. 4) Update the corresponding configuration files. 
3. Largely clean up the two most important source code files: FONcArray.cc and FONcTransform.cc. Reduce the code complexity by adding new methods.
4. Reduce memory leaks when an exception occurs.
5. Clean up comments.
6. Remove unused code blocks and methods. 
7. Update some baseline files.

## Tasks

- [x] Tests added/updated
- [x] Dead code removed
- [x] No TODOs added


modules/fileout_netcdf/FONcArray.cc
```diff
@@ -195,20 +213,91 @@ void FONcArray::convert(vector<string> embed, bool _dap4, bool is_dap4_group) {
...
+            }
+            else {
+                d_dont_use_it = true;
+            }
+        }
+    }
+
+    BESDEBUG("fonc", "FONcArray::convert() - done converting array " << d_varname << endl);
+}
+void FONcArray::convert_dimension_info(const std::vector<std::string> &embed) {
```
jgallagher59701, line 297: You might want to look at the sonar scan information for this function and see if some of the changes are warranted. There might be a libdap::Array::dimensions() method that would let you replace the dim_begin() and dim_end() methods with a range-based for loop. I cannot remember if that's the case, but it might make the code a bit easier to read.

modules/fileout_netcdf/FONcInt64.cc
```diff
@@ -38,21 +38,21 @@
...
+ * @param b A DAP BaseType that should be an Int64
  * @throws BESInternalError if the BaseType is not an Int64
  */
 FONcInt64::FONcInt64( BaseType *b )
-    : FONcBaseType(), _bt( b )
+    : FONcBaseType(), _int64(nullptr)
 {
-    Int64 *u64 = dynamic_cast<Int64 *>(b) ;
-    if( !u64 )
+    _int64 = dynamic_cast<Int64 *>(b) ;
```
jgallagher59701: Sonar says to assign this in an initializer, but I'm not sure if that's correct. I don't know if you can write a constructor like this:

   FONcInt64::FONcInt64( BaseType *b )
    : FONcBaseType(), _bt( b )
    : FONcBaseType(), _int64(dynamic_cast<Int64 *>(b))

Maybe give it a try? I don't know. Sonar might not be correct here.

jgallagher59701 (approved):
Two suggestions, but if those don't make sense, it can be merged as is. Thanks.

modules/fileout_netcdf/FONcArray.cc
kyang2014: Unfortunately this cannot be fone.  libdap::Array::dimensions()  returns the number of dimensions. Furthermore, I need to use the iterator to obtain the dimension size.

modules/fileout_netcdf/FONcInt64.cc
kyang2014: Give a try, work on the local system. Now see how travis behaves.

PR #1422, opened 2026-08-12
HYRAX-2222, for non-super chunk cases, release the buffer used by a c…

kyang2014:
…hunk after reading data of a dmrpp array.

## Description

**Reference ticket:** HYRAX-2020

Release memory after reading the array variable data when the data storage is contiguous or only contains one chunk.
Otherwise, the memory of these variables will only be released right before the dmr object is destroyed. 

## Tasks

- [x] No TODOs added

### DIFF
--- a/modules/dmrpp_module/Chunk.h
+++ b/modules/dmrpp_module/Chunk.h
@@ -372,7 +372,14 @@ public:
     {
         _duplicate(h4bs);
     }
-
+    virtual void release_chunk_buffer()
+    {
+        if (d_read_buffer_is_mine)
+            delete[] d_read_buffer;
+        d_read_buffer = nullptr;
+        d_read_buffer_is_mine = false;
+    }
+    
     virtual ~Chunk()
     {
         if(d_read_buffer_is_mine)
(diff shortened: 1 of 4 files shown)


modules/dmrpp_module/DmrppArray.cc
```diff
@@ -2220,10 +2220,13 @@ bool DmrppArray::read() {
     if ((var_type == dods_str_c || var_type == dods_url_c) && is_flsa()) 
         return read_string_array();
 
+    // Sonar cloud marks the following log vulnerable to injection attacks. So comment out for the time being.
+#if 0
```
jgallagher59701, line 2224: We could remove this.

modules/dmrpp_module/Chunk.h
```diff
@@ -372,7 +372,14 @@ class Chunk {
     {
         _duplicate(h4bs);
     }
-
+    virtual void release_chunk_buffer()
+    {
+        if (d_read_buffer_is_mine)
+            delete[] d_read_buffer;
```
jgallagher59701, line 378: This might be something we address in a future ticket. I wonder if Claude could provide suggestions WRT replacing the explicit memory management with unique pointers as private class members. 

But, for now, lets keep this as it is since this works and the change Sonar is asking for is more of a maintainability change.

jgallagher59701 (approved):
Two comments, but neither prevents merging the code.

PR #1424, opened 2026-08-24
first pass at SonarCloud fixes

sawer297:
## Description

TODO

## Tasks

- [ ] Tests added/updated
- [ ] Dead code removed
- [ ] No TODOs added

### DIFF
--- a/dispatch/BESFileLockingCache.cc
+++ b/dispatch/BESFileLockingCache.cc
@@ -751,17 +751,18 @@ void BESFileLockingCache::unlock_and_close(const string &file_name) {
  */
 unsigned long long BESFileLockingCache::update_cache_info(const string &target) {
     AdvisoryLockGuard write_alg(d_cache_info_fd, F_WRLCK);
-    unsigned long long current_size;
+    int fd = d_cache_info_fd;
+    unsigned long long current_size = 0;
 #if 0
     try {
         lock_cache_write();
 #endif
 
-    if (lseek(d_cache_info_fd, 0, SEEK_SET) == -1)
+    if (lseek(fd, 0, SEEK_SET) == -1)
         throw BESInternalError(prolog + "Could not rewind to front of cache info file.", __FILE__, __LINE__);
 
     // read the size from the cache info file
-    if (read(d_cache_info_fd, &current_size, sizeof(unsigned long long)) != sizeof(unsigned long long))
+    if (read(fd, &current_size, sizeof(unsigned long long)) != sizeof(unsigned long long))
         throw BESInternalError(prolog + "Could not get read size info from the cache info file!", __FILE__, __LINE__);
 
     struct stat buf;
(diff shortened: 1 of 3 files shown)


dispatch/BESFileLockingCache.cc
```diff
@@ -751,19 +751,23 @@ void BESFileLockingCache::unlock_and_close(const string &file_name) {
  */
 unsigned long long BESFileLockingCache::update_cache_info(const string &target) {
     AdvisoryLockGuard write_alg(d_cache_info_fd, F_WRLCK);
-    unsigned long long current_size;
+    int fd = d_cache_info_fd;
```
jgallagher59701, line 754: I don't understand why this is needed; can you explain?

dispatch/BESFileLockingCache.cc
```diff
@@ -751,19 +751,23 @@ void BESFileLockingCache::unlock_and_close(const string &file_name) {
...
+    unsigned long long current_size = 0;
 #if 0
```
jgallagher59701, line 756: We should get rid of this and the #if 0... #endif block at the end of the function.

ppt/TcpSocket.cc
```diff
@@ -136,8 +136,10 @@ void TcpSocket::connect()
     }
 
     sin.sin_port = htons(_portVal);
-    pProtoEnt = getprotobyname("tcp");
-    if (!pProtoEnt) {
+    struct protoent protoEntBuff;
+    char protoBuf[1024];
```
jgallagher59701, line 140: For this comment by sonarscan, C++11 has changes std::string so that we could use it here. I would use std::string and not std::array. Here's how it can work.

Yes. Here's an example:

```cpp
    constexpr int buffer_size = 64;
    
    // 1. Allocate string with desired capacity/size
    std::string result_str;
    result_str.resize(buffer_size);

    // 2. Pass writable buffer via &result_str[0]
    int written_len = legacy_c_function(&result_str[0], buffer_size);
```

Note that in code that's performance critical (this is not), you could resize the string after the call to `legacy_c_function()` using result_str.resize(written_len). That's not needed here because  1k is a pretty small chunk of memory and this is not called often.

ppt/TcpSocket.cc
```diff
@@ -281,8 +283,11 @@ void TcpSocket::listen()
     }
 
     BESDEBUG(MODULE, prolog << "Checking /etc/services for port " << _portVal << endl);
-    struct servent *sir = getservbyport(htons(_portVal), 0);
-    if (sir) {
+    struct servent servEntBuff;
+    char servBuf[1024];
+    struct servent *sir;
+    int servRet = getservbyport_r(htons(_portVal), 0, &servEntBuff, servBuf, sizeof(servBuf), &sir);
```
jgallagher59701, line 289: I'd replace the `0` with nullptr...

jgallagher59701 (approved):
I made some suggestions; this can be merged without another review regardless.

ppt/TcpSocket.cc
```diff
@@ -281,23 +302,31 @@ void TcpSocket::listen()
...
-                if (i > 0) error_oss << " or ";
+    struct sockaddr_in probe_addr{};
+    probe_addr.sin_family = AF_INET;
+    probe_addr.sin_port = htons(_portVal);
+    probe_addr.sin_addr.s_addr = INADDR_ANY;
 
-                error_oss << sir->s_aliases[i];
-            }
-        }
+    char service_name[NI_MAXSERV];
```
jhrg: For service_name[...] use

```
string service_name;
service_name.resize(NI_MAXSERV);
```

And then pass it using either `service_name.data()` or `&service_name[0]`

jhrg (commented):
Looks good, but see my suggestions

ppt/TcpSocket.cc
sawer297: done

PR #1430, opened 2026-09-09
Curl multi check in

kyang2014:
## Description

**Reference ticket:** HYRAX-2265

Use curl-multi APIs to manage the parallel data transfer. 

## Tasks


modules/dmrpp_module/CurlHandlePool.h
```diff
@@ -57,6 +57,7 @@ class dmrpp_easy_handle {
 
 public:
     dmrpp_easy_handle();
+    CURL * get_curl_handle() { return d_handle;}
```
jgallagher59701, line 60: Ignore this suggestion by SonarScan

modules/dmrpp_module/DmrppArray.cc
```diff
@@ -657,6 +662,323 @@ void DmrppArray::read_one_chunk_dio() {
     memcpy(target_buffer, source_buffer, the_one_chunk->get_size());
 }
 
+#if 0
```
jgallagher59701, line 665: Lets remove this if it is not used. If we're saving it to use later with the compute threads, lets add a comment about that.

modules/dmrpp_module/DmrppArray.cc
```diff
@@ -657,6 +662,323 @@ void DmrppArray::read_one_chunk_dio() {
...
+}
+
+// There may be 0.1% S3 failure rate, so we need to retry to see if we can obtain the data for a SuperChunk.
+enum class ParallelTransferStatus { PT_SUCCESS, PT_RETRYABLE, PT_FAILURE };
+
+static constexpr unsigned int MAX_ATTEMPTS = 3;
+static constexpr std::chrono::microseconds INITIAL_RETRY_BACKOFF{250000}; // 0.25s
+
+// Follow eval_curl_easy_perform_code() + eval_http_get_response() from CurlUtils.cc to check the retry results.
+static ParallelTransferStatus PT_result(CURL *easy, const shared_ptr<http::url> &data_url, CURLcode curl_code) {
```
jgallagher59701, line 712: Again, ignore these suggestions by SonarScan about CURL *

modules/dmrpp_module/DmrppArray.cc
```diff
@@ -657,6 +662,323 @@ void DmrppArray::read_one_chunk_dio() {
...
+        ostringstream oss;
+        oss << prolog << "Wrong number of bytes read for SuperChunk " << sc->id()
+            << "; read: " << transfer.super_chunk_internal->get_bytes_read() << ", expected: " << sc->get_size();
+        throw BESInternalError(oss.str(), __FILE__, __LINE__);
+    }
+    send_off(sc);
+
+}
+
+struct RetryCandidate {
```
jgallagher59701, line 756: We should  comment these fields. I think 'not_before' means don't retry this transfer before this time. The CurlMultiTransfer is, correct me if I'm wrong, one super_chunk that was being transferred in parallel with other super_chunks (that are, themselves, wrapped in CurlMultiTransfer objects).

modules/dmrpp_module/DmrppArray.cc
```diff
@@ -657,6 +662,323 @@ void DmrppArray::read_one_chunk_dio() {
...
+        cleanup();
+        throw;
+    }
+
+    cleanup();
+    
+}
+
+
+#if 0
```
jgallagher59701, line 920: Remove?

modules/dmrpp_module/DmrppArray.cc
```diff
@@ -657,6 +662,323 @@ void DmrppArray::read_one_chunk_dio() {
...
+                                           false);
+}
+
+void read_super_chunks_dio_concurrent(queue<shared_ptr<SuperChunk>> &super_chunks) {
+    read_super_chunks_concurrent_curl_multi(super_chunks,[](const shared_ptr<SuperChunk> &sc) { sc->read_dio_curl_multi(); },
+                                           true);
+}
+
+
+#if 0
```
jgallagher59701, line 968: Remove?

modules/dmrpp_module/DmrppArray.cc
```diff
@@ -657,6 +662,323 @@ void DmrppArray::read_one_chunk_dio() {
...
 }
 
+#if 0
+static ThreadCount &transfer_thread_count() {
+    static ThreadCount tc(DmrppRequestHandler::d_max_transfer_threads);
+    return tc;
+}
+#endif
+
+struct CurlMultiTransfer {
```
jgallagher59701, line 672: I realized this is pretty complicated - when I was reading the code it took me a while to understand that 'CurlMultiTransfer' is ours and not libcurl's! Anyway, I think I see how this is used, but I think some explanation is needed. The 'super_chunk' is transferred by the 'dmrpp_easy_handle.' The dmrpp_easy_handle unique_ptr has a custom deleter (that's worth pointing out).

The 'super_chunk_internal' is there to actually read the data.

Also, I think 're_try' should be 'retry.'

modules/dmrpp_module/DmrppArray.cc
```diff
@@ -657,6 +662,323 @@ void DmrppArray::read_one_chunk_dio() {
...
+    }
+}
+
+
+template <typename SendOffFn>
+static void finish_super_chunk_transfer(CurlMultiTransfer &transfer, SendOffFn send_off) {
+
+    auto &sc = transfer.super_chunk;
+    if (sc->get_size() != transfer.super_chunk_internal->get_bytes_read()) {
+        ostringstream oss;
```
jgallagher59701, line 747: One thing we can use to limit the number of ostringstream objects is to convert integers (and maybe flots/doubles) to strings with std::to_string. This can then become a string object and use '+' or string::append(). It's not that important here, but I thought I'd mention it.

modules/dmrpp_module/DmrppArray.cc
```diff
@@ -657,6 +662,323 @@ void DmrppArray::read_one_chunk_dio() {
...
+  
+                CURLMcode mc = curl_multi_add_handle(curl_multi, curl_handle);
+                if (mc != CURLM_OK)
+                    throw BESInternalError(prolog + "curl_multi_add_handle() failed: " +
+                                            curl_multi_strerror(mc), __FILE__, __LINE__);
+      
+                multi_transfer_maps.emplace(curl_handle, std::move(transfer));
+            }
+  
+            CURLMcode mc = curl_multi_perform(curl_multi, &still_running);
```
jgallagher59701, line 833: Ah, this is where the actual transfer(s) are initiated.  Let's comment that.

modules/dmrpp_module/SuperChunk.h
```diff
@@ -80,6 +83,11 @@ class SuperChunk {
     virtual unsigned long long get_offset() const { return d_offset; }
     virtual size_t get_chunk_count() const { return d_chunks.size(); }
     std::vector<std::shared_ptr<Chunk>> get_chunks() const { return d_chunks; }
+    char * get_read_buffer() { return d_read_buffer;}
+    bool get_d_is_read() const { return d_is_read;}
+    void set_read_buffer(unsigned long long size) {
+                                                   if(!d_read_buffer) 
+                                                        d_read_buffer = new char[size];}
```
jgallagher59701, line 90: We might be able to replace 'new char[...]' with a vector<char>(size) and avoid having to call delete.

modules/dmrpp_module/ThreadCount.h
```diff
@@ -0,0 +1,86 @@
...
+       }
+       return false;
+    }
+    void release_all_threads(std::list<std::future<bool>> &futures) const {
+
+        while (!futures.empty()) {
+            try {
+                if (futures.back().valid())
+                    futures.back().get();
+            } catch (...) {
```
jgallagher59701, line 76: I don't understand why there is an empty catch block. is there supposed to be 'pop_back()' call in here to be certain that if get() throws above the future is popped? If so, this should be 

futures.pop_back();
throw;

But if the block is not needed we should remove it.

jgallagher59701 (changes requested):
Most of my requests are for comments since this is such a complex and important part of the server, I think it needs to be a bit easier to understand.

modules/dmrpp_module/ThreadCount.h
kyang2014: Discussed. the code is correct.

jgallagher59701 (approved):
Looks good. Thanks!

